Issue #243, opened 2014-12-02
Error if file name contains French Character


cooljai:
Tried to upload a video and got the below error "You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings". Googled and found this connected to UTF-8. My filename contained the term "bibliothèque". Replaced the "è" with "e" and i was able to upload the video. Has anyone had similar experiences, and how can this be fixed?

tiptronic:
Same thing happens, if the asset description contains any umlaut-characters (äöüß, etc)

vpetersson:
@tiptronic Can you try pulling down the latest changes and see if that solves it?

tiptronic:
@vpetersson I can confirm - this works now! 

Thanks!

Btw: 
- Why can't I link an asset from the file-system? (e.g. '/home/pi/stuff/image.png')?
- Why can't I add an WebPage using the file-protocol? (e.g. 'file:///home/pi/stuff/index.html')

vpetersson:
Re your question: partially because we never got around to add that, but perhaps more importantly because of the security implications.

tiptronic:
One more thing - after the update, the settings are no longer read/written to '~/screenly/misc/screenly.conf'

tiptronic:
sorry - my bad: it's now in ''~/.screenly/screenly.conf'' (seems my docs are outdated - but what is the one in 'misc' good for?? ;))

vpetersson:
`~/.screenly/screenly.conf` is the correct path. I has been that all along.

> but what is the one in 'misc' good for?

That includes various scripts and config files used during the installation.

Issue #226, opened 2014-10-08
Non-ascii UTF-8 issue in asset name


tengil:
Getting the following error when editing asset name or activating asset when using non-ascii characters:

Server Error: 'ascii' codec can't encode character u'\xf8' in position 2: ordinal not in range(128)

I can create an asset containing any characters in the name but can not activate it.
I can not add non-ascii charaters to the name of an existing asset when editing.

Version: 2014-08-13

PR #244, opened 2014-12-02
GH-243 Fix unicode bug

richard-to:
Here's a fix for GH-243

This was a regression from GH-226. Basically I should have tested the case where the asset is uploaded. That's what the decode('utf-8') was for originally.

When an asset is upload, it gets uploaded as multipart/form-data. The form metadata for the asset gets posted as individual params instead of in one JSON encoded string via a param named model.

Since the data is not in JSON encoded string format, we skip the JSON decoding that converts the form data to unicode automatically.

This means we have to decode to utf-8 in this case. Here I make the assumption that utf-8 will always be the encoding if using Screenly's admin interface.

I did spend quite a bit of time trying to figure how to pass "Accept-Charset" using jQuery File upload, but it never added the header. You can add the `accept-charset` attribute to a form and jQuery File upload should look for it. You can also set the `formAcceptCharset` variable in the options. This had no effect though. Wasn't sure why.

Test Procedure:

1. Upload image file
2. Use unicode text for title
3. Save

1. Create asset using URL asset
2. Use unicode text for title
3. Save

Both should pass.

### DIFF
--- a/server.py
+++ b/server.py
@@ -114,7 +114,12 @@ def prepare_asset(request):
 
     def get(key):
         val = data.get(key, '')
-        return val.strip() if isinstance(val, basestring) else val
+        if isinstance(val, unicode):
+            return val.strip()
+        elif isinstance(val, basestring):
+            return val.strip().decode('utf-8')
+        else:
+            return val
 
     if all([get('name'),
             get('uri') or (request.files.file_upload != ""),


vpetersson:
Merging update without @aljungberg reviewing it :)

PR #313, opened 2015-09-29
Feature/merge new viewer

over64:
Merge with viewer PRO.
https://github.com/wireload/screenly-ose/issues/305

[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/313)


over64:
I have reverted check_update and log configuration. Also I removed wait_for_splash_page because it's unnecessary now. xloader.sh will wait for server.py ready. Also I fixed upgrade script. Now it kills xloader(no viewer) and restarts DM. After this xloader will provide all garanties that viewer.py will start AFTER server.py (via wget. PS - curl cannot retry on connection refused). Now wait_for_splash_page not needed anymore. I tested this case at least 5 times(reboot or upgrade). I added trap 'kill $(jobs -p)' EXIT to xloader.sh, because it's need for kill process tree, born by xloader. Without this xloader will die silently and child process(viwer.py and his childs) will be still running

assets_helper.py
```diff
@@ -28,7 +30,7 @@ def is_active(asset, at_time=None):
 
     if asset['is_enabled'] and asset['start_date'] and asset['end_date']:
         at = at_time or get_time()
-        return asset['start_date'] <= at <= asset['end_date']
+        return asset['start_date'] < at < asset['end_date']
```
aljungberg: Aren't date ranges normally inclusive for the beginning? E.g. if you want to play something between 12:00 and 13:00, it would start at 12:00 but it may end 12:59:59.999.

misc/upgrade.sh
```diff
@@ -33,8 +33,9 @@ python $SCREENLY/misc/migrate.py
 echo "Restarting app-server..."
 sudo supervisorctl restart screenly
 
-echo "Restarting viewer module..."
-pkill -f "viewer.py"
+echo "Restarting X(viewer)"
```
aljungberg: "Restarting X11 (viewer)..."

viewer.py
```diff
@@ -23,6 +18,12 @@
 import db
 import assets_helper
 
+
+__author__ = "Viktor Petersson"
```
aljungberg: I think we can change this to WireLoad Inc.

vpetersson: +1

misc/xloader.sh
```diff
@@ -9,6 +10,10 @@ xset -dpms          # Disable DPMS (Energy Star) features
 xset s noblank      # Don't blank the video device
 export NOREFRESH=1  # Fix white flickering in omxplayer
 
+# Wait for server.py to be up and running (maximum 28s)
+echo "Waiting for server.py to start" >> $LOG
```
aljungberg: "to start..." (adding the ... for consistency).

assets_helper.py
over64: asset.start_date and asset.end_date has datetime/ timestamp datatype. 
Related to -> 
```
CREATE TABLE assets(asset_id text primary key, name text, uri text, md5 text, start_date timestamp, end_date timestamp, duration text, mimetype text, is_enabled integer default 0, nocache integer default 0, play_order integer default 0);
```
And then we will compare dates with 1 ms accuracy. I tested just now with one minute accuracy. It works correct. In my opinion this is not important, strict equality or not. But I used rule: upstream(pro version) is more correct than not upstream :D I can revert if it need...

viewer.py
```diff
@@ -258,13 +269,14 @@ def pro_init():
     """Function to handle first-run on Screenly Pro"""
     is_pro_init = path.isfile(path.join(settings.get_configdir(), 'not_initialized'))
 
-    if is_pro_init:
-        logging.debug('Detected Pro initiation cycle.')
-        load_browser(url=HOME + INTRO)
-    else:
+    if not is_pro_init:
```
aljungberg: The code below this line is not needed in the OSE version. In the OSE version this is always False, and the Pro version uses different code here.

assets_helper.py
aljungberg: Actually we're changing this in a coming version. You're right it won't make much of a practical difference though. We're just changing it because we're detail obsessed developers. :)

misc/upgrade.sh
```diff
@@ -33,8 +33,9 @@ python $SCREENLY/misc/migrate.py
 echo "Restarting app-server..."
 sudo supervisorctl restart screenly
 
-echo "Restarting viewer module..."
-pkill -f "viewer.py"
+echo "Restarting X(viewer)..."
```
aljungberg: "Restarting X (viewer)..."

aljungberg:
LGTM.

vpetersson:
Thanks, @over64. Good job.

PR #318, opened 2015-10-05
fix asset time change after update

over64:
Use moment for crossbrowser convert utc to localtime. Spec for Date.parse (new Date) requires timezone. Server does not knows about timezones. We need to add 'Z' to provided by server string instead some browsers( IE and FF) will think that it is localtime

[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/318)

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -7,6 +7,8 @@ from sh import grep, netstat
 from urlparse import urlparse
 from datetime import timedelta
 from settings import settings
+from datetime import datetime
+import pytz
 
 HTTP_OK = xrange(200, 299)
 
@@ -79,8 +81,10 @@ def get_video_duration(file):
 
 
 def handler(obj):
-    if hasattr(obj, 'isoformat'):
-        return obj.isoformat()
+    # Set timezone as UTC if it's datetime and format as ISO
+    if isinstance(obj, datetime):
+        with_tz = obj.replace(tzinfo=pytz.utc)
+        return with_tz.isoformat()
     else:
         raise TypeError('Object of type %s with value of %s is not JSON serializable' % (type(obj), repr(obj)))
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,4 @@ six==1.8.0
 uptime==2.0.2
 werkzeug==0.10.4
 wsgiref==0.1.2
+python-dateutil==2.4.2
--- a/server.py
+++ b/server.py
@@ -30,6 +30,7 @@ from lib.utils import get_node_ip
 from lib.utils import validate_url
 from lib.utils import url_fails
 from lib.utils import get_video_duration
+from dateutil import parser as date_parser
 
 from settings import settings, DEFAULTS, CONFIGURABLE_SETTINGS
 from werkzeug.wrappers import Request
@@ -173,13 +174,14 @@ def prepare_asset(request):
             # Crashes if it's not an int. We want that.
             asset['duration'] = int(get('duration'))
 
+        # parse date via python-dateutil and remove timezone info
         if get('start_date'):
-            asset['start_date'] = datetime.strptime(get('start_date').split(".")[0], "%Y-%m-%dT%H:%M:%S")
+            asset['start_date'] = date_parser.parse(get('start_date')).replace(tzinfo=None)
         else:
             asset['start_date'] = ""
 
         if get('end_date'):
-            asset['end_date'] = datetime.strptime(get('end_date').split(".")[0], "%Y-%m-%dT%H:%M:%S")
+            asset['end_date'] = date_parser.parse(get('end_date')).replace(tzinfo=None)
         else:
             asset['end_date'] = ""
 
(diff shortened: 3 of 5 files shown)


aljungberg:
If the server is not sending full ISO8601 datetimes, shouldn't we fix the server rather than blindly adding a `Z`? We should send correct timestamps in the API even that it's just a local UI, it's just good practice.


---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[static/js/screenly-ose.coffee, line 23 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/318#-K0GNrMR5j40SKy0w4Q_)** ([raw file](https://github.com/wireload/screenly-ose/blob/7c4cfe1d90888c1266e5ed64aaf34d4f7f8a2b72/static/js/screenly-ose.coffee#L23)):</sup>
Formatting and typo fix: 

    # Cross-browser UTC to localtime conversion


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/318)

over64:
I checked this out at first. The problem that server knows nobody about timezones. Server sends correct ISO datetime format, but without timezones. But if tz not presented browser engine can interpret this as it want. Chrome interprets as UTC+0, firefox and IE as localtime. The main problem, that not our code provides date format. JSON serializer provide this format. And it seems, that we need change data type for dates at server. But it required too much modifications...


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/318)

aljungberg:
We generally prefer if the server knows about timezones and uses UTC but for Screenly OSE we might consider all times to be "floating", e.g. they are 'naive on purpose' and always relative to the current timezone.

Regardless the server should be able to add the local timezone in the API output (and take it in the input) even if we don't fix the server side situation today. We could use a custom JSON serialiser or pytz `localize`.


---

Reviewed 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/318)

aljungberg:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[utils.py, line 83 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/318#-K1sU69fYmRYqaCsZo0L)** ([raw file](https://github.com/wireload/screenly-ose/blob/c6eba0d51922d7c8e929c5992292524f4be1456e/utils.py#L83)):</sup>
Are you sure the local timezone is always UTC? The idea was to add the **local** timezone to the dates returned by our server to correctly convey that they are in local time, whatever zone that is.

If we always force the client browser to be in UTC this is fine though.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/318)

over64:
_[utils.py, line 83 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/318#-K1sU69fYmRYqaCsZo0L:-K3rf4lGJ5ILbYNhJhzz:-1156234138) ([raw file](https://github.com/wireload/screenly-ose/blob/c6eba0d51922d7c8e929c5992292524f4be1456e/utils.py#L83)):_
Server should always reponse dates(for assets) in UTC+0. SQLite knows anything about timezones. Browser will calculate 'local' offset. We can move this code to controller's methods to do this more explicit. But we will need to write this code in 4-5 places. But is it real case that we will return dates not in UTC+0 ?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/318)_

jelofson:
Looking forward to a fix for this. Causes quite a bit of confusion for the person managing the screenly assets. The pi is in MST and every asset added in FF is +7 hours ahead of local. Chrome ok. IE is not (in more ways than one).

vpetersson:
@over64 Looks like this helped us identify another issue. Could you confirm that:
 
* All times in the backend are stored in UTC.
* They are converted back and forth from UTC to the local timezone in the UI.

Also, please note that we cannot assume anything about the timezone on the actual system.

over64:
@vpetersson the main idea of patch is
1)on asset save:
on client: send time to server in UTC. Browser makes utc to local conversion itself.
```
# Cross-browser UTC to localtime conversion
dd = moment.utc(d).local()

```
on server: keep date in utc and truncate tz info
2) on render assets:
on server: return date with tz=UTC
on client: convert utc to localtime on render
Hence =>
1. There is no 'jumps' with time on FF and IE after patch
2. Time stored in db will never be the same (if you are not living in UTC+0) as at UI. But it's ok

over64:
rebased. Tested just now:
Chrome:
utc1
FF:
utc2

No jumps on update, time is equal. IE not tested now (but tested earlier), I have not windows now(

PR #324, opened 2015-10-22
Feature/refactor settings

over64:
refactor settings, settings load / save. Unit tests for this

[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/324)


server.py
```diff
@@ -34,13 +34,20 @@
...
 
-from settings import settings, DEFAULTS
 from werkzeug.wrappers import Request
+
+from settings import DEFAULTS
+from settings import config_dir
+from settings import config_file
+from settings import load as load_settings
+from settings import save as save_settings
+
```
femdom: What "load_setting" actually does? I don't think that performing actions silently during module load is good idea.

settings.py
```diff
@@ -2,13 +2,19 @@
...
 import ConfigParser
 import logging
-from UserDict import IterableUserDict
+import copy
 
 CONFIG_DIR = '.screenly/'
 CONFIG_FILE = 'screenly.conf'
+
+
+def config_dir(home=getenv('HOME')): return path.join(home, CONFIG_DIR)
```
femdom: E701 multiple statements on one line (colon)

settings.py
```diff
@@ -2,13 +2,19 @@
...
+
+
+def config_file(home=getenv('HOME')): return path.join(home, CONFIG_DIR, CONFIG_FILE)
```
femdom: E701 multiple statements on one line (colon)

settings.py
femdom: Please check your code with, at least, pep8 tool.

femdom:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[tests/settings_test.py, line 45 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1FP_B6uPR1txbvCfC4)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/tests/settings_test.py#L45)):</sup>
Consider using BDD style test names. Like test_some_behavior_description. test_accessing_missing_attribute_raises_attribute_error


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

server.py
over64: load_settings it's just a function, which can load settings. There is no silently actions on module load))

over64:
<sup>**[tests/settings_test.py, line 45 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1FP_B6uPR1txbvCfC4)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/tests/settings_test.py#L45)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[settings.py, line 13 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1Dh6RyMC1iv47KH0Sh-r1-13)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/settings.py#L13)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[settings.py, line 16 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1Dh6RyMC1iv47KH0Sh-r1-16)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/settings.py#L16)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

aljungberg:
Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

<sup>**[server.py, line 44 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1Dh6RyMC1iv47KH0Sg-r1-44)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/server.py#L44)):</sup>
There is still some silent actions on module load on the next line, `settings = load_settings()` -- can we move this into an `if __name__ == '__main__':` block? That way `server.py` may be imported for unit testing purposes etc without loading settings.


---

<sup>**[settings.py, line 77 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1s_MGlguZ_V25mEVnz)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/settings.py#L77)):</sup>
`class ScreenlySettings(object):`


---

<sup>**[tests/settings_test.py, line 45 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1FP_B6uPR1txbvCfC4)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/tests/settings_test.py#L45)):</sup>
You only fixed it for this method though. Let's fix it for all of them (where it makes sense). Here's [our guide on testing](https://ph.wireload.net/w/tests/) (sorry internal only).

In general you want `X, when Y, should Z` to be your format. Some very simple tests can be shortened -- so instead of `X should succeed` we often just write X.


---

<sup>**[tests/settings_test.py, line 49 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sb0o2HMS6cbm8c9EI)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L49)):</sup>
So in this case you might write `test_set` just like you have because the alternative would be `test_set_should_work` or whatever and the `_should_work` suffix is kind of redundant.


---

<sup>**[tests/settings_test.py, line 54 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbCBrULaQhfKLipmZ)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L54)):</sup>
But here you can use a full BDD name: `test_load_settings_for_missing_file_should_raise_ioerror`.

When you use names like this the names of the tests become like a checklist of things we have thought about. When the checklist makes sense at a glance we can be fairly confident about the code.


---

<sup>**[tests/settings_test.py, line 72 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbUsoOe2X39SkMRdE)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L72)):</sup>
`test_load_with_bad_data_should_provide_default_instead` or something like that.


---

<sup>**[tests/settings_test.py, line 84 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbgTOQIb16Ps4xEpk)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L84)):</sup>
Formatting. When breaking a list literal into multiple lines we put one entry on each line and indent +1 tab (hanging indent). We don't do ASCII art in our code (aligning stuff vertically beyond the basic indentation policy or artistically laying them out).

    hello(a, [
        1,
        2,
    ])


Although in this case though wouldn't a simple string literal make more sense and be easier to read?

    self.assertEqual(open(save_to, "rb").read(),
    """[main]
    assetdir = dir
    """)


---

<sup>**[tests/settings_test.py, line 98 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shoxyKo9x2qR6V_2Z)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L98)):</sup>
Don't forget to delete your tmp file.


---

<sup>**[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):</sup>
Can we avoid this? It's again import time code, plus it creates a second instance of the same settings object, doesn't it?

I think we should have a module which stores this global and initialises it to an empty object. Then when server.py loads it can replace it with proper settings, and so can any unit tests.


---

<sup>**[viewer.py, line 39 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sklvpVdIdumMw-DMj)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/viewer.py#L39)):</sup>
Same here -- we'll want to use a main block -- otherwise `viewer.py` can't be imported for unit testing without automatically loading settings.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):</sup>
how we can avoid this? This project has no properly code organization with OOP principal. All code(functions) written in modules - not in class. For modules we cannot do DI properly because all module(due to python spec) is singleton object. And we have no control on import/ creation. We can do something like this:
import utils
utils.settings = load_settings()

But it's wiered too! We have partial initialized module after import. 
We have 3 ways:
1) Use modules properly - module without initialization and all dependencies must be passed in functions(which will be called) directly
2) Use OOP and wrap all code with DI into classes. After this we will can control DI via object creation. 400+ lines of diffs expected xD
3) Live with this quietly. Yes, with this model we will load some components twice or more

Also settings it's really complexive thing. Because in some places of code we need to reinitialize this.

In my opinion 2 - it's best way. It solves all problems. But it is require too much diff. What way we should use?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):</sup>
Before this I used strategy: minimal diff - minimal breaks


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

aljungberg:
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

<sup>**[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):</sup>
Right but that's the point of moving it into its own module. It's not the best or the most clean solution, but at least it solves the immediate problem. These are the two problems we have with the current situation:

A) `utils.settings != server.settings` -- they're different objects.
B) importing `utils` or `server` automatically tries to load a settings file, which interferes with unit testing.

So instead of `settings = load_settings()` you'd do it like in Django: `from conf import settings` -- `settings` is a lazily initialised proxy for the real settings object and it's a singleton so you only get 1 in the whole app. And it's easy to mess with it when unit testing. In this case:

A) `utils.settings == server.settings == conf.settings` because they all share the same singleton proxy.
B) ...but nothing is actually executed when you import `utils` or `server` because the conf.settings object is lazy loaded.
C) ...and it's easy to mock/patch it to do something else before it's actually loaded. 


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):</sup>
D) after load settings becomes a real singleton. And it will be alive and unique for python VM instance. if we have 10 tests then we need to ran PVM 10 times for different settings config or we need to implement settings reload routine.

But ok, I undertand that this solution is best. Ok, I will keep this way


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[server.py, line 44 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1Dh6RyMC1iv47KH0Sg-r1-44)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/server.py#L44)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[settings.py, line 77 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1s_MGlguZ_V25mEVnz)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/settings.py#L77)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 45 \[r1\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1FP_B6uPR1txbvCfC4)** ([raw file](https://github.com/wireload/screenly-ose/blob/f718165ec8e2f5fc1b245900eb5ba146dfdcd505/tests/settings_test.py#L45)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 49 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sb0o2HMS6cbm8c9EI)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L49)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 54 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbCBrULaQhfKLipmZ)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L54)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 72 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbUsoOe2X39SkMRdE)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L72)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 84 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbgTOQIb16Ps4xEpk)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L84)):</sup>
this is result of PyCharm formatter work. Is it wrong?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[tests/settings_test.py, line 98 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shoxyKo9x2qR6V_2Z)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L98)):</sup>
Am. Maybe system should delete this file? Because it's tmp))


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

over64:
<sup>**[viewer.py, line 39 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sklvpVdIdumMw-DMj)** ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/viewer.py#L39)):</sup>
fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)

aljungberg:
Review status: 0 of 6 files reviewed at latest revision, 11 unresolved discussions.

---

_[tests/settings_test.py, line 84 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbgTOQIb16Ps4xEpk:-K3bHDLfKvXtU6jYDrPA:-1853349402) ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L84)):_
Yes, our formatting policy is different from PyCharm's default in at least two important ways:

- we like long lines
- we don't do vertical aligning or other such ASCII art

---

_[tests/settings_test.py, line 98 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shoxyKo9x2qR6V_2Z:-K3bH_IPDcvGEtHEXlGw:1925909495) ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L98)):_
It'll probably be deleted sooner or later but that could be in a long time. It's always better to clean up after ourselves.

---

_[tests/settings_test.py, line 51 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K3bGjmIRbqBQ3Swg6x-:-K3bGjmIRbqBQ3Swg6x0:-1181811498) ([raw file](https://github.com/wireload/screenly-ose/blob/e905aa94c8b99f201068032bb562d4a690ebc9fe/tests/settings_test.py#L51)):_
"rise" -> "raise" 

(rise means something rises by itself, raise means someone raises something else. So the sun rises but you raise an exception.)

---

_[utils.py, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shuq0hvv2H2B62TjY:-K3bHswKWV8Tyg0VYMmT:-1564744948) ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/utils.py#L10)):_
D) yes it becomes a real singleton but since it's just a proxy for the real settings it's easy to replace it or reinitialise its actual contents in unit tests.

---

_[viewer.py, line 41 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K3bIB8hVLlXenvXNtHC:-K3bIB8hVLlXenvXNtHD:-1092012749) ([raw file](https://github.com/wireload/screenly-ose/blob/e905aa94c8b99f201068032bb562d4a690ebc9fe/viewer.py#L41)):_
This line has vertical alignment (ASCII art). Take a look at our contribution guidelines -- we only use one extra indent for continuation lines.

Also this code still needs to be in a main block or it will run from merely importing `viewer`.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

aljungberg:
Reviewed 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

over64:
_[tests/settings_test.py, line 51 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K3bGjmIRbqBQ3Swg6x-:-K3jgZSEO2NTqqVDxQiv:-467915377) ([raw file](https://github.com/wireload/screenly-ose/blob/e905aa94c8b99f201068032bb562d4a690ebc9fe/tests/settings_test.py#L51)):_
fixed

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

over64:
_[tests/settings_test.py, line 84 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1sbgTOQIb16Ps4xEpk:-K3jgbFqLb0Jy-hhqxpU:-467915377) ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L84)):_
fixed

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

over64:
_[tests/settings_test.py, line 98 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K1shoxyKo9x2qR6V_2Z:-K3jgdlfXTJfICW3co8d:-2118096774) ([raw file](https://github.com/wireload/screenly-ose/blob/2ef88aad2e5a597441823e550b866dd93cceddc7/tests/settings_test.py#L98)):_
fixed via tearDown for test class

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

over64:
_[viewer.py, line 41 \[r3\]](https://reviewable.io:443/reviews/wireload/screenly-ose/324#-K3bIB8hVLlXenvXNtHC:-K3jhUJEa8tyYESH5LPB:225849999) ([raw file](https://github.com/wireload/screenly-ose/blob/e905aa94c8b99f201068032bb562d4a690ebc9fe/viewer.py#L41)):_
fixed this

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/wireload/screenly-ose/324)_

vpetersson:
@antonmolodykh Can you take a look at this one. Not sure if it is relevant anymore. Would be nice to close out if not.

antonmolodykh:
@vpetersson  I think we can close it.

PR #332, opened 2015-11-26
explicitly set 'sudo: false' for migration to new travis infrastructure

over64:
[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/332)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ before_script:
   - cp misc/screenly.conf ~/.screenly/
   - cp misc/uzbl.rc ~/screenly/misc/
 script: nosetests -v -a '!fixme'
+
+sudo: false


vpetersson:
Why do we need to do this?

over64:
https://docs.travis-ci.com/user/migrating-from-legacy/
from manual
````
How can I use container-based infrastructure? #

If you see This job is running on container-based infrastructure in your build log it means you are already running builds on our new container-based infrastructure.

If you don’t, to use the new infrastructure add the following line to your .travis.yml:

sudo: false
```

vpetersson:
Right...but we're not using a container based infrastructure. Check out the [Travis file](https://github.com/wireload/screenly-ose/blob/master/.travis.yml).

over64:
container based - it's new architecture for Travis based on Docker images. But Docker doesn't allows sudo.
Before this Travis used virtual machines which was slow and heavy. This repo was created when travis has no containers. For migration to dockerized containers(Travis team strongly recommends do it. And shows warning in UI) we need to explicitly said to travis that we are not using sudo. After this Travis will automaticaly switch our build to container-based arch. Tornado build for example https://github.com/tornadoweb/tornado/blob/master/.travis.yml

vpetersson:
Ok, sounds reasonable.

PR #335, opened 2015-12-08
try do HEAD then do GET if fails for url check 

over64:
[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/335)

PR #350, opened 2016-01-15
Small bug fix for servers that do not support HEAD

JoshuaEstes:
It took me awhile to track this down, when putting in the URL it returned an error that the file could not be found. Putting the URL into the browser returned status code 200. After doing a little research I discovered the the server does not support HEAD and will return a 404. However when doing a GET it will return a 200.

This should solve a problem for a small minority of users.

[Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/350)

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -93,7 +93,7 @@ def url_fails(url):
     """
     try:
         if validate_url(url):
-            obj = requests.head(url, allow_redirects=True, timeout=10, verify=settings['verify_ssl'])
+            obj = requests.get(url, allow_redirects=True, timeout=10, verify=settings['verify_ssl'])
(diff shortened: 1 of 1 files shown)


vpetersson:
I'm not too comfortable with this change. Fetching the whole page versus just HEAD adds a lot of overhead.

Besides, a web server that lacks proper support of HEAD is a pretty serious bug. I would suggest filing a bug report with that project instead.

vpetersson:
@JoshuaEstes There are situations where this is indeed required (like Google Calendar). However, after taking a closer look at this, I noticed that it's a duplicate of #335.

Thank you for your contribution though.

Issue #333, opened 2015-12-01
24 hr clock setting not applied


qbicdesign:
The 24hr clock feature seems to be broken. 
After turning it on, and clicking save, it reverts back to the off position, and the times remain in AM/PM format.

mlott:
Just updated the latest version.  I'm getting the same problem.

thorstenhartung:
Confirmed, same here

xh3ni0n:
I used a workaround on CLI
simply edit ```~/.screenly/screenly.conf``` and set ```use_24_hour_clock = on```

qbicdesign:
V cool. Thanks!

Frankstar:
@xh3ni0n ty.
Cant find (yet) why its not working ..

vpetersson:
Yeah I was able to reproduce this on the latest version.

vpetersson:
Landed #365 which solves this issue. Thanks. @robberwick.

PR #365, opened 2016-03-11
Save 24 clock setting when amended in UI

robberwick:
Addresses #333 - 24 hr clock setting not applied

Add CONFIGURABLE_SETTINGS dict to settings file and
iterate when saving settings from UI

UI settings include use_24_hour_clock setting, but only the
DEFAULTS['viewer'] dict were being checked, resulting in the time
preference not being persisted.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/365)

### DIFF
--- a/server.py
+++ b/server.py
@@ -34,7 +34,7 @@ from utils import validate_url
 from utils import url_fails
 from utils import get_video_duration
 
-from settings import settings, DEFAULTS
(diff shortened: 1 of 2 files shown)


vpetersson:
LGTM. Merging. Thanks.

Issue #348, opened 2016-01-12
enabling 24 hour clock reverses date


qbicdesign:
I enabled 24hr clock mode in screenly.conf, but unfortunately this affects the date format too.
Once enabled the date appears in reverse format YYYY/MM/DD. Whilst understandable, its impossible to change the year, as whilst typing the new year the cursor skips to the end and you're unwittingly changing the day...  basically the date-picker can't handle the reverse date

vpetersson:
Ah we had a similar issue in Screenly Pro. Will try to get this scheduled for a future spring.

PR #366, opened 2016-03-11
Fix datetime format for 24hr clock setting in UI & datepicker

robberwick:
Addresses #348 - enabling 24 hour clock reverses date

Updated values in `date_settings_24hr` to be analogous to 12hr versions - i.e. MM/DD/YY 

I updated this in the coffescript file and transpiled to JS using the latest version of coffeescript from npm. There is thus a few more changes to the resultant js file than just the object that was changed, thanks to changes in coffeescript since the last version was created. I hope that's ok - if you need it creating using the specific version just let me know and I'll update this PR. Assuming the change is acceptable, of course!

---
This change is [Review on Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/366)


vpetersson:
@robberwick Thanks! Sorry for the late follow up. LTGTM but would you mind fetching the changes upstream so that in can be merged?

robberwick:
No problem at all - I've rebased onto current master (01f03af53b163f721f3cf9b612e0c859c93e90ac) - as it was only a single commit it seemed cleaner. I hope that's ok.

vpetersson:
👍

PR #379, opened 2016-04-19
add gtkrc

over64:
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/379)

### DIFF
--- a/misc/ansible/screenly.yml
+++ b/misc/ansible/screenly.yml
@@ -19,11 +19,10 @@
         dest: /home/pi/.screenly/screenly.conf
         force: no
 
-    - name: Symlink gtk config
-      file:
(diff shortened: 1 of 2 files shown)


misc/ansible/screenly.yml
```diff
@@ -19,6 +19,12 @@
         dest: /home/pi/.screenly/screenly.conf
         force: no
 
+    - name: Symlink gtk config
+      file:
+        src: /home/pi/screenly/misc/gtkrc-2.0
+        dest: /home/pi/.gtkrc-2.0
+        state: link
```
vpetersson: Let's not symlink. Let's just copy in the file instead.

PR #380, opened 2016-04-21
make new directory structure

over64:
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/380)


vpetersson:
Nice! Looks like tests need to be updated tho.

vpetersson:
Also, looks like we're missing `/ansible/site.yml` which should call on the roles `system` and `screenly`.

ansible/screenly.yml
```diff
@@ -0,0 +1,14 @@
+- name: Install Screenly system dependencies
+  hosts: all
+  user: pi
+  become: yes
+
+  roles:
+    - system
+
```
vpetersson: Add `-screenly` here. That will speed up the run time.

ansible/screenly.yml
```diff
@@ -0,0 +1,14 @@
+- name: Install Screenly system dependencies
```
vpetersson: "Installs Screenly"

over64:
About uzbl.rc - it is not managed by Ansible because it's not a member of installation process. This file using in viewer.py https://github.com/wireload/screenly-ose/blob/master/viewer.py#L34
And this is bug. Because this file should not be home/screenly/misc/uzbl.rc. This file must be repo-related. As I know viewer uses repo root dir as CWD. After this fix this line https://github.com/wireload/screenly-ose/blob/master/.travis.yml#L15 will be redundant. And all things will be much more clearer)) Because if I will clone OSE repo to bare machine and I'll run viewer.py - I'll have stack trace because /home/pi/screenly/misc.uzbl.rc not found (because repo root screenly-ose). It's strange. I think that we could close this PR and just open new issue about this.

Issue #400, opened 2016-04-27
Move logging stdout/stderr


vpetersson:
With the move to systemd we should also move the logging.

Currently, we're logging to files, but since systemd already takes care of this, we should start to simply log to stdout/stderr.

vpetersson:
We also need to update [this](https://github.com/wireload/screenly-ose/blob/master/server.py#L294-L299) code.

vpetersson:
We also need to change the log format. Here's how the logging looks right now:

```
May 02 09:34:31 raspberrypi python[2229]: Mon, 02 May 2016 09:34:31 Generating asset-list...
May 02 09:34:32 raspberrypi python[2229]: Mon, 02 May 2016 09:34:32 Showing asset WireLoad (webpage)
May 02 09:34:32 raspberrypi python[2229]: Mon, 02 May 2016 09:34:32 current url is http://www.wireload.net
May 02 09:34:32 raspberrypi python[2229]: Mon, 02 May 2016 09:34:32 Sleeping for 30
May 02 09:35:03 raspberrypi python[2229]: Mon, 02 May 2016 09:35:03 Showing asset Screenly Tour (webpage)
May 02 09:35:03 raspberrypi python[2229]: Mon, 02 May 2016 09:35:03 current url is https://www.screenlyapp.com/tour/
May 02 09:35:03 raspberrypi python[2229]: Mon, 02 May 2016 09:35:03 Sleeping for 60
```

As you can see, the time stamp is redundant.

over64:
https://github.com/wireload/screenly-ose/pull/407

over64:
There is some issues:
It seems that we cannot use journalctl for this because:
1) on debian 7 it's named ```systemd-journalctl``` but on debian 8 - ```journalctl```
2) on debian 7 it does not recognize -u option for filter by unit name.
The simpliest way is use ```systemctl status screenly-web.service -n 50```. Is it ok?
It will produce something like
```
● ssh.service - OpenBSD Secure Shell server
   Loaded: loaded (/lib/systemd/system/ssh.service; enabled; vendor preset: enabled)
   Active: active (running) since Чт 2016-04-14 19:28:06 OMST; 2 weeks 3 days ago
  Process: 4645 ExecReload=/bin/kill -HUP $MAINPID (code=exited, status=0/SUCCESS)
 Main PID: 584 (sshd)
   CGroup: /system.slice/ssh.service
           └─584 /usr/sbin/sshd -D

[...]

PR #405, opened 2016-04-30
Fixes Gunicorn timeout

vpetersson:
Bumps up timeout to 240s. This isn't a proper fix to large uploads, but rather a bandaid. A proper fix would be to chunk the uploads to mitigate the timeout issue entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/405)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:wheezy
+FROM debian:jessie
 MAINTAINER Viktor Petersson <vpetersson@wireload.net>
 
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


jscaptura:
This helped with some of the large assets, but is still too short for the 400-600mb assets.

-- Jon

vpetersson:
@jscaptura Did this work in the past?

jscaptura:
Yes.

jscaptura:
As a temporary fix, I bumped the timeout up to 480 for now, and it is working. In a perfect world with full 100 Mbps transfer rates, 240 would probably work fine. I'm only seeing rates around 20 Mbps though.

vpetersson:
Yeah, it's most likely limited by CPU or disk i/o rather than the network.

jscaptura:
Since I bumped the timeout up to 480, it appears that the upgrades fail now.

`localhost | FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "Local modifications exist in repository (force=no)."
}
`
As a result, I no longer have a viewer log. Instead I have this message, even though the server is running and assets are showing on the screen.

`
(no viewer log present -- is only the screenly server running?)`

How can I force the upgrade?

vpetersson:
The error above is because your manual changes. 

You can fix it by doing the following:

```
 $ cd ~/screenly
 $ git reset --hard
```

After that, you should be able to run the upgrade.

In terms of logging, that's a known issue and is being tracked in #400 (as this is related to moving logging to systemd).

PR #406, opened 2016-05-02
log to stdout instead of file

over64:
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/406)


vpetersson:
That was easier than I anticipated. :)

vpetersson:
Refs #400.

PR #422, opened 2016-05-11
Adds reload functionality of existing asset.

vpetersson:
TODO: Needs to be tested.

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/422)

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -181,12 +181,9 @@ def browser_url(url, cb=lambda _: True, force=False):
     global current_browser_url
 
     if url == current_browser_url and not force:
-        logging.debug('Already showing %s.', current_browser_url)
-
-        # For some reason this invokes a weird black screen issue
-        # when going image -> image.
-        # logging.debug('Already showing %s, reloading it.', current_browser_url)
-        # browser_send('reload full')
+        # Todo: test image -> image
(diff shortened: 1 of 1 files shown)


xdozex:
Any idea when this fix will be included in the custom image?

vpetersson:
The easiest way is to just run the upgrade script. No need to re-flash the device.

xdozex:
Thanks for the info, unfortunately I'm a pretty low-level user, and have no idea how to run the upgrade script :). Mind linking to some reference material if it's available?

vpetersson:
Sure. Just SSH into the device and run:

```
bash <(curl -sL https://www.screenlyapp.com/install-ose.sh)
```

Freyli:
Is this fixed? doesn't seem to be working for me. IE Google Slides not updating

vpetersson:
@Freyli Yes this should have been fixed. How are you testing this?

dvblystone:
I have two Screenly OSE devices that are running on a wireless connection. I have a single Google Slides presentation that they both are displaying. Every morning the slides are updated with current information and after saving the new presentation does not update on the Screenly. I have updated one to the most current version and the other to the advanced version. Neither one will update the content.

Issue #431, opened 2016-05-31
Fix restart of systemd services


vpetersson:
We already have Ansible [code](https://github.com/wireload/screenly-ose/blob/master/ansible/roles/screenly/tasks/main.yml#L63-L67) for restarting the sytemd services. However, for some reason this doesn't work. We need to investigate and fix this.

vpetersson:
Closed in #440 440

PR #440, opened 2016-06-07
inline handlers into playbook

over64:
workaround for bug in ansible 2.x
https://github.com/ansible/ansible/issues/13485

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/440)


vpetersson:
👍 

Please provide proper references in pull requests going forward.

Should have been:
```
Closes #431. Moves playbook handlers to inline
``

Then a note with the explanation why.

PR #458, opened 2016-08-10
fix black screen between image assets

over64:
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/458)

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -180,9 +180,12 @@ def browser_clear(force=False):
 def browser_url(url, cb=lambda _: True, force=False):
     global current_browser_url
 
-    current_browser_url = url
-    browser_send('uri ' + current_browser_url, cb=cb)
-    logging.info('current url is %s', current_browser_url)
+    if url == current_browser_url and not force:
+        logging.debug('Already showing %s, reloading it.', current_browser_url)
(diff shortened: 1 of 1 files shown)


rsenio:
Is this now available on the regular update instead of the dev release?

vpetersson:
No, only in the dev release for now. We need to do some more testing first.

vpetersson:
@MydKnight This is for *images*. We can't do seamless transition between videos in Screenly OSE because of the way the playback mechanism works (i.e. it's using separate threads of `omxplayer` for each video). We do however have seamless playback in the commercial version. For more info see [this page](https://www.screenly.io/screenly-ose-vs-pro/).

MydKnight:
Ah, ok. So screenly isn't going to work as a replacement for the product
I've been using for a few years which seems to get around the problem by
precaching. I like the web interface, but the black screen is going to be a
poison pill for us, unfortunately.

vpetersson:
@MydKnight we're happy to accept contributions to address the issue, but as mentioned, it's a solved problem in the commercial version.

MydKnight:
No worries. Found another solution that I was able to successfully containerize.

PR #464, opened 2016-08-12
Ensure the Dockerfile works with current version of PIP

tomvleeuwen:
I was unable to create a docker due to PIP issues. Using these fixes, building the docker works fine.

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/464)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,9 @@ RUN apt-get update && \
     apt-get clean
 
 # Install Python requirements
+RUN pip install web.py
 ADD requirements.txt /tmp/requirements.txt
+RUN pip install --upgrade cffi
 RUN pip install -r /tmp/requirements.txt
 
 # Create runtime user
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,7 +6,9 @@ RUN apt-get update && \
     apt-get clean
 
 # Install Python requirements
+RUN pip install web.py
```
vpetersson: Hmm isn't this a dependency of Bottle. It should be covered by there. Generally speaking, we should have all `pip` requirements inside `requirements.txt`.

tomvleeuwen:
I am sorry, that line was not supposed to end in here, it is not required to build the docker. I removed it from the pull request.

vpetersson:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[Dockerfile, line 10 \[r2\]](https://reviewable.io:443/reviews/wireload/screenly-ose/464#-KP4kk_tXdD8EUCd1U9-:-KP4kk_tXdD8EUCd1U90:b-bwhttr) ([raw file](https://github.com/wireload/screenly-ose/blob/662a242ee152f16d585764a4185edf9ab79080cd/Dockerfile#L10)):*
> ```Dockerfile
> # Install Python requirements
> ADD requirements.txt /tmp/requirements.txt
> RUN pip install --upgrade cffi
> ```

What exact version do we need here? I would prefer locking that down in `requirements.txt` instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wireload/screenly-ose/464)*

PR #482, opened 2016-08-30
Update URL to new domain, force HTTPS only

leonelwilliams:
--proto '=https' forces curl to only use https:// which will prevent another http:// redirect mistake from occurring anywhere in the chain

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/482)


vpetersson:
LGTM. Thanks.

PR #483, opened 2016-08-30
Use https:// for GitHub repo

leonelwilliams:
GitHub does a 301 from http:// to https:// - we can avoid the redirect and interception risk if https:// is used directly

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/483)


vpetersson:
LGTM. Thanks.

Issue #444, opened 2016-06-14
Large asset uploads timing out


jscaptura:
Since the switch to GUNICORN, I've had trouble uploading video assets of 400mb+ size. To get around this, I had increased the timeout to give the uploads time to finish. I've since reverted to using the distributed code without any local modifications.

Now, I have a new batch of video assets to upload, and while the progress bar gets to the end, after a while it goes back to the beginning and the "The operation failed. Please reload the page and try again." error pops up. Is there another time limit somewhere that is preventing the processing of large files?

Lil-briX:
I got exectly the same problem. 
in var/log/nginx i found an error log. 
```
2016/06/14 11:57:43 [error] 602#0: *1761 client intended to send too large body: 9886664 bytes, client: 10.0.54.94, server: , request: "POST /api/assets HTTP/1.1", host: "test-raspberry-pi", referrer: "https://test-raspberry-pi/"
2016/06/14 11:57:55 [error] 601#0: *1764 client intended to send too large body: 2340451 bytes, client: 10.0.54.94, server: , request: "POST /api/assets HTTP/1.1", host: "test-raspberry-pi", referrer: "https://test-raspberry-pi/"
2016/06/14 12:11:16 [error] 602#0: *1880 client intended to send too large body: 581950138 bytes, client: 10.0.66.109, server: , request: "POST /api/assets HTTP/1.1", host: "test-raspberry-pi", referrer: "https://test-raspberry-pi/"
2016/06/14 12:22:08 [error] 600#0: *1881 client intended to send too large body: 581950169 bytes, client: 10.0.66.109, server: , request: "POST /api/assets HTTP/1.1", host: "test-raspberry-pi", referrer: "https://test-raspberry-pi/"
2016/06/14 12:24:45 [notice] 14571#0: signal process started
2016/06/14 12:29:48 [error] 14575#0: *1914 upstream timed out (110: Connection timed out) while reading response header from upstream, client: 10.0.66.109, server: , request: "POST /api/assets HTTP/1.1", upstream: "http://127.0.0.1:8080/api/assets", host: "test-raspberry-pi", referrer: "https://test-raspberry-pi/"
2016/06/14 13:07:00 [notice] 19277#0: using inherited sockets from "6;7;"
[...]

PR #489, opened 2016-09-18
Increase server timeout

anayrat:
On raspberrypi upload of a big file can take a lot of time.

---
This change is [Reviewable](https://reviewable.io/reviews/wireload/screenly-ose/489)


vpetersson:
Thanks for the effort. However, this is not the proper solution, it's merely a bandaid. We @over64 is working on a proper fix here #444.

PR #496, opened 2016-10-13
web tests with splinter & FF

over64:


.travis.yml
```diff
@@ -13,7 +14,14 @@ before_script:
   - mkdir -p ~/.screenly ~/.config/uzbl/
   - cp ansible/roles/screenly/files/uzbl-config ~/.config/uzbl/config-screenly
   - cp ansible/roles/screenly/files/screenly.conf ~/.screenly/
+  - cp ansible/roles/screenly/files/screenly.db ~/.screenly/
   - echo -e "[local]\nlocalhost ansible_connection=local" > ansible/localhost
+  - curl http://orig04.deviantart.net/7f89/f/2016/173/7/0/ussr_tyan_by_kawaielli-da76ghz.jpg > /tmp/image.jpg
```
vpetersson: Let's not rely on assets that we do not control ourselves. Instead, we should host it ourselves like we do with the trailer.

over64:
It's so not obviously, but problem was not in travis! Thats too funny, but problem in Selenium. Selenium 3 was released at day when build was broken! We are using splinter 0.7.4 which uses selenium >= 2.6.53! And selenium 3 does not supports old firefox 38... Generally it's all about reproducable builds...And that in 2016... sad but true

lib/utils.py
```diff
@@ -20,6 +23,12 @@
 except:
     pass
 
+# This will work on x86-based machines
+try:
```
vpetersson: Since this is only for x86, make it conditional instead.

server.py
```diff
@@ -297,7 +297,11 @@ def settings_page():
 
 @route('/system_info')
 def system_info():
-    viewlog = check_output(['sudo', 'systemctl', 'status', 'screenly-viewer.service', '-n', '20']).split('\n')
+    viewlog = None
+    try:
+        viewlog = check_output(['sudo', 'systemctl', 'status', 'screenly-viewer.service', '-n', '20']).split('\n')
```
vpetersson: At some point, we should probably switch over to use the systemd python library to get logs instead.

lib/utils.py
over64: There is incorrect comment. This code will works on any platform if you have mplayer installed, but we perfer omxplayer for arm and mplayer for x86, x86_64

PR #497, opened 2016-10-13
Hotfix/upload timeout

over64:
Using gunicorn threads worker, allows process bulk requests without huge timeouts


over64:
all fixed

PR #505, opened 2016-11-01
Add random on asset with a level on % of display 

forum6691:
Would it be possible to add this function:
When I set shuffle display for the configuration of the assets, they are displayed in random. That's fine, but I 'd like some asset to appear more or less than other.
It could be achieve in adding a percentage parameter for the asset.
If the % is 100 %, the asset appears normaly.
If the % is 50 % the asset appears twice less than other asset put at 100 %
if the % is 50 % the asset appears four times less than other asset put at 100 % and so on.

I think it's an interesting function for screenly.


vpetersson:
Please merge in master into this branch first, as it includes a lot of changes that are already in master.

vpetersson:
Actually strike that. This is a diff between production and master. I'm not sure what this is for as I don't see any new code.

PR #510, opened 2016-11-16
Fix problem with large file upload if SSL(via nginx) enabled

over64:


vpetersson:
Do we really need this when we are uploading in chunks?

over64:
@vpetersson, I've tested this today. I've tried to upload 19Mb video via SSL(nginx) and got 'The request was failed'. Also I found error about body size limit in nginx error logs (/var/log/nginx/err.log). Nginx uses `Content-Length` to detect http body size http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

over64:
We could make chunk upload, but we need to implement it at browser-side too. Also chunk upload in Gunicorn is 'implemented', but not tested yet. And about security. With chunk upload we will not control client body size too.
    For best security we could separate url's for file upload and for other part of API

vpetersson:
Ok, can you write a new ticket about implementing it browser side? We definitely need that. I thought it was already implemented (my bad).

over64:
https://github.com/wireload/screenly-ose/issues/512

PR #514, opened 2016-11-23
update apt repo before install nginx

over64:


bin/enable_ssl.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -ex
 
 cd ~/screenly/ansible
+sudo apt-get update
```
vpetersson: Let's do this natively in Ansible instead with:

```
  apt:
    update_cache: yes
```

PR #525, opened 2016-12-12
escape @ in URI for UZBL

over64:
https://github.com/wireload/screenly-ose/issues/515

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -191,7 +191,12 @@ def browser_url(url, cb=lambda _: True, force=False):
         logging.debug('Already showing %s, reloading it.', current_browser_url)
     else:
         current_browser_url = url
-        browser_send('uri ' + current_browser_url, cb=cb)
+
+        """Uzbl handles full URI format incorrect: scheme://uname:passwd@domain:port/path
+        We need to escape @"""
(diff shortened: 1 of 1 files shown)


vpetersson:
Awesome. Did you end-to-end test this?

sodabrew:
:heart:

sodabrew:
Note that the @ sign is legal in a URL in other positions as a literal character (i.e. `http://foo:bar@example.com/legal@url`), so I would expect that this fix actually _fixes_ the behavior where UZBL was consuming the @ sign in other locations as a variable sigil. Since UZBL variables are not accessible from Screenly, there shouldn't be a down side of preventing their use.

vpetersson:
@sodabrew That's a very good point. @over64 we should split the string at the first `@` and only parse that (if you know what i mean).

sodabrew:
Well, I mean that the previous behavior did allow UZBL to interpret at-signs as variable sigils, but users of Screenly probably would not have known that / would not have had access to set those variables, so this change actually fixes the ability to put an at-sign into a URL for a Screenly Asset.

over64:
@sodabrew, @vpetersson thanks. Fixed in https://github.com/wireload/screenly-ose/pull/526

Issue #515, opened 2016-11-23
Add support for assets with basic auth


vpetersson:
Contrary to what we previously thought, UZBL does indeed support Basic Auth. It's just supported through a somewhat odd format where you need to escape the `@`  (i.e. `http://username:password\@domain:port/path`. Kudos to @sodabrew for finding out about this. More information about this can be found [here](https://github.com/uzbl/uzbl/issues/311#issuecomment-249572291).

In order to add this support, we need to add custom handling for this. 

* In the HEAD check (i.e. python request), we might need to parse the credentials and pass them on as follows: `requests.head('https://domain/path', auth=HTTPBasicAuth('username', 'password')`
* In the UZBL call, we need to modify the `uri` we pass on to UZBL as described above (i.e. `http://username:password\@domain:port/path`).

over64:
Where should we get username and password?

vpetersson:
We'd parse it from the URL string, just like most browsers do.

I.e., you'd add an asset with the URL `http://username:password@domain:port` and we'd parse it from there.

vpetersson:
@over64 I've reopened this issue. @sodabrew raised a valid point in #526.

farhangithub27:
> We'd parse it from the URL string, just like most browsers do.
> 
> I.e., you'd add an asset with the URL `http://username:password@domain:port` and we'd parse it from there.

Hi @vpetersson , I'm trying to configure web asset with URL  format as `http://username:password@domain:port` 
following cases I have used example. I'm trying to land at the home page of my reporting server.

- http://myusername:mypassword@myhost.mydomain:myport/reportingserver/flow.html?_flowid=searchFlow
This url is not getting accepted by screenly. Throws out error message 
 **Server Error: Could not retrieve file. Check the asset URL**.I guess uzbl doesn't accept _flowid parameter?
- http://myusername:mypassword@myhost.mydomain:myport/reportingserver/flow.html
This one doesn't get accepted by screenly throws out same error message as above.
- http://myusername:mypassword@myhost.mydomain:myport/reportingserver/login.html
This one gets accepted but nothing gets displayed.
- http://myhost.mydomain:myport/reportingserver/login.html
This url gets accepted and login screen gets displayed but then again I cannot use keyboard.

FYI a get request in POSTMAN and firefox  browser works fine with standard get request that lands me to the home page of the server. Postman request below is
[...]

PR #526, opened 2016-12-13
escape only first @ in URI for UZBL

over64:


over64:
https://github.com/wireload/screenly-ose/pull/525

sodabrew:
:-1: this is bad. It will work for the `username:password@example.com` case but cause confusion in this case:
`http://example.com/weird/@strange/@url`
becomes:
`http://example.com/weird/\@strange/@url`

vpetersson:
@sodabrew yes, you raise a valid point. I've re-opened #515.

vpetersson:
Actually, upon further research, @over64's fix is indeed correct. `@` is a [reserved character](https://stackoverflow.com/questions/1856785/characters-allowed-in-a-url). As such, it is not a legal character in the URL.

@over64 can you redo this PR so that I can accept it again?

sodabrew:
It's not reserved, it's legal in a URL https://tools.ietf.org/html/rfc3986

It has special meaning to the left of the hostname, but no special meaning to the right of the hostname.

For example, this is a legal link: https://www.wikipedia.org/wiki/@

vpetersson:
Interesting. In section 2.2.  (Reserved Characters),  it says `reserved = gen-delims / sub-delims`, and then goes on to define `gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"`. That said, in typical RFC fashion, reserved character are may or may not be reserved. :)

In any case, yes, you're right. It's obviously used in some edge cases. 

We're not using any variables in UZBL, so I guess b) is a better option. 

@over64 you agree?

Issue #546, opened 2017-02-23
Screenly OSE Active Asset Not Displaying


rono711:
Dear All,
I am 100% new in Screenly. I added asset and turned it ON. It's under Active Status. However, nothing is on my big screen? Any advice or tip would be much appreciated. 
![test](https://cloud.githubusercontent.com/assets/25971546/23244537/1e9d1ffe-f9da-11e6-9610-0c84ddf3e8f4.png)

zaphod2010:
I had the same issue a while back., Turned out that the Raspberry's clock was way off.
I added this as a root cronjob:

`@reboot /bin/sleep 30 ; /bin/date -s "$(/usr/bin/curl -s --head http://google.com | /bin/grep ^Date: | /bin/sed 's/Date: //g')"`

vpetersson:
Just to chime in here. Normally, the `ntp` service should be responsible for setting the clock. As such, the root cause is likely either that NTP service is incorrectly configured, or NTP is blocked by the firewall. Since the Raspberry Pi doesn't have any RTC, NTP is required.

If you are unable to use NTP, I suggest you look into [HTP](http://www.vervest.org/htp/). I will close this issue as this is not related to Screenly per se, but rather a network issue.

PR #555, opened 2017-04-04
Time setting on Raspberry after screenly loading

antonmolodykh:
for #546

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -6,11 +6,13 @@ from os import path, getenv, utime
 from platform import machine
 from random import shuffle
 from requests import get as req_get
-from time import sleep
+from time import sleep, ctime
 from json import load as json_load
 from signal import signal, SIGUSR1, SIGUSR2
 import logging
 import sh
+import subprocess
+import ntplib
 
 from settings import settings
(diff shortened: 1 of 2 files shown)


viewer.py
```diff
@@ -315,6 +317,12 @@ def setup():
 
     signal(SIGUSR1, sigusr1)
     signal(SIGUSR2, sigusr2)
+    
+    c = ntplib.NTPClient() 
+    responce = c.request('europe.pool.ntp.org', version=3) 
+    cur_date = ctime(responce.tx_time) 
+    sudodate = subprocess.Popen(["sudo", "date", "-s", cur_date])
```
vpetersson: This won't work unfortunately because we cannot assume we can run `sudo` without a password.

vpetersson:
@antonmolodykh I'm not too keen on this approach. Screenly shouldn't communicate with the NTP servers. That's the job of the ntp daemon. We could perhaps add this configuration to Ansible.

antonmolodykh:
@vpetersson Ok, I will try

antonmolodykh:
This PR is not actual

Issue #531, opened 2016-12-22
DNS not work after reboot


scyhe:
After reboot screenly dns servers from resolv.conf are deleted. I have line : dns-nameservers 8.8.8.8 in /etc/network/interfaces and line dns=8.8.8.8,8.8.4.4 in /boot/network.ini, but everytime when reboot RaspberryPI DNS not working, file resolv.conf is empty.  Version OSE : 2016-12-01

vpetersson:
Hmm it shouldn't be empty. However if you're using a custom DNS, you need to follow [these](https://support.screenly.io/hc/en-us/articles/212107026-Can-I-use-a-custom-DNS-as-well-as-DHCP-with-Screenly-) instructions or simply disable the Screenly network configuration tool.

scyhe:
I have this lines in /boot/network.ini 👍 
```
[eth0]
mode=static
ip=10.0.0.100
netmask=255.255.255.0
gateway=10.0.0.1

[generic]
dns=8.8.8.8,8.8.4.4
```

vpetersson:
Can you cat `/etc/network/interfaces` just after a reboot?

scyhe:
```

pi@raspberrypi:~ $ cat /etc/network/interfaces
# Generated by Screenly Network Manager
auto lo
  iface lo inet loopback

auto eth0
  iface eth0 inet static
  address 10.20.0.199
  netmask 255.255.255.0
  gateway 10.20.0.254
  dns-nameservers 8.8.8.8

```

vpetersson:
The interfaces file does look correct. Strange.

vpetersson:
Ahhh. I see what's going on here. There were changes upstream that switched to using `resolvconf
`, so the changes that `screenly_net_mgr.py` are being overwritten. Will need to investigate how to fix this best.

vpetersson:
Fixed in #557. Sorry for the delay sorting this one out.

PR #557, opened 2017-04-11
Refs #531. Moves to resolvconf based DNS config

vpetersson:
* Raspbian recently moved to `resolveconf` for DNS config. This means that any changes made to `/etc/resolv.conf` will be overwritten. To remedy this, we need to instead use `dns-nameservers` in the interfaces file
* Removes quoting of wpa psk and ssid (as it’s not needed and can cause issues).
* Adds default configuration for eth0 with DHCP if no config is present.


vpetersson:
@antonmolodykh Please take a look at this diff and let me know see any issues. Always good with a 2nd pair of eyes.

antonmolodykh:
@vpetersson I don't see problems

PR #560, opened 2017-04-17
exception handling

antonmolodykh:
Added exception handling during video playback, as unforeseen situations are possible. 
For example, on the hosting with video the limit of requests is established. In this case, sometimes the following happens: asset is passed check in url_fails function, but when trying to play the video the request is rejected, which causes viewer to crash.


vpetersson:
@antonmolodykh we should probably not just do a try/except, but rather try capturing the exact issue.

antonmolodykh:
@vpetersson ok, now we catch only the error which related with issue

vpetersson:
LGTM

PR #561, opened 2017-04-29
Upload multiple images

antonmolodykh:
Added the ability to upload multiple assets.
Now the progress bar displays the progress of the entire load, but not every file. You need to decide whether you need a progress bar for each file, and if so, decide how it should look.


vpetersson:
@antonmolodykh Looks like some tests are failing. We need to fix that before we can land.

vpetersson:
@antonmolodykh it might be my changes that broke some of that in the PEP section. Sorry about that, but please go ahead and solve those. Shouldn't take many seconds.

PR #563, opened 2017-05-15
add backup option

antonmolodykh:


lib/backup_helper.py
```diff
@@ -0,0 +1,44 @@
...
+        for directory in directories:
+            if directory not in tar.getnames():
+                raise Exception("Archive is wrong.")
+
+        sh.sudo('service', 'screenly-viewer', 'stop')
+        for directory in directories:
+            rmtree(path.join(home, directory))
+
+        tar.extractall(path=path.join(home))
+        sh.sudo('service', 'screenly-viewer', 'start')
```
vpetersson: We can't do this as we cannot assume there is no password set. Instead, we need to do something like this:

 * Create a new file in Ansible called something like this `/usr/local/bin/screenly_util.sh`
 * Use argparse or similar to make allow it to take a variable like `screenly_util.sh backup /path/to/backup.tgz`
 * Configure `/etc/sudoers` to allow the user `pi` to execute this command with `sudo` passwordless

antonmolodykh: reasonable, okay

ckrzen: Sudo w/o passwd example from my ScreenBacklight script @ https://help.ubuntu.com/community/ScreenBacklight:

`
$ sudo visudo -f /etc/sudoers.d/myOverrides
`

Then copy/paste the following script into your new file and customize it:

`
%sudo	ALL=NOPASSWD:/usr/local/bin/sysbacklight
`

Make the script read-only:

`
$ sudo chmod 440 /etc/sudoers.d/myOverrides
`

Hope this helps.

vpetersson: @ckrzen Yup that syntax looks about right. We do however need to manage it using ansible.

ansible/roles/screenly/files/screenly_overrides
```diff
@@ -0,0 +1 @@
+%sudo ALL=NOPASSWD:/usr/local/bin/screenly_utils.sh
```
vpetersson: Strictly speaking, this should be the user `pi`.

ansible/roles/screenly/tasks/main.yml
```diff
@@ -49,6 +49,21 @@
 
 - debug: msg="{{ migrate.stdout }}"
 
+- name: Copy screenly_utils.sh
+  copy:
```
vpetersson: You need to set `mode: 0755`. Otherwise it cannot be executed.

ansible/roles/screenly/files/screenly_utils.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+recover(){
+	sudo service screenly-viewer stop
+	rm -rf /home/pi/.screenly /home/pi/screenly_assets
+	tar -xvf "$1" -C /home/pi/
```
vpetersson: Before starting the recovery, we need to:
 * Verify that $1 is a string
 * The file exists

vpetersson: Also, please use soft tabs (4 spaces).

ansible/roles/screenly/files/screenly_utils.sh
```diff
@@ -0,0 +1,24 @@
...
+			if [ -n "$2" ]; then
+				recover "$2"
+				shift
+				exit 0
+			else
+				exit 1
+			fi
+			;;
+		*)
+			exit 1
```
vpetersson: Add fallback error message.

lib/backup_helper.py
```diff
@@ -0,0 +1,36 @@
...
+from os import path, getenv, remove
+import sh
+
+directories = ['.screenly', 'screenly_assets']
+archive_name = "screenly-backup.tar.gz"
+
+
+def create_backup():
+    home = getenv('HOME')
+    file_path = path.join('static', archive_name)
```
vpetersson: This should probably be an absolute path.

server.py
```diff
@@ -270,6 +271,31 @@ def playlist_order():
...
+
+@route('/api/recover', method="POST")
+@auth_basic
+@api
+def recover():
+    req = Request(request.environ)
+    file_upload = (req.files['backup_upload'])
+    filename = file_upload.filename
+
+    if not filename.endswith('.tar.gz'):
```
vpetersson: It's probably better to use the mimetype than merely the file extension.

ansible/roles/screenly/files/screenly_utils.sh
antonmolodykh: Hm. The python part ensures that the file exists and the name is correct. I do not see the point of doing repeated checks. Am I mistaken?

vpetersson: Yeah, I noticed. However, given the magnitude of what we're doing here, it doesn't hurt with an extra level of security in case someone is trying to use the script directly.

ansible/roles/screenly/files/screenly_utils.sh
```diff
@@ -0,0 +1,26 @@
...
+
+while :; do
+    case $1 in
+        recover)
+            if [ -f "$2" ]; then
+                recover "$2"
+                shift
+                exit 0
+            else
+                echo "File is not exist"
```
vpetersson: The backup file does not exist.

pifiu:
So is there a backup option now?

vpetersson:
Experimental support in the development branch, yes.

PR #572, opened 2017-06-12
Add support for live streaming

### LINKED ISSUE #90: Add support for live streaming
As far as I could see the current implementation for a video stream is "start it and wait until omxplayer is finished", however I think it would be useful to optionally specify a duration, in the case of continuous running livestreams:

Switch to the stream for 5 min, then go on to the next asset. As far as I could see omxplayer does not support this directly, but I guess spawning a shell process and doing a "sleep x; kill <omx_pid>" might do the trick just fine (killall omxplayer probably not a good idea, if omxplayer finishes, crashes before the timeout is reached you might accidentally kill another instance already playing the next asset.).

Maybe I will come around to do this myself later this month, however I just thought of putting it here, in case somebody picks it up before me :)

antonmolodykh:
Resolves #90.

### DIFF
--- a/ansible/roles/screenly/files/screenly.conf
+++ b/ansible/roles/screenly/files/screenly.conf
@@ -15,7 +15,10 @@ audio_output = hdmi
 ; Shuffle playlist before playing (otherwise sorted by name)
 shuffle_playlist = False
 
-; Asset duration default value
+; Stream asset duration default value (seconds)
+default_streaming_duration = 300
+
+; Asset duration default value (seconds)
 default_duration = 10
 
 ; Debug logging toggle
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -33,7 +33,6 @@ if machine() in ['x86', 'x86_64']:
 
 def validate_url(string):
     """Simple URL verification.
-
     >>> validate_url("hello")
     False
     >>> validate_url("ftp://example.com")
@@ -44,11 +43,10 @@ def validate_url(string):
     True
     >>> validate_url("https://wireload.net/logo.png")
     True
-
     """
 
     checker = urlparse(string)
-    return bool(checker.scheme in ('http', 'https') and checker.netloc)
+    return bool(checker.scheme in ('http', 'https', 'rtsp', 'rtmp') and checker.netloc)
 
 
 def get_node_ip():
@@ -111,6 +109,23 @@ def json_dump(obj):
 
 
 def url_fails(url):
+    """
+    If it is streaming
+    """
+    if urlparse(url).scheme in ('rtsp', 'rtmp'):
+        if arch in ('armv6l', 'armv7l'):
+            run_omxplayer = omxplayer(url, info=True, _err_to_out=True, _ok_code=[0, 1])
+            for line in run_omxplayer.split('\n'):
+                if 'Input #0' in line:
+                    return False
+            return True
+        else:
+            run_mplayer = mplayer('-identify', '-frames', '0', '-nosound', url)
+            for line in run_mplayer.split('\n'):
+                if 'Clip info:' in line:
+                    return False
+            return True
+
     """
     Try HEAD and GET for URL availability check.
     """
--- a/static/js/screenly-ose.coffee
+++ b/static/js/screenly-ose.coffee
@@ -65,7 +65,7 @@ viduris   = ('rtsp rtmp'.split ' ')
 get_mimetype = (filename) =>
   scheme = (_.first filename.split ':').toLowerCase()
   match = scheme in viduris
-  if match then return 'video'
+  if match then return 'streaming'
   ext = (_.last filename.split '.').toLowerCase()
   mt = _.find mimetypes, (mt) -> ext in mt[0]
   if mt then mt[1] else null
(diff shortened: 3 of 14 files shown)


ansible/roles/screenly/files/screenly.conf
```diff
@@ -15,6 +15,9 @@ audio_output = hdmi
 ; Shuffle playlist before playing (otherwise sorted by name)
 shuffle_playlist = False
 
+; Stream asset duration default value
+default_streaming_duration = 300
```
vpetersson: What unit is this? Seconds? Minutes? Hours?

lib/utils.py
```diff
@@ -112,6 +110,23 @@ def json_dump(obj):
 
 def url_fails(url):
     """
+    If it is streaming
+    """
+    if urlparse(url).scheme in ('rtsp', 'rtmp'):
+        if arch in ('armv6l', 'armv7l'):
+            run_omxplayer = omxplayer(url, info=True, _err_to_out=True, _ok_code=[0, 1])
+            for line in run_omxplayer.split('\n'):
+                if 'Input #0' in line:
```
vpetersson: That seems a bit fragile. I guess we can't use exit codes instead?

tests/splinter_test.py
```diff
@@ -204,6 +204,33 @@ def test_add_two_assets_upload(self):
...
             self.assertEqual(assets[1]['duration'], settings['default_duration'])
 
+    def test_add_asset_streaming(self):
+        with Browser() as browser:
+            browser.visit(main_page_url)
+
+            wait_for_and_do(browser, '#add-asset-button', lambda btn: btn.click())
+            sleep(1)
+
+            wait_for_and_do(browser, 'input[name="uri"]', lambda field: field.fill('rtmp://controlstreams.com:1935/vtvmoviles3/vtvmoviles3'))
```
vpetersson: We shouldn't use external URLs for unittests. It is prone to break. Instead,  can you simulate this?

lib/utils.py
antonmolodykh: no. This refers to the problem of omxplayer: when we try to get information, we always get exit code 1

ansible/roles/screenly/files/screenly.conf
antonmolodykh: Seconds, the same format as for default_duration. something is wrong?

vpetersson: Yeah, I assumed it was in seconds, but let's just add that for clarity. Just add "(seconds)" to the end of the sentence. Do the same for `default_duration`.

lib/utils.py
vpetersson: Yeah, that's what I thought. Oh well, in that case, let's do it this way. Not idea, but it is what it is.

tests/splinter_test.py
antonmolodykh: Yes, I agree with that. But if I understand correctly, then we need to use something like vlc or gstreamer. Do we want this?

vpetersson: @aljungberg How do you reckon we should test this?

static/js/screenly-ose.coffee
```diff
@@ -232,7 +234,12 @@ API.View.AddAssetView = class AddAssetView extends Backbone.View
   updateFileUploadMimetype: (filename) => @updateMimetype filename
   updateMimetype: (filename) =>
     mt = get_mimetype filename
+    if mt
+      console.log "MT!"
```
aljungberg: Is this debug logging? I think we don't need to commit that, if someone needs it they can add it when they do.

static/js/screenly-ose.coffee
```diff
@@ -232,7 +234,12 @@ API.View.AddAssetView = class AddAssetView extends Backbone.View
...
     @$fv 'mimetype', mt if mt
+    s = @$fv 'mimetype'
+    console.log "mt " + mt
+    console.log s
```
aljungberg: Same here.

static/js/screenly-ose.coffee
```diff
@@ -413,6 +420,7 @@ API.View.AssetRowView = class AssetRowView extends Backbone.View
     (@$ ".toggle input").prop "checked", @model.get 'is_enabled'
     (@$ ".asset-icon").addClass switch @model.get "mimetype"
       when "video"   then "icon-facetime-video"
+      when "streaming"   then "icon-facetime-video"
```
aljungberg: The reason there are x spaces before "then" is that someone wanted to align all the `then`s vertically. So either align 'em all or don't align them all, but either way this`then` won't have more than 1 space before it.

tests/splinter_test.py
aljungberg: Maybe we should install https://github.com/theintencity/rtmplite and spin that up with some static material in the unit test setup? Hopefully it's pretty easy to get going. Then we can kill it after teardown.

tests/splinter_test.py
```diff
@@ -204,6 +204,33 @@ def test_add_two_assets_upload(self):
...
+            sleep(1)
+
+            wait_for_and_do(browser, '#add-form', lambda form: form.click())
+            sleep(1)
```
aljungberg: I guess this is not something we can work on right now but we really should update `wait_for_and_do` to micro sleep until the element is available (or there's a timeout) so we can remove all static `sleep` (which slow down the unit tests a lot more than necessary).

antonmolodykh:
Hmm. It seems to me that the rtmp server did not start in travis. All tests are passed on my computer

Issue #485, opened 2016-09-01
 "Schedule Overview" title


thefyfy:
Hi,
Is it possible to have a text box in "Settings" to change the default title "Schedule Overview" ? 
When we have different screens with screenly, I think it would be great if we can identitfy quickly which screen you are changing with a custom title.
What do you think about it?
In any case, thanks for your amazing job!

vpetersson:
> In any case, thanks for your amazing job!

Thank you. 

Closing out as this is a duplicate of #349.

Issue #349, opened 2016-01-12
title for index.haml


GeoHolz:
Hello,

Is it possible to add the hostname of raspberry in HTML title ?

I try :
%title Screnly OSE server.hostname

But doesn't works

vpetersson:
Yeah, that wouldn't work. You'd need to pass it on as from `server.py` as a variable similar to `loadavg` in `system_info.haml`

PR #594, opened 2017-07-23
Added a setting for player name

daniep01:
Added a setting player_name which is displayed at the top of the schedule page and in html <title> to help identify which screen you are editing when multiple boxes are in use.

### DIFF
--- a/server.py
+++ b/server.py
@@ -303,7 +303,8 @@ def recover():
 @route('/')
 @auth_basic
 def viewIndex():
-    return template('index')
+    system_name=settings['system_name']
+    return template('index', system_name=system_name)
 
 
 @route('/settings', method=["GET", "POST"])
--- a/server.py
+++ b/server.py
@@ -303,7 +303,7 @@ def recover():
 @route('/')
 @auth_basic
 def viewIndex():
-    player_name=settings['player_name']
+    player_name = settings['player_name']
     return template('index', player_name = player_name)
 
 
--- a/server.py
+++ b/server.py
@@ -304,7 +304,7 @@ def recover():
 @auth_basic
 def viewIndex():
     player_name = settings['player_name']
-    return template('index', player_name = player_name)
+    return template('index', player_name=player_name)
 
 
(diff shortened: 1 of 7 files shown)


vpetersson:
@daniep01 - looks sensible. However:

 * Let's change it to 'Player Name' instead of system name
 * The linting failed as you can see in the Travis run, so let's fix that too.

server.py
```diff
@@ -303,7 +303,8 @@ def recover():
 @route('/')
 @auth_basic
 def viewIndex():
-    return template('index')
+    system_name=settings['system_name']
```
vpetersson: What happens if this setting is absent?

daniep01: It has a default in settings.py so should be always set

daniep01:
The travis error line is
socket.error: [Errno 98] Address already in use
I'm not sure how to resolve this - doesn't seem to be related to this change.

vpetersson:
@daniep01 the issue is not that (that one is suppressed). The issue is:

```
0.85s$ find . ! -path "*/rtmplite/*" -name \*.py -exec pep8 --ignore=E402,E501,E731 {} +
./server.py:306:16: E225 missing whitespace around operator
```

daniep01:
Thanks, white space issue now resolved and travis happy.

daniep01:
As requested in #349  and #485.

vpetersson:
Just one last thing @daniep01 - what happens if the setting is absent in `setting.ini`? Since there is no migration, it will be absent on all nodes who are upgrading.

daniep01:
I can't find `setting.ini` in the code, do you mean `screenly.conf`?

vpetersson:
@daniep01 yes - sorry meant `screenly.conf`. In theory, it should use the default, but let's double check.

vpetersson:
Cool. Thanks!

vpetersson:
@antonmolodykh could you just take a quick look at this as well so that we have a second pair of eyes.

antonmolodykh:
LGTM

Issue #586, opened 2017-07-01
Add streaming support for YouTube


vpetersson:
I want to add experimental support for YouTube using [youtube-dl](https://rg3.github.io/youtube-dl/). This support will not be rocket solid, as there is a lot of things that could go wrong. 

The workflow however would be something like this:

* The user adds a YouTube URL. 
* We detect that this is a YouTube URL (i.e it uses the domains 'youtu.be' or 'youtube.com`)

Once we've established it is a YouTube URL, we can move on.

First, we can get the title by running `youtube-dl -e https://youtu.be/ABC`

Next, we can download the actual asset:
```
$ youtube-dl -f mp4 -o /home/pi/screenly_assets/[asset_id].mp4 https://youtu.be/ABC
[youtube] Setting language
[youtube] pP4acHxygqk: Downloading webpage
[youtube] pP4acHxygqk: Downloading video info webpage
[youtube] pP4acHxygqk: Extracting video information
[download] Destination: [asset_id].mp4
[download] 100% of 35.14MiB in 00:14
```

Finally, we add it into the rotation as a regular asset.

This will also require that we add `youtube-dl` as a dependency through `pip` to Screenly in Ansible.

PR #596, opened 2017-07-24
Support for YouTube

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/586


requirements.txt
```diff
@@ -20,3 +20,8 @@ uptime==2.0.2
 werkzeug==0.10.4
 wsgiref==0.1.2
 futures==3.0.5
+youtube_dl==2017.7.2
```
vpetersson: We should probably not pin this one. At the very least use `>=`.

server.py
```diff
@@ -151,12 +153,20 @@ def get(key):
...
+        asset['is_processing'] = 1
+
     asset['uri'] = uri
 
     if "video" in asset['mimetype']:
-        video_duration = get_video_duration(asset['uri'])
-        if video_duration:
-            asset['duration'] = int(video_duration.total_seconds())
+        if asset['is_processing'] == 0:
+            video_duration = get_video_duration(uri)
```
vpetersson: Don't we get this from `youtube_dl`? It's a pretty heavy operation on the Pi, so we don't want to do it unless we need to.

antonmolodykh: We can get information about the duration of the video from youtube_dl, but it does not make much sense: after downloading a video from YouTube, we work with the asset as with a regular video asset (https://github.com/Screenly/screenly-ose/commit/c07d82319cff7b7bfbcafe607c871a378d98c456#diff-d355af669766ee3562643e8e5034f715R158). This means that when editing an asset, for example, changing the date - in any case, `get_video_duration` will be called.

vpetersson: But the duration is immutable. It should never change. The only video asset type it could change for would be for RTSP assets.

antonmolodykh: I understand it. But now this is how the processing of any video asset works: when adding or editing, the `get_video_duratio`n function works. I just made a video from YouTube a simple video asset.

vpetersson: Alright, please create a ticket about that then.

Issue #581, opened 2017-06-29
Remove unused assets


vpetersson:
I was looking around and I found a number of unused images. Let's remove all unused images from the repository.

PR #598, opened 2017-07-25
Remove unused assets

antonmolodykh:
Removed unused resources and renamed wireload-bg.jpeg into screenly-bg.jpeg

Refs https://github.com/Screenly/screenly-ose/issues/581


vpetersson:
I'm not sure `static/img/drag-handle@2x.png` should be deleted if the file `static/img/drag-handle.png` exists.

antonmolodykh:
@vpetersson I checked: we really don't use this picture

vpetersson:
@antonmolodykh hmm should we not fix the `@2x` issue then?

antonmolodykh:
@vpetersson you mean just leave `drag-handle@2x.png` ?

vpetersson:
@petermonte Could you point @antonmolodykh in the right direction here?

petermonte:
@vpetersson and @antonmolodykh,

In fact `drag-handle@2x.png` is not being used [here](https://github.com/Screenly/screenly-ose/blob/87b59815255b9aa1b09f2a327bfd988a181b796e/views/index.haml#L65). But this doesn't mean we should remove it.

What needs to be done is either point css to use the `@2x` under a media query for higher DPI screens or take advantage of the `srcset` attribute for `img` tags. You can find [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Example_3_Using_the_srcset_attribute) some good documentation

antonmolodykh:
@petermonte understand, thank you :)

Issue #585, opened 2017-07-01
Add result of display status to system page


vpetersson:
As discussed in #108, it's possible to toggle the display power using `vcgencmd`. We should add this status to the settings page.

We can get the display power by running:
```
pi@raspberrypi:~ $ vcgencmd  display_power
display_power=1
```

Let's `1` means that the display power is on, and `0` that it is off.  Let's just add a simple text that says "Display Power: True|False"

daniep01:
@vpetersson I'm working on this in branch display_power

PR #601, opened 2017-07-27
Status/display power

daniep01:
#585 
Added display_power status. Not sure if it should be added to the end of the table on the status page or if maybe the table would be better vertically with left/right split of parameter/value.


vpetersson:
Nice one, @daniep01. Could you post a screenshot of the page?

PR #602, opened 2017-07-28
Display power

daniep01:
New pull request replacing #601 which was missing recent merges
Feature #585 

Here's a screenshot showing two possible versions of the info table.
systeminfo

### DIFF
--- a/lib/diagnostics.py
+++ b/lib/diagnostics.py
@@ -64,9 +64,9 @@ def get_display_power():
     try:
         display_status = sh.vcgencmd('display_power').stdout.strip().split('=')
         if display_status[1] == '1':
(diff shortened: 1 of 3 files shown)


vpetersson:
Great! Thanks. Can we just make a minor tweak and change it to 'Display Power: On/Off' instead? That feels like a more natural way to describe the state.

daniep01:
Yes, will change the wording. Any thoughts on the table layout?

vpetersson:
Awesome. Thanks, @daniep01.

@antonmolodykh can you take a look as well as a second pair of eyes?

antonmolodykh:
LGTM

Issue #579, opened 2017-06-22
Improve network status detection


vpetersson:
When there's a networking issue, we currently only throw the error "Unable to resolve management IP."

We should add some more logic for this. Perhaps something like this:

![image](https://user-images.githubusercontent.com/357664/27447284-51f45934-5778-11e7-878e-0a159ea223e3.png)

PR #604, opened 2017-08-01
Improving network status detection

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/579

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -2,7 +2,7 @@ import requests
 import json
 import re
 import certifi
-from netifaces import ifaddresses
+from netifaces import ifaddresses, interfaces
 from sh import grep, netstat
 from subprocess import check_output, call
 from urlparse import urlparse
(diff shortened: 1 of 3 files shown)


lib/utils.py
```diff
@@ -56,18 +56,33 @@ def validate_url(string):
...
-        my_ip = ifaddresses(default_interface)[2][0]['addr']
-        return my_ip
-    except:
-        pass
+        if 'eth0' in active_interfaces:
+            interface = 'eth0'
+        elif 'wlan0' in active_interfaces:
+            interface = 'wlan0'
+        else:
+            raise Exception("Display error saying no active network interface found")
```
vpetersson: 'No active network interface found.'

lib/utils.py
```diff
@@ -56,18 +56,33 @@ def validate_url(string):
...
+
+        try:
+            my_ip = ifaddresses(interface)[2][0]['addr']
+            return my_ip
+        except KeyError:
+            raise Exception("Display error saying that we were unable to retrieve an IP")
```
vpetersson: 'Unable to retrieve an IP.'

lib/utils.py
```diff
@@ -56,18 +56,33 @@ def validate_url(string):
...
+        """Returns the node's IP, for the interface
+        that is being used as the default gateway.
+        This shuld work on both MacOS X and Linux."""
+
+        try:
+            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
+            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            return my_ip
+        except:
+            raise Exception("Unable to look up your installation's IP address.")
```
vpetersson: 'Unable to resolve local IP address.'

server.py
```diff
@@ -382,8 +382,12 @@ def system_info():
 
 @route('/splash_page')
 def splash_page():
-    my_ip = get_node_ip()
-    if my_ip:
+    try:
+        my_ip = get_node_ip()
+    except Exception as e:
+        ip_lookup = False
+        url = e
```
vpetersson: `url` seems to be an odd choice for this variable. How about `error_msg` or something like that instead?

lib/utils.py
antonmolodykh: Oh, my mistake. I was in a hurry and just copied

Issue #599, opened 2017-07-25
Localise date format


daniep01:
Hi,
It would be better if the displayed date format in the schedule could be localised (dd/mm/yyyy) or (mm/dd/yyyy) based on a user setting. I'm happy to work on this but would like some pointers from more experienced contributors - 
* is there a reason this hasn't/can't be done
* where in the code to start looking

Thanks.

vpetersson:
@daniep01 If I'm not mistaken, there should already be support for this. I guess the main consideration is that we render them to native dates in the backend.

vpetersson:
@antonmolodykh do you have any pointers here?

antonmolodykh:
If we are talking about a new parameter in the settings, then it's not difficult:
1. you need to add a new parameter in `settings.py` and `screenly.conf`
2. Pass this parameter to the template (function `template` in `server.py`)
3. Add a new parameter to the `settings.haml` template
3. In `screenly-ose.coffee` make a condition which in the format will be selected according to the passed parameter.

If I'm not mistaken, now we don't have the ability to detect a location

vpetersson:
@antonmolodykh Shouldn't we be able to detect that with JS [relatively easy](https://stackoverflow.com/questions/673905/best-way-to-determine-users-locale-within-browser)?

antonmolodykh:
@vpetersson As I understand it is about parsing the "Accept-Language" header. If I understand correctly, then parsing involves processing on the server side, and even in this case it's just an attempt to find out the locale, nothing more. We still have to provide a choice in case of an error. It seems to me easier to provide a choice of format through settings

vpetersson:
Perhaps we can save ourselves the effort and use [this feature](https://bottle-utils.readthedocs.io/en/latest/i18n.html) in Bottle?

antonmolodykh:
If I understand correctly, this module provides translation assistance, I'm not sure that I understand how this can help us. In any case, the essence of this task is not related to the language: for example, in the US and in the UK different date formats are used.

vpetersson:
I just saw this piece:

```
<p>{{ _('Current time') }}: {{ time }}</p>
```

I *think* that provides the localized time, but i could be wrong.

antonmolodykh:
It seems to me to detected a local on the server side is not a good idea. If I understand correctly, for users who downloaded the Screenly image and did not configure the locale on their own, the data will not be correct. Am I mistaken?

vpetersson:
[...]

PR #607, opened 2017-08-03
Date time locale setting

daniep01:
#599 
First attempt, doesn't automatically detect locale yet.


vpetersson:
@daniep01 interesting approach. The issue with reading the locale from the system (rather than the browser), is that it will often time be wrong. For instance, if someone installs Screenly using NOOBS, chances are they will not bother updating the locale on the system. 

As such, I think it makes more sense to read it from the browser using some kind of JS framework.

daniep01:
That would be ideal but it seems to be a bit of a minefield. It looks like there is no consistent way that is guaranteed to work across browsers.
[https://zzz.buzz/2016/01/13/detect-browser-language-in-javascript/](url)
Do you know of any specific bit of framework that we could use?

vpetersson:
@petermonte Do you have any recommendations here?

petermonte:
@vpetersson and @daniep01,

I would recommend this **[answer](https://stackoverflow.com/questions/10659523/how-to-get-the-exact-local-time-of-client#answer-39161134)** on **Stackoverflow**. 

While working in several apps that make use of local time I've found **[moment.js](https://momentjs.com/)** to be a robust JS framework. It holds several methods to use and manipulate time and also includes a complete library for locale strings and time format.

Now this goes without saying that using the browser time will always be dependent of the system time, which by the way is completely vulnerable to user manipulation. Once I change my system time my web app suffers the same change.

To really get a precise time I would say that the use of a **[server](https://time.is/)** API holding time combined with bowser time interval would be the case. Check **[time.is](https://time.is/)** and change your local system time. You will see a moment where the time changes in the browser but in a matter of a second it gets corrected.

Hope this help.

antonmolodykh:
Given @petermonte comment and the fact that Screenly should also work without connection, in my opinion unfortunately this PR now does not make sense

PR #610, opened 2017-08-04
Add support for WiFi passwords with the '%' character

mbardelmeijer:
As the default `ConfigParser` does interpolation with the `%` character, WiFi password etc. with this character won't work. This ensures these fields are parsed as-is.


vpetersson:
Interesting find, @mbardelmeijer. LGTM. Accepting.

PR #614, opened 2017-08-10
Fix network status detection

antonmolodykh:
Refs #579

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -2,7 +2,7 @@ import requests
 import json
 import re
 import certifi
-from netifaces import ifaddresses, interfaces
+from netifaces import ifaddresses
 from sh import grep, netstat
 from subprocess import check_output, call
 from urlparse import urlparse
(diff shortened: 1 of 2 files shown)


lib/utils.py
```diff
@@ -57,14 +57,24 @@ def validate_url(string):
...
-        active_interfaces = interfaces()
 
-        if 'eth0' in active_interfaces:
-            interface = 'eth0'
-        elif 'wlan0' in active_interfaces:
-            interface = 'wlan0'
-        else:
-            raise Exception("No active network interface found.")
+        interface = None
+        for ifc in ['eth0', 'wlan0']:
```
vpetersson: I don't think this works on `wlan0`? It's a physical link, so I don't think this concept is applicable.

antonmolodykh: Tomorrow I will have a new PI and I'll check it out

lib/utils.py
```diff
@@ -57,14 +57,24 @@ def validate_url(string):
...
-            interface = 'wlan0'
-        else:
-            raise Exception("No active network interface found.")
+        interface = None
+        try:
+            file_carrier = open('/sys/class/net/eth0/carrier')
+            file_operstate = open('/sys/class/net/eth0/operstate')
+
+            if "1" in file_carrier.read() and "up" in file_operstate.read():
+                interface = 'eth0'
```
vpetersson: Actually, take a look at this:
https://github.com/Screenly/screenly-ose/blob/master/ansible/roles/network/files/screenly_net_mgr.py#L157-L162

lib/utils.py
```diff
@@ -57,14 +57,30 @@ def validate_url(string):
...
+        if not interface:
+            file_interfaces = open('/etc/network/interfaces')
+            for n in range(10):
+                iface = 'wlan{}'.format(n)
+                if iface in file_interfaces.read():
+                    interface = iface
+                    break
+
+        if not interface:
+            raise Exception("No active network connection found")
```
vpetersson: Add trailing period.

PR #616, opened 2017-08-17
Restart websocket layer

antonmolodykh:
It is necessary that websocket_server_layer starts after installing Screenly


vpetersson:
LGTM

PR #617, opened 2017-08-18
Fix websocket connection in enable_ssl mode

antonmolodykh:


ansible/roles/ssl/tasks/main.yml
```diff
@@ -37,7 +37,7 @@
   copy:
     src: "{{ item }}"
     dest: "/etc/ssl/{{ item }}"
-    mode: 0600
+    mode: 0644
```
vpetersson: Hmm we don't want 0644 for certificates. That's very insecure.

vpetersson:
LGTM

Issue #611, opened 2017-08-08
Migrate to Flask


vpetersson:
We're right now mixing Flask and Bottle code (due to limitations in Bottle). Let's clean this up and migrate everything to Flask.

PR #622, opened 2017-08-31
Migrate to flask

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/611


server.py
```diff
@@ -186,130 +187,124 @@ def get(key):
...
-
-
 # api view decorator. handles errors
-def api(view):
+def api_response(view):
     @wraps(view)
     def api_view(*args, **kwargs):
         try:
-            return make_json_response(view(*args, **kwargs))
-        except HTTPResponse:
```
antonmolodykh: I have a little doubt about this fragment.
For the bottle was this code:
'except HTTPResponse:`
I don't quite understand its purpose and I don't know if it's necessary to do something similar for Flask

Issue #624, opened 2017-09-02
"No active network connection found" on Raspbian stretch


anayrat:
Hello,

I tried a fresh install on rasbian stretch. After the installation I got this error "No active network connection found"

```
sept. 02 23:14:09 raspberrypi python[529]: Traceback (most recent call last):
sept. 02 23:14:09 raspberrypi python[529]:   File "/usr/local/lib/python2.7/dist-packages/bottle.py", line 862, in _handle
sept. 02 23:14:09 raspberrypi python[529]:     return route.call(**args)
sept. 02 23:14:09 raspberrypi python[529]:   File "/usr/local/lib/python2.7/dist-packages/bottle.py", line 1732, in wrapper
sept. 02 23:14:09 raspberrypi python[529]:     rv = callback(*a, **ka)
sept. 02 23:14:09 raspberrypi python[529]:   File "/home/pi/screenly/server.py", line 316, in viewIndex
sept. 02 23:14:09 raspberrypi python[529]:     my_ip = get_node_ip()
sept. 02 23:14:09 raspberrypi python[529]:   File "/home/pi/screenly/lib/utils.py", line 83, in get_node_ip
sept. 02 23:14:09 raspberrypi python[529]:     raise Exception("No active network connection found.")
sept. 02 23:14:09 raspberrypi python[529]: Exception: No active network connection found.
```

My Rpi has network connection with ethernet cable:

```
 ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enxb827eb6c50a7: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether b8:27:eb:6c:50:a7 brd ff:ff:ff:ff:ff:ff
    inet 192.168.1.19/24 brd 192.168.1.255 scope global enxb827eb6c50a7
       valid_lft forever preferred_lft forever
    inet6 2a01:cb14:2e8:900:34a1:8d59:3e8f:5f43/64 scope global dynamic
       valid_lft 852sec preferred_lft 252sec
[...]

PR #631, opened 2017-09-11
Fix network address detection for Stretch

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/624

### DIFF
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -51,11 +51,17 @@
   when: not is_berryboot
 
 - name: Use Systemd as init and quiet boot process
-  lineinfile:
+  replace:
     dest: /boot/cmdline.txt
-    regexp: (.*)
-    backrefs: yes
-    line: \1 quiet init=/lib/systemd/systemd
+    regexp: (^(?!$)((?!quiet init=/lib/systemd/systemd).)*$)
+    replace: \1 quiet init=/lib/systemd/systemd
+  when: not is_berryboot
+
+- name: ethN/wlanN names for interfaces
+  replace:
+    dest: /boot/cmdline.txt
+    regexp: (^(?!$)((?!net\.ifnames=0).)*$)
+    replace: \1 net.ifnames=0
   when: not is_berryboot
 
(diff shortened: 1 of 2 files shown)


ansible/roles/network/tasks/main.yml
```diff
@@ -50,3 +50,9 @@
     minute: 0
     job: /usr/bin/python /usr/sbin/screenly_net_watchdog.py
     state: absent
+
+- name: ethN names for interfaces
```
vpetersson: ethN/wifiN

ansible/roles/network/tasks/main.yml
```diff
@@ -50,3 +50,9 @@
...
+  command: "{{item}}"
```
vpetersson: Check out this:
https://github.com/Screenly/screenly-ose/blob/master/ansible/roles/system/tasks/main.yml#L36-L41

We already have some logic for dealing with `cmdline.txt`, so let's reuse that.

antonmolodykh: hmm. but in this code there is handling for berryboot. or do you mean just place a new task for cmdline.txt in the same file?

vpetersson: No, check the lines before. It's the opposite. It's *only* triggered it's *not* BerryBoot.

ansible/roles/system/tasks/main.yml
```diff
@@ -58,6 +58,14 @@
     line: \1 quiet init=/lib/systemd/systemd
   when: not is_berryboot
 
+- name: ethN/wlanN names for interfaces
+  lineinfile:
+    dest: /boot/cmdline.txt
+    regexp: (.*)
```
vpetersson: Hmm, I don't think this works. Ansible scripts should be immutable. That said, `lineinfile` should work, but I think you can simplify this by simply skipping the regex part.

antonmolodykh: @vpetersson I checked it works. cmdline.txt requires that the rules be placed on one line and separated by a space. for this I use the regexp part (otherwise the rule will be inserted on a new line)

Issue #591, opened 2017-07-07
Add API call for next/previous


vpetersson:
Let's add an API call for going back and forward in the asset list.

PR #637, opened 2017-09-25
Add API call for next/previous

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/591

### DIFF
--- a/server.py
+++ b/server.py
@@ -8,7 +8,7 @@ __license__ = "Dual License: GPLv2 and Commercial License"
 from datetime import timedelta
 from functools import wraps
 from hurry.filesize import size
-from os import path, makedirs, statvfs, mkdir
+from os import path, makedirs, statvfs, mkdir, system
 from sh import git
 import sh
 from subprocess import check_output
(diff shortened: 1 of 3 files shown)


viewer.py
```diff
@@ -59,13 +61,20 @@ def sigusr2(signum, frame):
     load_settings()
 
 
+def sigurg(signum, frame):
+    scheduler.reverse = True
+    logging.info('USR1 received, skipping.')
```
vpetersson: s/USR1/USRG/g

server.py
```diff
@@ -515,12 +515,40 @@ def post(self):
...
+        ],
+        'responses': {
+            '200': {
+                'description': 'Asset switched'
+            }
+        }
+    })
+    def get(self, command):
+        if command == "next":
+            sh.sudo('systemctl', 'kill', '--signal=SIGUSR1', '--kill-who=main', 'screenly-viewer.service')
```
vpetersson: No sudo needed.

server.py
```diff
@@ -515,12 +515,40 @@ def post(self):
...
+            return "Asset switched"
+        if command == "previous":
+            sh.sudo('systemctl', 'kill', '--signal=SIGURG', '--kill-who=main', 'screenly-viewer.service')
```
vpetersson: same.

vpetersson:
Let's restructure this a bit:

 * Let's change SIGUSR2 (reload) to SIGHUP instead, which is more industry standards
 * Let's use SIGUSR2 for going backward.

server.py
vpetersson: Also, let's use avoid using the `sh` module for system calls and instead use the [OS library] (https://docs.python.org/3/library/os.html)

antonmolodykh: hmm. it seems that for systemctl calls we still need sudo

vpetersson: Why do we need to use `systemctl`?  It's just a regular kill signal. For instance, you could do `pkill -SIGUSR1 -f viewer.py`, but ideally you'd do it in [native python](https://stackoverflow.com/questions/2940858/kill-process-by-name).

PR #643, opened 2017-10-05
Fix upgrade

antonmolodykh:


vpetersson:
Looks good, but can you just add a comment block describing why we are doing what we are doing in the steps below the remove/installation?

ansible/roles/system/tasks/main.yml
```diff
@@ -64,6 +68,48 @@
     replace: \1 net.ifnames=0
   when: not is_berryboot
 
+# Sometimes lacks files from some packages. they are required to install pip dependencies.
```
vpetersson: Capitalization.

Issue #609, opened 2017-08-04
Inherit network configuration from NOOBS


vpetersson:
It appears as there is a mechanism in NOOBS for configuring the network. Currently we do not take that into account, but we should look into how this is done and how we can tap into it.

rdcole:
When Running WiFi, If I boot NOOBS and select screenly-ose install it installs fine then starts with no problem.  The Pi will do a Warm boot and the Network settings remain and screenly runs as expected. 
However after a Power Cycle Cold Boot the Network does not Restart so screenly reports it cant resolve the management IP address and screenly has no access to content.

tomscragg:
Hi Viktor,

I installed today, Raspbian, Windows IoT and Screenly today and initially booted into Win IoT. having messed with that for a while first, I unplugged and restarted the system into Screenly. this did initially connect to the network and I got an IP address. upon restart I had the same issue as before. I thought it worth a mention as it maybe to do with the first time the software starts?

vpetersson:
Interesting. Yes, it must be something NOOBS does differently between different boot types. I've added this to the upcoming sprint, so we will try to reproduce and solve this.

tomscragg:
Hi Viktor,

any updates on this? I'm struggling to install the product another way and would be great to get this working :-)

vpetersson:
Hi @tomscragg - yes @antonmolodykh is working on this. I think he made some progress today.

vpetersson:
Just to summarize. Instead of inheriting the network configuration, we have disabled the Screenly Network Manager if NOOBS is detected to avoid conflicts.

PR #647, opened 2017-10-10
Inherit network config from NOOBS

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/609


ansible/roles/network/files/screenly_net_mgr.py
```diff
@@ -163,6 +164,11 @@ def get_active_iface(config, prefix):
 
 
 def main():
+    with open(BOOT_CONFIG_PATH, 'r') as f:
+        boot_config = f.read()
+    if "NOOBS" in boot_config:
+        return
```
vpetersson: Let's add some logging here as well just for simplicity.

PR #652, opened 2017-10-10
Add reboot after installation

antonmolodykh:
Refs #624


vpetersson:
What happens if the installation fails? Will it still reboot?

antonmolodykh:
hmm. I'm not sure, it seems that yeah

vpetersson:
Ok, we need to fix that then. There would be no way for the user to know if the installation failed or not. Perhaps a simple `read` would do with a 'You need to reboot the system for the installation to complete. Would you like to reboot now?'

vpetersson:
LGTM

PR #653, opened 2017-10-13
Fix splash screen

antonmolodykh:


ansible/roles/splashscreen/files/screenly.plymouth
```diff
@@ -0,0 +1,8 @@
+[Plymouth Theme]
+Name=Screenly
+Description=Screenly Splash
```
vpetersson: Screenly Splash Screen

vpetersson:
@antonmolodykh We need to have the 'uninstall' part for the older versions such that they don't end up with both splashscreen modes.

antonmolodykh:
@vpetersson added

ansible/roles/splashscreen/tasks/main.yml
```diff
@@ -1,22 +1,60 @@
...
 
 - name: Copies in rc script
   copy:
     src: asplashscreen
     dest: /etc/init.d/asplashscreen
     mode: 0755
+  when: ansible_distribution_major_version|int <= 7
 
 - name: Enables asplashscreen
   command: update-rc.d asplashscreen defaults
```
vpetersson: Don't you need an inverse of this too?

antonmolodykh: @vpetersson yes, you're right. added.

vpetersson:
LGTM.

PR #659, opened 2017-10-18
Removed unnecessary slashes

antonmolodykh:


vpetersson:
ping @daniep01

daniep01:
Nice, I was about to investigate why the default player name is / on a new install. This is probably why. Is this a regression due to flask migration?

vpetersson:
Could be. Anyways, we're done with the migration now so we should be fine.

PR #668, opened 2017-10-30
Force periodic reload of repeated web assets

cworobetz:
This forces Screenly to check whether the file behind the asset has changed periodically. If so, it will update the content accordingly.

My use case: I host a web server linked to a file share. Assets are set up in Screenly as http://x.x.x.x/1.jpg, http://x.x.x.x/2.jpg, etc., pointed towards the web server's IP.
When users want to update or add an image to the rotation for Screenly, they can copy in a new photo to the file share, rename it to (e.g.) 3.jpg, and it will be included shortly afterwards. 
Without this change, when photos were changed in the file share, the asset wasn't updated and the rotation kept playing the same photos.

Signed-off-by: Cooper Worobetz <cooper@worobetz.ca>

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -307,9 +307,7 @@ def asset_loop(scheduler):
         if 'image' in mime:
             view_image(uri)
         elif 'web' in mime:
-            # FIXME If we want to force periodic reloads of repeated web assets, force=True could be used here.
-            # See e38e6fef3a70906e7f8739294ffd523af6ce66be.
-            browser_url(uri)
+            browser_url(uri, force=True)
         elif 'video' or 'streaming' in mime:
             view_video(uri, asset['duration'])
         else:


boschie:
What happens when a picture does not exist, aka it is empty?

cworobetz:
The picture is left out of the rotation. For example, if there were pictures 1.jpg, 2.jpg, and 3.jpg in a rotation, and you removed 2.jpg, it would rotate between 1.jpg and 3.jpg without any gaps.

vpetersson:
@cworobetz Thanks!

We would need to QA this one throughly. I recall that there was a good reason why we disabled this functionality in the prior commit.

Also, I think we need to make it a setting as it isn't desirable in most cases IMHO. Normally caching should be determined by the browser. If I'm not mistaken, this should *only* be relevant if you only have *one* asset in the playlist. If you have two assets, the caching in the browser should cause it to reload properly.

cworobetz:
@vpetersson Thanks for your feedback. Is there anything specific you'd be looking for in QA? I'll go through 3, 4, and 5 in the [QA checklist](https://www.forgett.com/checklist/1789089623) today.

As far as only being relevant if you have one asset in the playlist, I don't think this is the case. Prior to making this change in the code, if I updated a picture on the file server's end, it would continue displaying the old picture asset until I restarted the pi. I was testing with 3 rotating assets on Screenly.

vpetersson:
> if I updated a picture on the file server's end in my example, it would continue displaying the old picture asset until I restarted the pi. 

Thinking about this a bit more, this could actually be by design in the HTTP protocol.

For instance, here's an image:

```
$ curl -I https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_116x41dp.png                                                                     
[...]
date: Tue, 31 Oct 2017 16:42:30 GMT
expires: Tue, 31 Oct 2017 16:42:30 GMT
cache-control: private, max-age=31536000
last-modified: Thu, 08 Dec 2016 01:00:57 GMT
[...]
```

In this case, the browser is told by the web server that it's OK to cache the image up until `Tue, 31 Oct 2017 16:42:30 GMT` and that it was last modified on `Thu, 08 Dec 2016 01:00:57 GMT`.

When the browser loads it again, unless the last-modified flag is updated, it will keep the cache. 

It would be interesting what those variables would be in your setup. Perhaps either the web server isn't behaving the way we it should or Uzbl (the browser) isn't behaving properly. 

Could you run `curl -I http://[...]` against the asset, then replace it, and run it again and post the result?

cworobetz:
Before changing the photo:
```
curl -I 10.x.x.x/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 17:40:08 GMT
Content-Type: image/jpeg
Content-Length: 136091
Last-Modified: Tue, 29 Sep 2015 08:52:28 GMT
Connection: keep-alive
ETag: "560a514c-2139b"
Accept-Ranges: bytes
```
After changing the photo:
```
curl -I 10.x.x.x/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 17:41:32 GMT
Content-Type: image/jpeg
Content-Length: 3384200
Last-Modified: Wed, 06 Jan 2016 09:24:13 GMT
Connection: keep-alive
ETag: "568cdd3d-33a388"
Accept-Ranges: bytes
```

vpetersson:
Interesting. I guess it doesn't have an expiration header by default.

Just out of curiosity, if you do something like this:

```
location ~* \.(?:ico|css|js|gif|jpe?g|png)$ {
    expires 30m;
    [...]
}
```

Would that cause the player to refresh after 30 minutes?

cworobetz:
I altered my /etc/nginx/sites-enabled/x file to be like so:
```
server {
  server_name 10.x.x.x;
  root /var/www/x;
  location / {
    index index.html;
  }
  location ~* \.(?:ico|css|js|gif|jpe?g|png)$ {
    expires 30m;
  }
}
```
Restarted nginx, then on the Screenly pi re-ran the curl:
```
curl -I http://10.x.x.x/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 18:07:28 GMT
Content-Type: image/jpeg
Content-Length: 136091
Last-Modified: Tue, 29 Sep 2015 08:52:28 GMT
Connection: keep-alive
ETag: "560a514c-2139b"
Expires: Tue, 31 Oct 2017 18:37:28 GMT
Cache-Control: max-age=1800
Accept-Ranges: bytes
```
I removed my commit change from viewer.py, returning it to:
```browser_url(uri)```
and restarted the pi. After connecting again, re-ran curl:
```
curl -I http://10.x.x.x/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 18:16:04 GMT
Content-Type: image/jpeg
Content-Length: 136091
Last-Modified: Tue, 29 Sep 2015 08:52:28 GMT
Connection: keep-alive
ETag: "560a514c-2139b"
Expires: Tue, 31 Oct 2017 18:46:04 GMT
Cache-Control: max-age=1800
Accept-Ranges: bytes
```
And then replaced the image. Not 5 minutes later, the new image was showing up in the rotation. This had me thinking that either including that nginx change caused Screenly / Ubzl to start refreshing the images more frequently, or I made a mistake in thinking that the images weren't updating before. I reverted the nginx conf to the original, restarted nginx and the Pi, then ran curl:
```
curl -I http://10.x.x.x/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 18:34:34 GMT
Content-Type: image/jpeg
Content-Length: 3384200
Last-Modified: Wed, 06 Jan 2016 09:24:13 GMT
Connection: keep-alive
ETag: "568cdd3d-33a388"
Accept-Ranges: bytes
```
Then I swapped the photo:
```
curl -I http://10.16.0.168/mnt/2.jpg
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Tue, 31 Oct 2017 18:42:19 GMT
Content-Type: image/jpeg
Content-Length: 136091
Last-Modified: Tue, 29 Sep 2015 08:52:28 GMT
Connection: keep-alive
ETag: "560a514c-2139b"
Accept-Ranges: bytes
```
And a few minutes later the new photo was showing up. I'm pretty sure that I made a mistake thinking that that my commit fixed the issue. Sorry for wasting your time, I suppose I'll be withdrawing my pull request for now.

vpetersson:
No worries, @cworobetz. Glad you got it working.

Issue #657, opened 2017-10-16
Add support for Resin.io


vpetersson:
There has been attempts to [port](https://github.com/resin-io-playground/resin-screenly-ose) Screenly OSE to Resin.io. 

Some time ago @imrehg started working on a port, but it was never completed. Let's work on this jointly to finalize this.

In short, this is what we need to do:

 * Create a new Dockerfile that with the appropriate FROM value (ARM instead of x86)
 * Add in required components (omxplayer, uzbl etc)
 * Figure out how to fire off X11 and the backend components simultaneously and port the X11 tweaks we use

We can look at [resin-electronjs](https://github.com/resin-io/resin-electronjs) for some inspiration.

PR #670, opened 2017-11-06
Add support for Resin.io

antonmolodykh:
Refs https://github.com/Screenly/screenly-ose/issues/657

### DIFF
--- a/bin/start_resin.sh
+++ b/bin/start_resin.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+systemctl start X.service
+systemctl start matchbox.service
+systemctl start screenly-viewer.service
+systemctl start screenly-web.service
+systemctl start screenly-websocket_server_layer.service
+
+mkdir /data/screenly /data/.screenly /data/screenly_assets /data/.config /data/.config/uzbl
+cp -n ansible/roles/screenly/files/screenly.conf /data/.screenly/screenly.conf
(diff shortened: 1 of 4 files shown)


bin/start_resin.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+systemctl start X.service
+systemctl start matchbox.service
+systemctl start screenly-viewer.service
+systemctl start screenly-web.service
+systemctl start screenly-websocket_server_layer.service
+
+mkdir /data/screenly /data/.screenly /data/screenly_assets /data/.config /data/.config/uzbl
+cp -n ansible/roles/screenly/files/screenly.conf /data/.screenly/screenly.conf
```
vpetersson: We can't do this. This would overwrite the config on each start. Must be wrapped around some conditional logic (if file not exist..)

bin/start_resin.sh
```diff
@@ -0,0 +1,19 @@
...
+cp -n ansible/roles/screenly/files/screenly.db /data/.screenly/screenly.db
```
vpetersson: Same here.

bin/start_resin.sh
vpetersson: Ah, ok. Sorry, overlooked that. I usually do an if condition for this.

vpetersson:
Good job on this, @antonmolodykh.

Issue #600, opened 2017-07-26
Video duration changing


antonmolodykh:
There should be able to change the duration of the video only for rtsp/rtmp video.
In addition, the `get_video_duration` function is very heavy and we don't want to call it every time

PR #671, opened 2017-11-07
Video duration changing

antonmolodykh:
Now for the YouTube video, we get the information about the duration through youtube-dl.
For all videos, we take the duration only once (first) and remain unchanged

Refs #600

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -108,7 +108,7 @@ def get_video_duration(file):
     time = None
     try:
         if arch in ('armv6l', 'armv7l'):
-            run_omxplayer = omxplayer(file, info=True, _err_to_out=True)
+            run_omxplayer = omxplayer(file, info=True, _err_to_out=True, _ok_code=[0, 1])
             for line in run_omxplayer.split('\n'):
                 if 'Duration' in line:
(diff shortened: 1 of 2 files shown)


lib/utils.py
```diff
@@ -108,7 +108,7 @@ def get_video_duration(file):
     time = None
     try:
         if arch in ('armv6l', 'armv7l'):
-            run_omxplayer = omxplayer(file, info=True, _err_to_out=True)
+            run_omxplayer = omxplayer(file, info=True, _err_to_out=True, _ok_code=[0, 1])
```
vpetersson: Why do we accept error code 1 here/

antonmolodykh: if briefly, all the time the `get_video_duration` function did not work correctly on Raspberry Pi. in omxplayer there is an error, because of which information about the video is output in strerr (this was taken into account), but in addition omxplayer returns code 1 (this was not taken into account). We did not get an error from `get_video_duration`, because almost the entire function is wrapped in try-except (which is needed there, but it has other target). as a result, the video always received a duration of "N / A".
now the duration is taken correctly.

vpetersson:
LGTM.

Issue #651, opened 2017-10-10
"Modernize" /api/v1/assets API endpoint


vpetersson:
Currently the API endpoint `/api/v1/assets` uses `Content-Type: application/x-www-form-urlencoded`. This isn't a very good design and makes for a poor API. Let's introduce a new API call (`/api/v1.1/assets`) that uses proper JSON and update the UI to use this while we're at it.

PR #673, opened 2017-11-08
Modernize assets endpoint

antonmolodykh:
Refs #651

### DIFF
--- a/server.py
+++ b/server.py
@@ -158,7 +158,13 @@ def prepare_asset(request):
     req = Request(request.environ)
     data = None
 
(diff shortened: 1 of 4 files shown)


vpetersson:
@antonmolodykh We still need to maintain backward compatibility. That's why the ticket mentioned the new version for the API. We need to have both API points working for some time before we deprecate the old one.

antonmolodykh:
@vpetersson 
oh, my bad. moment

vpetersson:
LGTM.

Issue #674, opened 2017-11-12
Fix final Resin issues


vpetersson:
I was playing with the Resin addition today and found a few issues

* Update documentation for how to use remote URL. There also appears to be an issue since we use port 8080
* It appears as the [tests now fail](https://travis-ci.org/Screenly/screenly-ose/builds/301023439?utm_source=email&utm_medium=notification)
* There is no mention of `DISABLE_UPDATE_CHECK` in the documentation

imrehg:
The remote URL currently can be used:

* Screenly server running on port 8080, and then use "http://<uuid>.resindevice.io:8080", manually setting HTTP and the port
* or internally redirecting from port 80->8080 in the resin application, and manually set HTTP ("http://<uuid>.resindevice.io").

For the default Public URL to work without manual setting, would need to enable to use wss:// for websockets (instead of unencrypted ws://) and adding the internal redirect.

vpetersson:
@imrehg - Is it  possible to do port mapping in Docker? We're running on 8080 because we drop permission to a runtime user rather than running as root.

imrehg:
@vpetersson Yes, understood the permission drop. The port mapping was part of the original example I've sent, with iptables setting up the redirection https://github.com/resin-io-playground/resin-screenly-ose/blob/master/startup.sh#L45-L59 This is I guess not the only way, but it was working well enough when I've tested it.

vpetersson:
@imrehg We support HTTP/HTTPS internally (by using Nginx as a reverse proxy). If we enable this on a self-signed cert, would that work?

vpetersson:
@imrehg -  ignore the issue above with Nginx. I take it that your proxy reverse proxy from 443 to 80 on the device only. As such, there is no need to enable SSL on the device.

@antonmolodykh - in regards to @imrehg comment above about `ws://` vs `wss://`, we  need to fix this as it affects `enable_ssl.sh` as well. How about something like this:

 * When the user runs `enable_ssl.sh`, we add a new key to `screenly.conf`:
```
admin_ssl = True
```

 * If this key is present, we use `wss://` instead of `ws://` for the websocket
 * We can then piggyback on this same logic for Resin by simply adding this key

vpetersson:
Also, as discussed with @antonmolodykh - we need a new Nginx config that only reverse proxies 80 -> 8080 on the device.

vpetersson:
[...]

PR #678, opened 2017-11-14
Fix Resin issues

antonmolodykh:
Refs #674

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,5 +1,5 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-python
-MAINTAINER Viktor Petersson <vpetersson@screenly.io>
+FROM resin/%%RESIN_MACHINE_NAME%%-debian
+MAINTAINER Anton Molodykh <amolodykh@screenly.io>
 
 RUN apt-get update && \
     apt-get -y install \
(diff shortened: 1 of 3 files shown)


Dockerfile.template
```diff
@@ -1,4 +1,4 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-python
+FROM resin/%%RESIN_MACHINE_NAME%%-debian
 MAINTAINER Viktor Petersson <vpetersson@screenly.io>
```
vpetersson: @antonmolodykh this should be you :)

antonmolodykh: ok :)

Dockerfile.template
```diff
@@ -21,38 +22,36 @@ RUN apt-get update && \
         x11-xserver-utils \
         xserver-xorg && \
     apt-get clean
+RUN apt-get -y remove stunnel4
```
vpetersson: Why do we have this? If stunnel is installed already? Feels a bit redundant.

antonmolodykh: it was in an ansible script, and I thought it was some kind of precaution

vpetersson: Ah, yeah i think that was legacy. You should be able to remove that:

```
$ grep 'stunnel' ./* -R
./ansible/roles/ssl/tasks/main.yml:- name: Removes stunnel
./ansible/roles/ssl/tasks/main.yml:    name: stunnel4
```

antonmolodykh:
Includes contributions from @imrehg

PR #680, opened 2017-11-17
Fix final Resin issues

antonmolodykh:
solved the problem with wifi
added output of systemd logs
solved the problem related ws connection

Refs #674


bin/start_resin.sh
```diff
@@ -7,6 +7,8 @@ mkdir -p \
     /data/screenly \
     /data/screenly_assets
 
+touch /data/.resin
```
vpetersson: This is probably redundant now that you pass on the Resin UUID

vpetersson:
LGTM

Issue #638, opened 2017-09-26
Create front-end buttons for next/previous asset


vpetersson:
Now that we have the API calls for back/forward, let's add some UI elements to allow these controls too.

PR #684, opened 2017-11-20
Next previous buttons

antonmolodykh:
Added front-end buttons for next-prev. 
In addition, before https://github.com/Screenly/screenly-ose/commit/0f3df0d80d727c4bc79e54a182d1dd6d57c31057, we often received an incorrect `video` duration ('N/A'). after this, when trying to edit such `videos`, we are getting an error. now it is fixed.

Refs #638


vpetersson:
LGTM

Issue #689, opened 2017-11-26
Switch default to use nginx reverse proxy


vpetersson:
With the Raspberry Pis getting more and more powerful, it makes sense to enable the Nginx reverse proxy to be the default. This will make it more natural to navigate to the device. However, we should still make it possible to change the listening port to not use Nginx.

PR #695, opened 2017-11-29
Make nginx default

antonmolodykh:
Refs #689


vpetersson:
@antonmolodykh Just glanced over this, but it looks like this adds SSL support, no? I'm thinking we should just have HTTP and then if you run the HTTPS enabler, we enable HTTPS.

antonmolodykh:
@vpetersson ah, apparently I misunderstood the ticket

PR #696, opened 2017-11-29
api add get info

heberalmeida:
included in api device information, what is currently being loaded and size of space, to be used in external applications.

### DIFF
--- a/server.py
+++ b/server.py
@@ -654,10 +654,6 @@ class Info(Resource):
         slash = statvfs("/")
         free_space = size(slash.f_bavail * slash.f_frsize)
 
-        # Get uptime
-        uptime_in_seconds = diagnostics.get_uptime()
-        system_uptime = timedelta(seconds=uptime_in_seconds)
-
         return {
             'viewlog': viewlog,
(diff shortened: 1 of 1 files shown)


vpetersson:
Thanks, @heberalmeida. Glanced over it and it seems sensible.

@antonmolodykh can you audit this PR?

Thanks.

antonmolodykh:
@heberalmeida hmm. as far as I can see, you declare `system_uptime` and don't use it in any way.
Also, does it make sense to print systemctl logs here, if this information is on the "system info" tab?

heberalmeida:
I did this because I needed this information because I use the api to register in multiple raspberry and this did not have in the api, in some moments I needed to see what was happening at the moment, and the amount of space available, so it includes in the api

antonmolodykh:
@heberalmeida I tried to run it. When I send a request to the endpoint I get a server error

heberalmeida:
@antonmolodykh I will correct

antonmolodykh:
LGTM.
Thanks, @heberalmeida

PR #698, opened 2017-12-01
Make nginx default

antonmolodykh:
Refs #689

### DIFF
--- a/ansible/roles/nginx/tasks/main.yml
+++ b/ansible/roles/nginx/tasks/main.yml
@@ -26,11 +26,11 @@
     group: root
     force: no
 
-- name: Modifies Screenly Server to only listen on localhost
+- name: Modifies screenly-web service to only listen on localhost
   lineinfile:
-    line: ';listen = 0.0.0.0:8080'
-    regexp: '^.*listen.*'
-    dest: /home/pi/.screenly/screenly.conf
+    regexp: '^.*LISTEN.*'
+    state: absent
+    dest: /etc/systemd/system/screenly-web.service
   when: no_ssl
 
 # Disable Nginx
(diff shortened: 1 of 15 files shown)


ansible/roles/enable-nginx/files/nginx.conf
```diff
@@ -0,0 +1,17 @@
...
+  server_tokens off;
+  listen 80 default_server;
+
+  location / {
+    client_max_body_size 4G;
+    proxy_pass http://ose;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header Host $http_host;
+    proxy_set_header X-Forwarded-Protocol https;
```
vpetersson: This is wrong.

ansible/site.yml
```diff
@@ -30,4 +30,6 @@
     - rpi-update
     - screenly
     - splashscreen
+    - enable-nginx
```
vpetersson: Feels a bit award to have both. Let's merge them into one role called `nginx` and mark it up properly with comments.

vpetersson:
How about Swagger? I didn't see any changes related to it, which makes me think it will not work.

antonmolodykh:
You're right, I did not take one case into account. Fixed.

I don't see a good way to remove the `listen` variable from the configurations. at least for the reason that from there we get the ip on which we start the server.

ansible/roles/nginx/tasks/main.yml
```diff
@@ -0,0 +1,81 @@
...
+    - disable-nginx
+
+- name: Remove nginx config
+  file:
+    path:  /etc/nginx/sites-enabled/screenly.conf
+    state: absent
+  tags:
+    - disable-nginx
+
+- name: Remove self-signed certificates
```
vpetersson: Let's leave these actually. No point in regenerating them if this is recreated later.

vpetersson:
LGTM.

PR #699, opened 2017-12-03
Add support for usb0 device

crdotson:
In USB gadget mode, the network interface shows up as usb0.  You might want to use this interface for a Pi Zero running signage near another machine that can provide network connectivity via USB.


vpetersson:
@crdotson Thanks! looks like a good improvement, but could you fix the test? There's a lint error.

crdotson:
Whoops, I didn't notice that!  Fixed, thanks.

vpetersson:
LGTM. Merging. Thanks.

Issue #700, opened 2017-12-03
Issue with API calls on master


vpetersson:
Not exactly sure when this broke, but the API calls for navigating are broken both on the production and master branch. 

I'm getting this in the logs which is weird since it worked earlier:

```
03.12.17 11:31:15 (+0000) Dec 03 11:31:15 a9c5713 python[698]: File "/home/pi/screenly/viewer.py", line 59, in sigusr2
03.12.17 11:31:15 (+0000) Dec 03 11:31:15 a9c5713 python[698]: scheduler.reverse = True
03.12.17 11:31:15 (+0000) Dec 03 11:31:15 a9c5713 python[698]: AttributeError: 'NoneType' object has no attribute 'reverse'
03.12.17 11:31:16 (+0000) Dec 03 11:31:16 a9c5713 systemd[1]: screenly-viewer.service: main process exited, code=exited, status=1/FAILURE
```

PR #702, opened 2017-12-07
Fix viewer for Resin

antonmolodykh:
Apparently, this arose if we pressed back before the `scheduler` was created. for example, while the `splash_page` on the screen. in addition there was a problem if we click on any navigation buttons quickly several times. it was only on Resin

Refs #700


vpetersson:
LGTM

Issue #688, opened 2017-11-25
Add auto-resize of disk on first boot


vpetersson:
There is a functionality in Raspbian to resize the disk on first boot. We should try to invoke this during the build process.

Let's read the code startup code for Raspbian and see what is triggering this behavior.

vpetersson:
I can confirm that this does not work. You can reproduce it with the latest disk image.

antonmolodykh:
@vpetersson in the new release there is no `init=/usr/lib/raspi-config/init_resize.sh` in the `/boot/cmdline.txt` file. I think that the cause of the problem in this.

vpetersson:
There is. I'm 99.9% confident as I added it before. It will however disappear after first boot. If you mount the disk image, you should be able to see it there.

antonmolodykh:
I just downloaded a new image and checked the file. I did NOT run the device, I checked it using computer.

vpetersson:
Strange. Ok, I guess we might have to do a hotfix then.

vpetersson:
I stand corrected. You're right. I screwed up. WIll do a new image.

PR #710, opened 2017-12-13
Add auto-resize of disk on first boot

antonmolodykh:
Refs #688


vpetersson:
Why do we need to fetch an external file for this? Since it's already built into Raspbian, we should simply use the same logic (i.e. trigger it again).

antonmolodykh:
@vpetersson this is the same file that is used in Raspbian. The matter is that it works exactly once: after that the file and the things connected with it are deleted.

Issue #691, opened 2017-11-27
Remove Screenly Network Manager


vpetersson:
It seems like there has been a fair amount of issues with Screenly Network Manager on Screenly OSE. Let's disable this in favor of #686.

vpetersson:
@antonmolodykh Can you provide instructions for how to disable Screenly Network Manager (the `systemctl` commands) for future references?

antonmolodykh:
To just turn off the service, just enter `systemctl disable screenly-net-manager.service` command

antonmolodykh:
I think that we can not completely abandon this, because as @majorz [said](https://github.com/Screenly/screenly-ose/issues/686#issuecomment-346333810), resin-wifi-connect is now being run on Stretch. so we need to leave support for Jessie.

vpetersson:
@antonmolodykh - I think that's fine. We will just revert to the default method in Raspbian for earlier releases.

vpetersson:
Solved in #715.

Issue #686, opened 2017-11-21
Add resin-wifi-connect


mckaycr:
Wondering if it might add any value to add the resin-wifi-connect?

vpetersson:
Interesting idea. I don't think it would be compatible with regular Raspbian, so we'd have to make it unique to the resin.io Dockerfile. Perhaps @imrehg can chime in.

mckaycr:
Their documentation says it has also been successfully tested with the on-board WiFi on a Raspberry Pi 3, not sure if that helps.

vpetersson:
Yeah the problem isn't that. The problem is to integrate it into the stack.

mckaycr:
Makes sense.  Its not a must have, its a nice to have, so if it doesn't make it in it's no big deal.

vpetersson:
Happy to accept a PR if you want to hack on it @mckaycr.

imrehg:
@vpetersson it should not require a resin base image to work as it stands now, so should be possible to integrate. One thing, though, not sure if the access point interface's port is configurable or not, since here both port 80 and 8080 are in use. @majorz might now a bit more.

majorz:
The port is not configurable yet, but I thihnk we can add such option.

Latest version of WiFi Connect runs on Raspbian Stretch, but requires switching the network manager from dhcpcd to NetworkManager, which maybe will not suit your use case. We have installation script that does that: https://github.com/resin-io/resin-wifi-connect/blob/master/scripts/raspbian-install.sh
Documentation about this will land soon together with other improvements.

vpetersson:
Thanks @imrehg / @majorz. Good pointers.

> [...]  but requires switching the network manager from dhcpcd to NetworkManager, which maybe will not suit your use case.

I don't see that as a problem. Should work fine.

> Documentation about this will land soon together with other improvements.

Excellent. Basically what we need is to figure out the logic. Do you guys have any thoughts on this? Basically I think it would make sense to only display/activate this hotspot when there is no connection/DHCP found through Ethernet.

majorz:
Hey, sorry for the delayed response. I missed somehow the notification earlier this week and found it now by checking my backlogs :/

I will be working on customizing the port number today and which will be included in the version we are about to release next week.

Yes, it makes sense to run WiFi Connect only when similar condition exists. I will think a bit more about that and follow up soon!

vpetersson:
[...]

PR #715, opened 2017-12-23
Add resin-wifi-connect

antonmolodykh:
Now `resin-wifi-connect` is turned on only at the first boot and if there is no connection.
also it works in NOOBS

Refs #686 
Refs #691

### DIFF
--- a/ansible/roles/network/tasks/main.yml
+++ b/ansible/roles/network/tasks/main.yml
@@ -1,100 +1,71 @@
-- name: Copies network manger and watchdog
-  copy:
-    owner: root
-    group: root
-    mode: 0755
-    src: "{{ item }}"
-    dest: "/usr/sbin/{{ item }}"
+- name: Check if screenly-network-manager files exist
+  stat:
+    path: /usr/sbin/screenly_net_mgr.py
+  register: screenly_network_manager
+
+- set_fact: screenly_network_manager_exist="{{screenly_network_manager.stat.exists}}"
+
+- name: Disable network manager
+  command: systemctl disable screenly-net-manager.service
+  when: screenly_network_manager_exist
+
+- name: Disable network watchdog
+  command: systemctl disable screenly-net-watchdog.timer
+  when: screenly_network_manager_exist
+
+- name: Remove network manger and watchdog
+  file:
+    state: absent
+    path: "/usr/sbin/{{ item }}"
   with_items:
     - screenly_net_mgr.py
     - screenly_net_watchdog.py
-  when: ansible_distribution_major_version|int != 9
 
-- name: Copies network manager unit file
-  copy:
-    owner: root
-    group: root
-    src: screenly-net-manager.service
-    dest: /etc/systemd/system/screenly-net-manager.service
-  when: ansible_distribution_major_version|int != 9
-
-- name: Copies network watchdog service file
-  copy:
-    owner: root
-    group: root
-    src: screenly-net-watchdog.service
-    dest: /etc/systemd/system/screenly-net-watchdog.service
-  when: ansible_distribution_major_version|int != 9
-
-- name: Copies network watchdog timer file
-  copy:
-    owner: root
-    group: root
-    src: screenly-net-watchdog.timer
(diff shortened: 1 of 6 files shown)


ansible/roles/network/tasks/main.yml
```diff
@@ -8,27 +8,31 @@
   with_items:
     - screenly_net_mgr.py
     - screenly_net_watchdog.py
+  when: ansible_distribution_major_version|int != 9
```
vpetersson: `!= 9` should be `< 9`

ansible/roles/network/tasks/main.yml
```diff
@@ -39,16 +43,86 @@
...
+
+
+# Use resin-wifi-connect if Stretch
+
+- name: Install NetworkManager
+  apt:
+    name: network-manager
+    state: present
+  when: ansible_distribution_major_version|int == 9
```
vpetersson: `== 9` should be `>= 9`

ansible/roles/network/tasks/main.yml
```diff
@@ -39,16 +43,86 @@
...
+  command: systemctl disable dhcpcd
+  when: ansible_distribution_major_version|int == 9
+
+- name: Activate NetworkManager
+  command: systemctl enable NetworkManager
+  when: ansible_distribution_major_version|int == 9
+
+- name: Download resin-wifi-connect release
+  get_url:
+    url: https://github.com/resin-io/resin-wifi-connect/releases/download/v3.0.4/wifi-connect-v3.0.4-linux-rpi.tar.gz
```
vpetersson: Let's make `3.0.4` a variable in Ansible such that we can easily control the version.

ansible/roles/network/tasks/main.yml
```diff
@@ -39,16 +43,86 @@
...
+    dest: /home/pi/resin-wifi-connect.tar.gz
+  when: ansible_distribution_major_version|int == 9
```
vpetersson: You should add a secondary when condition here to prevent the installer from running every time. Perhaps you could touch the file `/usr/local/share/wifi-connect/ui/.$VERSION`

ansible/roles/network/tasks/main.yml
```diff
@@ -39,16 +43,86 @@
...
+
+- name: Activate NetworkManager
+  command: systemctl enable NetworkManager
+  when: ansible_distribution_major_version|int == 9
+
+- name: Download resin-wifi-connect release
+  get_url:
+    url: https://github.com/resin-io/resin-wifi-connect/releases/download/v3.0.4/wifi-connect-v3.0.4-linux-rpi.tar.gz
+    dest: /home/pi/resin-wifi-connect.tar.gz
```
vpetersson: It's better to use `/tmp` to avoid writes to the SD card.

antonmolodykh: I agree and I did so initially, but it caused some problems. I mentioned this in one of the reports:
>  if I unpack `resin-wifi-connect` into `/tmp` folder, then the installation of apt packages does not work. I could not find an explanation for this, but now I'm just unpacking it in another directory.

ansible/roles/network/tasks/main.yml
antonmolodykh: I alsk throught about it. but in the [script](https://github.com/resin-io/resin-wifi-connect/blob/master/scripts/raspbian-install.sh) it is strictly set to version 9 (although I suppose they could do `<=`). I  think that maybe it has reasons

vpetersson: Well, basically we want to leave networking as is (i.e. remove our own network manager) for versions older than Stretch (9) and for Stretch and later, use WiFi Network Manager.

ansible/roles/network/tasks/main.yml
vpetersson: Interesting. I'm not sure how big this package is. Since `/tmp` is tmpfs, perhaps it exceeds the amount of available RAM. In any case, ignore this comment then.

vpetersson:
LGTM. Good job.

PR #716, opened 2017-12-30
Fix netowrk tasks

antonmolodykh:
here are some fixes, but I have not completed all the checks yet


ansible/site.yml
```diff
@@ -9,6 +9,11 @@
         name: nginx
         state: restarted
 
+    - name: install network-manager
```
vpetersson: Shouldn't this be conditional to `ansible_distribution_major_version|int >= 9`?

antonmolodykh: it's just a handler. it is called from the `network`. in the task that calls it there is a condition

vpetersson: Not sure that I follow. This installs the package 'network-manager' from `apt`. That is a fairly big change.

antonmolodykh: The matter is that this should be done when installing the `resin-wifi-connect`, but if we are connected via wifi, then when the NetworkManager is installed the network will be reconfigured and this will cause ansible fall. so I want to make sure that this will be called the last one

PR #718, opened 2018-01-02
Refactoring of check_update()

vpetersson:
This diff refactors the `check_update()` function in `viewer.py`. In short, it:

 * Changes the update mechanism to pull the statistics directly from Github.
 * Sends a anonymized ping to Mixpanel to better understand the number of active devices (an opt-out is available).


viewer.py
```diff
@@ -307,30 +313,49 @@ def check_update():
...
     logging.debug('Last update: %s' % str(last_update))
 
     git_branch = sh.git('rev-parse', '--abbrev-ref', 'HEAD').strip()
+    git_hash = sh.git('rev-parse', '--short', 'HEAD').strip()
+
     if last_update is None or last_update < (datetime.now() - timedelta(days=1)):
 
-        if not url_fails('http://stats.screenlyapp.com'):
-            latest_sha = req_get('http://stats.screenlyapp.com/latest/{}'.format(git_branch))
+        if not settings['analytics_opt_out'] or not is_ci():
```
antonmolodykh: it seems to me that this should be "and"

vpetersson: No, because we don't want that code to run on CI ever as it would skew the data.

PR #719, opened 2018-01-03
docs: By default, listen on all (more reasonable default for new users)

AdrianOrenstein:
I wasn't able to run the docker image, and found that the docker wasn't listening to connections from my computer, this is a quick hack for it to work.


vpetersson:
Good catch, @AdrianOrenstein.

Issue #728, opened 2018-01-15
`.screenly/wifi_set` is created prematurely 


vpetersson:
Sub-ticket to #725.

It turns out that if you run an upgrade from a build prior to a build that includes WiFi connect, the file `.screenly/wifi_set` is prematurely created. 

For instance, if you run the upgrade with an Ethernet connection, then unplug it during the reboot (as explained in #725), somehow the wifi_set file is created prior to the reboot, which causes the behavior in #727 to occur.

Issue #727, opened 2018-01-15
Screen stuck in 'Unable to resolve local IP address' (breakout)


vpetersson:
Sub-ticket to #725.

In short, if the file `.screenly/wifi_set` exists and there is no Ethernet connection, the screen is stuck on "Unable to resolve local IP address indefinitely.

kams01:
Please refer to my comment in issue: https://github.com/Screenly/screenly-ose/issues/725#issuecomment-480512237

I cannot find wifi_set and I am still stuck on the unable screen. Short of reinstalling this, I am stuck! using NOOBS

Issue #725, opened 2018-01-09
Screen stuck in 'Unable to resolve local IP address'


vpetersson:
It appears as something in the latest update broke the viewer logic. If you boot a device without any network connection, it will get stuck displaying the splash screen.

vpetersson:
I just reproduced this again.

Steps to reproduce (on a Pi3 Model B):

 * Flash out the [latest disk image](https://github.com/Screenly/screenly-ose/releases/download/v0.13/2017-11-25-Screenly_OSE_4GB.zip)
 * Connect the device using Ethernet
 * Optional: Enable SSH
 * Run `./bin/run_upgrade.sh` and install the development release
 * Reboot the device and unplug the Ethernet cable when the four Raspberries appears on the screen.

**Result:** The device just shows "Unable to resolve local IP Address" indefinitely
**Expected result:** The Resin WiFi Connect to appear

vpetersson:
So the root cause for this was that `.screenly/wifi_set` was created *prior* to the reboot. This causes the wifi connect never to be launched. There are subsequently two related bugs:
 
* #727
* #728

vpetersson:
Unable to reproduce this after the merge.

kams01:
I recently starting working on this as a user to setup OSE at our facility. I used NOOBS to install OSE. After you move the Pi to another network, the network selection tool does not pop back up after "Unable to resolve IP"

Since then I have used several things - Updated to latest build <- by the way, how do I know which version of OSE am I running? there is no simple way to know this!

Updated to the latest development build (all using SSH or local)
Brought it back to the original network where it attached and hit the "Rerun Network Detection" and it does not reset anything.
Tried to remove the wifi-set file, but it cannot find it.
So basically had to reinstall OSE at the final location and deploy it there - not very practical if you want to test a few things out earlier.

So am I missing anything here?

PR #729, opened 2018-01-17
Remove wifi_set after installation

antonmolodykh:
Added removal of `wifi_set` after installation

Refs #725 #727 #728

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -82,6 +82,8 @@ sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*' 
 
 cd ~/screenly && git rev-parse HEAD > ~/.screenly/latest_screenly_sha
 
+if test -f ~/.screenly/wifi_set; then  rm ~/.screenly/wifi_set; fi
+
 set +x
 echo "Installation completed."
 


bin/install.sh
```diff
@@ -87,5 +87,8 @@ echo "Installation completed."
 
 read -p "You need to reboot the system for the installation to complete. Would you like to reboot now? (y/N)" -n 1 -r -s REBOOT && echo
 if [ "$REBOOT" == 'y' ]; then
+  if test -f ~/.screenly/wifi_set; then  rm ~/.screenly/wifi_set; fi
```
vpetersson: Why do you wrap this into an if statement if it is repeated regardless of the outcome below?

antonmolodykh: because the next line reboots the system

vpetersson: Sure, but if you move up `if test -f ~/.screenly/wifi_set; then  rm ~/.screenly/wifi_set; fi` to before 88, it will have the same effect but you won't have to repeat it.

antonmolodykh: in fact, there is a slight difference. it may take a couple of minutes before you hit "y" or "N".
but I think you're right. if the viewer does restart, it will happen earler and without segnificant delay. I'll fix it now

lbaumgartner15:
Okay all these refs are super confusing. I just need some clarification.

I get connected after a fresh install and then I reboot after changing the orientation of the screen and then I get the error "unable to resolve local IP address." 

So next time I install what do I need to do. (Already wiped SD card and am currently reloading). I am very new to all of this so please don't just type some code. Please give step by step instructions of what I need to type. Thank you!

Issue #741, opened 2018-02-06
Screenly Viewer crashes after a few hours without internet connection


Andreas879:
First of all, thank you for developing Screenly. It's just great.
I have installed the latest Screenly OSE 20180119 Image. Also in the previous version, after a few hours (approx. 15h) the Screenly Viewer crashes. Screenly runs in a network without internet connection. NTP service is installed and running. As soon as Screenly can access to the internet, the viewer works again. I can reproduce this error by blocking the Internet access. I think it is because of the missing update check

The screen stays black and flashes white after approximately 6 seconds

Hardware: Raspberry Pi 3 Model B with SanDisk 16GB Class 10

syslog

Feb  6 17:12:37 Infoboard systemd[1]: Started Screenly Viewer.
Feb  6 17:12:38 Infoboard systemd[1]: Started Fail2Ban Service.
Feb  6 17:12:38 Infoboard systemd[1]: Reached target Multi-User System.
Feb  6 17:12:38 Infoboard systemd[1]: Reached target Graphical Interface.
Feb  6 17:12:38 Infoboard systemd[1]: Starting Update UTMP about System Runlevel Changes...
Feb  6 17:12:38 Infoboard systemd[1]: Started Update UTMP about System Runlevel Changes.
Feb  6 17:12:38 Infoboard systemd[1]: Startup finished in 1.477s (kernel) + 10.164s (userspace) = 11.642s.
Feb  6 17:12:39 Infoboard python[540]: [2018-02-06 17:12:39 +0000] [540] [INFO] Starting gunicorn 19.4.5
Feb  6 17:12:39 Infoboard python[540]: [2018-02-06 17:12:39 +0000] [540] [INFO] Listening at: http://127.0.0.1:8080 (540)
Feb  6 17:12:39 Infoboard python[540]: [2018-02-06 17:12:39 +0000] [540] [INFO] Using worker: threads
Feb  6 17:12:39 Infoboard python[540]: [2018-02-06 17:12:39 +0000] [622] [INFO] Booting worker with pid: 622
Feb  6 17:12:39 Infoboard python[568]: browser found dead, restarting
Feb  6 17:12:39 Infoboard python[568]: Loading browser...
Feb  6 17:12:39 Infoboard python[568]: Browser loading file:///tmp/screenly_html/black_page.html. Running as PID 626.
Feb  6 17:12:48 Infoboard systemd[1]: Time has been changed
[...]

PR #742, opened 2018-02-07
Fix for check_update

antonmolodykh:
The `viewer` breaks down when trying to check for updates without an internet connection. Here it's prevented.

Refs #741


vpetersson:
LGTM

PR #756, opened 2018-03-02
Make ipv6-only setups work

p-rintz:
Change get_node_ip to also work with ipv6-only setups.

Without this, if a pi does not have an ipv4 connection, get_node_ip will fail.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -79,9 +79,9 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            interface_id = next(iter([gateways()][0]['default']), None)
+            interface_id = max(list(gateways()['default']))
             default_interface = [gateways()][0]['default'][interface_id][1]
-            my_ip = ifaddresses(default_interface)[min(ifaddresses(default_interface))][0]['addr']
+            my_ip = ifaddresses(default_interface)[interface_id][0]['addr']
             return my_ip
         except:
             raise Exception("Unable to resolve local IP address.")
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -9,6 +9,7 @@ import requests
 from datetime import datetime, timedelta
 from distutils.util import strtobool
 from netifaces import ifaddresses
+from netifaces import gateways
 from os import getenv, path, utime
 from platform import machine
 from settings import settings, ZmqPublisher
@@ -78,16 +79,12 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-6nr'), '-e', '^::/0.*UG').split()[-1]
-            my_ip = ifaddresses(default_interface)[10][0]['addr']
+            interface_id = next(iter([gateways()][0]['default']), None)
+            default_interface = [gateways()][0]['default'][interface_id][1]
+            my_ip = ifaddresses(default_interface)[min(ifaddresses(default_interface))][0]['addr']
             return my_ip
         except:
-          try:
-              default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-              my_ip = ifaddresses(default_interface)[2][0]['addr']
-              return my_ip
-          except:
-              raise Exception("Unable to resolve local IP address.")
+            raise Exception("Unable to resolve local IP address.")
 
 
 def get_video_duration(file):
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -78,11 +78,16 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            default_interface = grep(netstat('-6nr'), '-e', '^::/0.*UG').split()[-1]
+            my_ip = ifaddresses(default_interface)[10][0]['addr']
             return my_ip
         except:
-            raise Exception("Unable to resolve local IP address.")
+          try:
+              default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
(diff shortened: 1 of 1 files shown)


vpetersson:
Nice touch with IPv6. However, from what I can tell, this would break the IPv4 support. I'd love to get IPv6 support in, but we'd need to maintain the IPv4 support.

p-rintz:
I dont have much python experience, but as it is in the PR it should jump to the old ipv4 solution if a ipv6 gateway was not found, or not?

vpetersson:
My bad, @p-rintz. I misread the code. @antonmolodykh can you do a check on the code?

antonmolodykh:
Sorry that I answer so late.
@p-rintz I really don't like this try-except-try-except construction. perhaps instead you can use [netifaces](https://pypi.org/project/netifaces/) module?

p-rintz:
@vpetersson Yes, sorry. Took me a while to get back to this.

I have changed the PR as requested.
Its not quite as I would like it to be, as this way an IPv4 address would always be preferred over IPv6 if both are available, but it works at least.

lib/utils.py
```diff
@@ -78,8 +79,9 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            interface_id = next(iter([gateways()][0]['default']), None)
+            default_interface = [gateways()][0]['default'][interface_id][1]
+            my_ip = ifaddresses(default_interface)[min(ifaddresses(default_interface))][0]['addr']
```
antonmolodykh: If i'm not mistaken this is not correct. We cannot just take a min value. Please read paragraph 2 (How do I use it?), pay attention to the use of AF_INET and AF_INET6. If I understand correctly there may be cases when AF_LINK < AF_INET6

lib/utils.py
```diff
@@ -78,8 +79,9 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            interface_id = next(iter([gateways()][0]['default']), None)
+            default_interface = [gateways()][0]['default'][interface_id][1]
```
antonmolodykh: If I understand correctly 82 and 83 lines can be replaces by
```
default_interface = gateways()['default']
```

lib/utils.py
```diff
@@ -9,6 +9,7 @@
 from datetime import datetime, timedelta
 from distutils.util import strtobool
 from netifaces import ifaddresses
+from netifaces import gateways
```
antonmolodykh: `from netifaces import gateways, ifaddresses`

lib/utils.py
p-rintz: AF_LINK is always < AF_INET6, which is why I mentioned that its not my preferred solution as it would mean that IPv4 will be selected over IPv6 when both are available.

Other solutions that I came up with would have required a try/except again, which was not wanted above.

lib/utils.py
p-rintz: This is not the same sadly.

A gateways()['default'] will return multiple interfaces and also gateway ips.

lib/utils.py
antonmolodykh: I'm talking about AF_LINK, not about AF_INET

lib/utils.py
antonmolodykh: oh, my bad, you're right :) I meant something like this:
```
gateways()['default'].items()[0][1][1]
```
I'm not saying that this is the best option, it just seems to me 82 and 83 lines is not the most elegant solution and its really hard to read

lib/utils.py
p-rintz: I misunderstood you then. I tested it on multiple raspberries and a few laptops and have not found AF_LINK to be smaller than AF_INET6, but if this solution is not wanted, I currently dont see another without try/except, other than displaying all ips available to the system, which would require larger changes at different places from a quick look I took.

lib/utils.py
p-rintz: Yes, it would work. But only in python2, in case that would be a problem. An easier to read solution that would also work in python3 would be:
`interface_id = list(gateways()['default'])[0]`

I dont really have a preferred solution here.

p-rintz:
Changed the use of min() as requested in the comments. 

I think the use of max() in the selection of the interface_id in the gateways() list should be unproblematic though and would lead to the selection of an ipv6 over ipv4 address when both are available. 

If this is not wanted, the solution by @antonmolodykh in the comments of the previous commit would of course also work.

lib/utils.py
```diff
@@ -78,8 +78,9 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            interface_id = max(list(gateways()['default']))
+            default_interface = [gateways()][0]['default'][interface_id][1]
```
antonmolodykh: [gateways()][0] == gateways() right?

lib/utils.py
```diff
@@ -78,8 +78,9 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
-            default_interface = grep(netstat('-nr'), '-e', '^default', '-e' '^0.0.0.0').split()[-1]
-            my_ip = ifaddresses(default_interface)[2][0]['addr']
+            interface_id = max(list(gateways()['default']))
```
antonmolodykh: In fact this is not `interface_id` but `address_family_id`

lib/utils.py
p-rintz: Yes, you are correct.  Ill change that, ofc.

antonmolodykh:
LGTM
If @vpetersson  doesn't mind that ipv6 is in priority, then I believe it can be merged.
Thanks, @p-rintz!

Issue #764, opened 2018-03-15
Enable offline mode


antonmolodykh:
We need to add a new item in the settings, something like "enable offline mode". 
By default if there is no connection, it will be offered to select wifi network (using resin wifi), but if this option is enabled, assets will just start to be displayed.

ParrainNN:
Hi @vpetersson !

I will interested by this feature because I would like to use Screenly OSE as a complete standalone solution (no Internet connected) and managed only by mobile devices.  So this feature should propose to enable and configure an access-point (hidden or not / SSID / passphrase / security type / IP range).

Thanks!

vpetersson:
@ParrainNN It's not that simple. You can't have the Pi operate like that without adding additional hardware (namely an RTC). Because of this, we have not pursued this path. In "offline" mode as instructed here, you'd still net a network connection to connect to an NTP server to set the time.

mlott:
@ParrainNN The offline feature you describe is exactly what I'm after.  RTC hardware is cheap and easy to attach and the current time could be updated when a device (ie Mobile) connects to the AP.    I've been updating Screenly OSE by using the Wifi Hotspot function on my mobile then connecting with a second device to upload new content.  I would love to work my own fork, but unfortunately my grasp of python far to limited.  Maybe one day!

urb5tar:
Some times you don't need the exact time. In my case i want to play a video on a exhibition stand. it is just a loop of videos and pictures. the exact time is not important. 

in my opinion there is no need for an offline option. Just display the asset list after 3 minutes.

vpetersson:
In theory, you can do this by setting the start date from 1979 and onwards (i.e. beginning of UNIX time).

urb5tar:
> 
> 
> In theory, you can do this by setting the start date from 1979 and onwards (i.e. beginning of UNIX time).

yes, but the assets wont start, because it is waiting for a network connection.

vpetersson:
We need to understand better what is causing this and if it is in the system or in the Screenly stack. If the former, there might be just some configuration option.

urb5tar:
OK I try to explain it better.

When Screenly starts, it want to have a network connection. if there is no network connection it starts a resin.io AP. Then waits endless till a connection is established. No assets are played till a connection is present.

The demanded "offline mode" is nothing but to start playing the assets whether there is a connection to a wifi or not. In a previous version this works without any further configuration.

albanoalfredo:
[...]

PR #768, opened 2018-03-19
Enable offline mode

antonmolodykh:
Added "Enable offline mode" item.
By default if there is no connection, it will be offered to select wifi network (using resin wifi), but if this option is enabled, assets will just start to be displayed.

Refs #764

### DIFF
--- a/ansible/roles/screenly/files/screenly.conf
+++ b/ansible/roles/screenly/files/screenly.conf
@@ -33,6 +33,9 @@ debug_logging = False
 ; Only display pages with proper SSL certificates
 verify_ssl = True
 
+; Run Resin wifi connect if there is no connection
+enable_offline_mode = False
+
 [auth]
 ; If desired, fill in with appropriate username and password
 user=
--- a/settings.py
+++ b/settings.py
@@ -32,6 +32,7 @@ DEFAULTS = {
         'default_streaming_duration': '300',
         'debug_logging': False,
         'verify_ssl': True,
+        'enable_offline_mode': False
     },
     'auth': {
         'user': '',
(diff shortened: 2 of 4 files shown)


drpaulatkinson:
Thank you for this. I've been struggling to get Resin.io Wifi Connect to work properly, either in Screenly OSE, I also tried using Resin.io and using the OSE container. This mode means I can now deploy and continue playing later...

obewankanobi:
Hello,

Just added this change to my recently installed screenly on my raspberry pi zero and it works as advertise but I notice that even if offline mode is enable, screenly will create a Hotspot regarless of the setting and since its skipping the wifi setup, we have an AP enable with no password.

vpetersson:
@antonmolodykh are you able to reproduce this?

priya2212:
Hi, 

Thanks for adding it! I am also looking to explore this feature for my digital signage. Can you please let me know how to enable this feature to play the asset video in offline mode? I am not sure where to change it to work properly, I tried changing the settings.py file and screenly.conf file but it doesn't make any difference, it shows white screen with SSID and Password given.

vpetersson:
@priya2212 Try switching to the development version to see if that solves the issue (run the upgrade/installer and answer yes to 'Developer version'). I'm not sure this is 100% resolved.

priya2212:
@vpetersson, Thanks for your time & positive response. I tried to give it a try, might be I didn't understand your solution properly. Because I didn't find 'Developer version' option in the Screenly OSE browser, can you please help me with it? 

Thank you.

siludiyi:
> @priya2212 Try switching to the development version to see if that solves the issue (run the upgrade/installer and answer yes to 'Developer version'). I'm not sure this is 100% resolved.

Hi I am thinking if you mind telling me how to enable this mode, I am a newbie. Thanks

Issue #775, opened 2018-03-26
Screen name with spaces


JonnyTech:
If I set the screen name to something with spaces ie "test unit" then all is ok, the page title and management overview display the name correctly - but in the settings page it displays only the player name up to the space, ie "test" rather than "test unit".

PR #777, opened 2018-03-27
Fix spaces in player name

JonnyTech:
Hopefully fixes https://github.com/Screenly/screenly-ose/issues/775 (could it be that easy?)


vpetersson:
Not sure if it fixes it but it's sensible to quote strings regardless. Thanks!

JonnyTech:
Cool - I wish that I understood the code betterer :)

JonnyTech:
Just confirming that this does indeed fix the issue.

PR #811, opened 2018-04-30
Local ingester

watersb:


vpetersson:
@watersb Could you provide a summary of this diff please? Basically what it adds/solves and the expected behavior.

vpetersson:
Looking at the diff, I think it would be better to use the API for this. Sure, it's possible to write directly to the database, but that will break as soon as something changes. If you use the API on the other hand, it will be far more future proof.

watersb:
Absolutely. Actually I created this PR by mistake; I had intended it to be local to my fork as I learned more about the Screenly OSE codebase.

Thanks very much for looking at it! I will clean it up a bit and re-implement via the API.

(Next up, I intend to implement playlists as schedule-able groups. For our museum kiosk, we have six groups, each of which might show four or five short videos.)

I had intended to just roll with bare shell scripts for the OMXPlayer, but the Screenly start-up process, configurable WiFi SSID, and web-app management tool solve lots of pain points that would break this project if I tried to just mail this to the end-user.

vpetersson:
@watersb I'm closing out this issue. Would love to get a PR when you're ready tho.

Issue #527, opened 2016-12-13
Improved web loading - Replace UZBL with something else


vpetersson:
As discussed in #257, it would be desirable to replace UZBL with a more modern browser. Chromium appears to be the best out-of-the-box browser for the Raspberry Pi right now. 

The issue is that it's not as easy to remote control as UZBL, but it's worth exploring.

The primary way to remote control Chromium/Chrome appears to be using the [JavaScript API](https://developer.chrome.com/extensions/api_index). There's also a project called [Chromix-too](https://github.com/smblott-github/chromix-too) which provides a command line tool on top of the JS API.

As a side-note, my initial idea was to use Selenium, but unfortunately [this is not an option](https://stackoverflow.com/questions/31190164/using-selenium-on-raspberry-pi-with-chromium) due to the lack of ARM support in the driver.

Let's evaluate if this is valid in a proof-of-concept branch.

gustavosbarreto:
Hi @vpetersson 

Do you know the ```electric-webview```? It's a scriptable WebView built on top of QtWebEngine which provides a simple protocol over Unix Socket, TCP or WebSocket to communicate with its WebView.

I'm the author of ```electric-webview```. Please let me if you want to evaluate.

https://github.com/gustavosbarreto/electric-webview

vpetersson:
@gustavosbarreto Interesting idea. Have you tested it on the Raspberry Pi? We have been working with QT on the Raspberry Pi in for another project and from that experience, we know that WebKit etc on the Raspberry Pi requires a fair amount of patches to work well.

gustavosbarreto:
@vpetersson Yes, I have tested it on the Raspberry Pi. Works fine. I'm using Yocto to build a image for Raspberry Pi 3.

![15085590_10202547738450405_8873589425774040677_n](https://cloud.githubusercontent.com/assets/86747/22716224/10cc722c-ed7c-11e6-841c-3d1e78a96679.jpg)

vpetersson:
@gustavosbarreto Very cool. Do you have an apt package that we can test with?

gustavosbarreto:
@vpetersson I don't have a DEB package, sorry, but you can build it yourself following the instructions at https://github.com/gustavosbarreto/electric-webview/blob/master/README.md#building

vpetersson:
Cool. Looks straight forward. We will investigate this going forward.

xlegs:
[...]

PR #820, opened 2018-05-18
Moving to new webview

antonmolodykh:
Refs #527


Dockerfile.viewer
```diff
@@ -0,0 +1,57 @@
...
+
+RUN apt-get update && \
+    apt-get -y install \
+        build-essential \
+        curl \
+        git-core \
+        gstreamer0.10-plugins-good \
+        libdbus-glib-1-dev \
+        libffi-dev \
+        libqt5webkit5 \
```
vpetersson: Do we need this given that we have compiled the binary in the other repo?

Dockerfile.viewer
```diff
@@ -0,0 +1,57 @@
...
+        python-netifaces \
+        sqlite3 \
+        x11-utils \
+        x11-xserver-utils \
+        xserver-xorg && \
+    apt-get clean
+
+# Install Python requirements
+ADD requirements.viewer.txt /tmp/requirements.txt
+RUN curl -s https://bootstrap.pypa.io/get-pip.py | python && \
```
vpetersson: Do we really need this on Stretch? I would assume it is running a fairly recent version. it was a workaround on earlier version of Raspbian.

Dockerfile.websocket_server
```diff
@@ -0,0 +1,23 @@
...
+    apt-get -y install \
+        build-essential \
+        curl \
+        libffi-dev \
+        python-dev && \
+    apt-get clean
+
+# Install Python requirements
+ADD requirements.websocket_server.txt /tmp/requirements.txt
+RUN curl -s https://bootstrap.pypa.io/get-pip.py | python && \
```
vpetersson: Same.

Dockerfile.viewer
```diff
@@ -0,0 +1,57 @@
...
+RUN tar -xzf /tmp/screenly-webview-v0.1.tar.gz -C /tmp
+
+RUN cp -r /tmp/res /usr/local/share/ScreenlyWebview/res
+RUN cp /tmp/ScreenlyWebview /usr/local/bin/ScreenlyWebview
+
+# Fix white flickering in omxplayer
+ENV NOREFRESH 1
+
+# Copy X11 configs
+COPY ansible/roles/system/files/10-serverflags.conf /usr/share/X11/xorg.conf.d/10-serverflags.conf
```
vpetersson: Let's move these file out of the Ansible folder to reduce confusion later.

ansible/roles/screenly/tasks/main.yml
```diff
@@ -78,12 +63,64 @@
...
     group: root
 
-- name: Copy screenly systemd units
-  copy:
-    src: "{{ item }}"
-    dest: "/etc/systemd/system/{{ item }}"
+- name: Check if deprecated systemd services exists
+  stat:
+    path: /etc/systemd/system/X.service
+  register: X_service
```
vpetersson: X_service -> x_service

(to keep things consistent)

Dockerfile.viewer
antonmolodykh: yes, the binary file has static links to the files from these libraries

Dockerfile.viewer
antonmolodykh: yes, it is needed on Stretch. pip is not installed by default

vpetersson: Instal the package `python-pip` install.

PR #826, opened 2018-06-02
Use defaults if config file missing.

mscreations:
I ran into an issue where my config file wasn't where it was expected. As it was, Screenly-OSE simply crashed with exit(1) rather than just using the defaults. This PR will fix that issue by loading the defaults and saving the configuration file.


vpetersson:
Nice catch, @mscreations. @antonmolodykh can you take a look as well?

antonmolodykh:
LGTM
Thanks @mscreations

PR #828, opened 2018-06-07
Gui auth setting

mscreations:
I have a client that had an issue with his Screenly OSE device where it was compromised. Basically, someone popped the memory card out and readout the authentication password and then was able to log in remotely after it powered back up. This pull request does a couple of things:

1. Provides an easy way to remotely turn on/off authentication from settings. (Client didn't want to have to remove card to change password)
2. Stores the password as a hash (SHA256) instead of plaintext.

I consciously made it a point to never send the existing password or password hash to the client so if a password is currently set, it will appear blank when you view it. The only improvement I could see would be to add a seed value from the server to the client, but I wasn't really sure how to implement that. 

This is a finished commit, but feel free to comment on whether you think there should be any changes or anything.


vpetersson:
Nice work, @mscreations. 

@antonmolodykh Can you take a look at this guy?

antonmolodykh:
LGTM

vpetersson:
:1st_place_medal:

mscreations:
Just thought of something. This will break anyone's existing installation that has a password setup, because of the hashing.

vpetersson:
@mscreations Oh crap. Can we add a migration quickly?

mscreations:
See #832.

PR #834, opened 2018-06-15
Moving to new webview

antonmolodykh:
* Replaced browser
* server, viewer and websocket-server run in dockers containers
* Travis tests run in a docker container

Refs #527


Dockerfile.viewer
```diff
@@ -0,0 +1,71 @@
...
+RUN pip install --upgrade pip
+
+# Install Python requirements
+ADD requirements.txt /tmp/requirements.txt
+ADD requirements.viewer.txt /tmp/requirements.viewer.txt
+RUN pip install -r /tmp/requirements.txt
+RUN pip install -r /tmp/requirements.viewer.txt
+
+# QtBase from packages does not support eglfs
+RUN curl -L https://github.com/antonmolodykh/Screenly-WebVew/releases/download/0.1/qtbase.tar.gz > /tmp/qtbase.tar.gz
```
vpetersson: We need to change this to the Screenly hosted one.

bin/install.sh
```diff
@@ -35,13 +35,23 @@ if [ "$?" = "1" ]; then
...
+  if [ "$DEV" != 'y'  ]; then
+    export DOCKER_TAG="production"
+    BRANCH="production"
+  else
+    export DOCKER_TAG="latest"
+    BRANCH="master"
+  fi
 else
-  BRANCH="master"
+  export DOCKER_TAG="latest"
```
vpetersson: You're reusing the docker-tag for multiple branches.

Dockerfile.viewer
antonmolodykh: yes, this is what I mentioned several times

bin/install.sh
antonmolodykh: Yes, I suppose to use only "production" and "latest" tags , I see no reason to add others for now. Am I mistaken?

vpetersson:
@antonmolodykh For docker tags, the best practice is to basically "double push" the image under both the branch (latest/experimental/production) as well as under the git hash (`git-$(git rev-parse --short HEAD)"`)

bin/install.sh
vpetersson: I suggest we have the following mapping:

 * master - latest
 * production - production
 * experimental - experimental

PR #835, opened 2018-06-20
Updates install script

antonmolodykh:


bin/install.sh
```diff
@@ -82,8 +92,6 @@ sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*'
 
 cd ~/screenly && git rev-parse HEAD > ~/.screenly/latest_screenly_sha
 
-if test -f ~/.screenly/wifi_set; then  rm ~/.screenly/wifi_set; fi
```
vpetersson: Why is this removed?

bin/install.sh
```diff
@@ -65,7 +74,8 @@ sudo apt-get update
 sudo apt-get purge -y python-setuptools python-pip python-pyasn1
 sudo apt-get install -y python-dev git-core libffi-dev libssl-dev
 curl -s https://bootstrap.pypa.io/get-pip.py | sudo python
-sudo pip install ansible==2.1.0.0
+
+sudo pip install ansible==2.5.3
```
vpetersson: This change might break things.

bin/install.sh
antonmolodykh: because we stopped using this (now we are using `initialized` file)

bin/install.sh
antonmolodykh: hmm, I just copied this from the experimental branch. there it does not break compatibility

antonmolodykh: it seems you have already updated ansible in the requirements file
https://github.com/Screenly/screenly-ose/blob/master/requirements.txt#L3

PR #844, opened 2018-06-23
Add button to download an asset

LorianColtof:


vpetersson:
@LorianColtof Thanks! Could you please also post a screenshot of the feature?

@antonmolodykh Please take a look.

antonmolodykh:
I think we can not accept this PR now. 
@LorianColtof  this functionality should be implemented as API endpoint, not as a page. Now if I click on the download button for a web asset, I get "Internal Server Error" page. For processing this we have `api_response` decorator (see other endpoints). 
In addition, pay attention to these changes break down the layout ("PM" should be on the same line)
image

LorianColtof:
@antonmolodykh I was struggling implementing in the API as it expects JSON and not binary data, but it nevertheless seems possible with `make_response`. So should I just implement it in V1.2 as it's still backwards compatible?

server.py
```diff
@@ -964,12 +964,54 @@ def get(self, command):
...
+                'description': 'The binary content of the asset'
+            }
+        }
+    })
+    def get(self, asset_id):
+        with db.conn(settings['database']) as conn:
+            asset = assets_helper.read(conn, asset_id)
+
+        if isinstance(asset, list):
+            return {'error': 'Invalid asset ID provided'}, 404
```
antonmolodykh: `raise Exception('Invalid asset ID provided')`

because we have [this handler](https://github.com/Screenly/screenly-ose/blob/master/server.py#L344)

antonmolodykh:
Okay, this is much better, but it seems I see a problem. If it fails, we see a new page with the error text. but we need something like this:
image
for this we have an [error handler](https://github.com/Screenly/screenly-ose/blob/master/static/js/screenly-ose.coffee#L532). therefore, the request should be sent as ajax-request.

LorianColtof:
@antonmolodykh How exactly do I reproduce this behavior?

antonmolodykh:
@LorianColtof To see the standard error message (as in my screenshot), you can open the web interface, then turn off the server and try enable/disable an asset.

LorianColtof:
@antonmolodykh So how exactly should this work as an ajax-request? I just created something using Blob, but it seems rather ugly. Another solution would be to first 'test' the url using ajax and then call `window.open`, but that also seems ugly.

antonmolodykh:
@LorianColtof I cannot now say how best to do this. However, definitely we can't open the error message in a new page

LorianColtof:
@antonmolodykh I think this is the least ugly way to solve this problem, and as a bonus the API endpoint will always return JSON, which I think is somewhat neater.

antonmolodykh:
Awesome! a couple of style fixes and it can be merged

server.py
```diff
@@ -964,12 +980,74 @@ def get(self, command):
...
+        with db.conn(settings['database']) as conn:
+            asset = assets_helper.read(conn, asset_id)
+
+        if isinstance(asset, list):
+            raise Exception('Invalid asset ID provided')
+
+        uri = asset['uri']
+
+        if path.isfile(uri):
+            fn = asset['name']
```
antonmolodykh: replace with 'filename'

server.py
```diff
@@ -964,12 +980,74 @@ def get(self, command):
...
+        }
+    })
+    def get(self, asset_id):
+        with db.conn(settings['database']) as conn:
+            asset = assets_helper.read(conn, asset_id)
+
+        if isinstance(asset, list):
+            raise Exception('Invalid asset ID provided')
+
+        uri = asset['uri']
```
antonmolodykh: redundant variable

server.py
```diff
@@ -964,12 +980,74 @@ def get(self, command):
...
+        return result
+
+
 api.add_resource(Assets, '/api/v1/assets')
 api.add_resource(Asset, '/api/v1/assets/<asset_id>')
 api.add_resource(AssetsV1_1, '/api/v1.1/assets')
 api.add_resource(AssetV1_1, '/api/v1.1/assets/<asset_id>')
 api.add_resource(AssetsV1_2, '/api/v1.2/assets')
 api.add_resource(AssetV1_2, '/api/v1.2/assets/<asset_id>')
+api.add_resource(AssetContent, '/api/v1.2/assets/<asset_id>/content')
```
antonmolodykh: v1

static/sass/screenly.scss
```diff
@@ -252,10 +252,13 @@ input[name="file_upload"] {
 #inactive-assets img { display: none; }
 .asset_row_name {
     @include ellip();
-    max-width: 320px;
-    min-width: 320px;
+    max-width: 290px;
+    min-width: 290px;
+}
+
+.asset_row_btns { padding: 6px 8px 10px 8px !important;
```
antonmolodykh: new line

LorianColtof:
Apparently the `atob` function in Javascript was giving the wrong result (in Chrome, at least) and the automatically clicking the link was not working at all in Firefox. Those two things are now fixed as well.

antonmolodykh:
LGTM
thanks @LorianColtof

vpetersson:
Thanks @LorianColtof!

PR #850, opened 2018-07-07
Disable_manage_network install option

antonmolodykh:
Added a new install option that ignores all network stuff

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -50,6 +50,12 @@ else
   BRANCH="experimental"
 fi
 
+echo && read -p "Do you want Screenly to manage your network? This is recommended for most users. (y/N)" -n 1 -r -s NETWORK && echo
+if [ "$NETWORK" != 'y' ]; then
+  export MANAGE_NETWORK=false
+else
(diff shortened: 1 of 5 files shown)


bin/install.sh
```diff
@@ -50,6 +50,12 @@ else
   BRANCH="experimental"
 fi
 
+echo && read -p "Do you want Screenly to manage your network? This is recommended for most users. (y/N)" -n 1 -r -s NETWORK && echo
+if [ "$NETWORK" != 'y' ]; then
+  export MANAGE_NETWORK=false
```
vpetersson: This is a bit inconsistent with the other options. Shouldn't we keep it consistent with an ansible tag?

antonmolodykh: I agree. However, there are some tasks that must be performed *only* if `manage_network=true` and there are some tasks that must be performed *only* if `manage_network=false`. so we have to create two tags for one function. it seems to me more logical to use an environment variable. Or don't you agree?

vpetersson:
Discussed and let's keep it as is for now and then perhaps move all to env variables later.

PR #859, opened 2018-07-11
Fix issue with unicode metadata on uploaded video.

mscreations:
Ran into an issue uploading a movie file created in iMovie. The file had issues decoding unicode stored in the video's metadata. This patch fixes the issue by making it ignore the decoding errors. I only run on a raspberry pi, so I am not sure if the same issue would exist with ffprobe, so I did not include the fix on line 97 as well.


vpetersson:
LGTM

PR #862, opened 2018-07-13
Separate viewer image for each type of device

antonmolodykh:


vpetersson:
@antonmolodykh Let's use one Dockerfile instead and instead use [ARG](https://docs.docker.com/engine/reference/builder/#arg) like this:


```
ARG PI-VERSION=pi3
RUN curl -L \
    "https://github.com/Screenly/screenly-ose-webview/releases/download/v0.1/qtbase-$PI-VERSION.tar.gz" \
    > /tmp/qtbase.tar.gz
```
You can then build with:

```
$ docker build --build-arg PI-VERSION=pi2 .
```

PR #869, opened 2018-07-21
add function to auto-mount usb sticks and add all files as assets

escoand:
I've copied multiple files and snippets together so we can plug usb sticks, auto-mount and add all files as assets to the playlist.
When removing the stick the playlist gets cleaned.


escoand:
I'm completely fine that. Like I said it was just copy and paste. More quick and dirty, proof of concept.
The http invocation especially is absolutely not rock solid...

vpetersson:
@antonmolodykh Can you please take a look at this? Also, here are the outstanding tasks:

- [ ] Lint code according to ShellCheck
- [ ] Make this an opt-in option with a toggle in the UI

Also, perhaps it makes more sense to refactor the asset interaction code to make it a CLI to make it more re-usable. I'm thinking a simple Python script that allows you to do common actions like add/remove assets. Then we can remove all the (rather complex) `curl` commands.  What do you think, @escoand?

escoand:
Yes, a clean Python solution would be much better. Currently the shell code just cutting the JSON string which is more than ugly, no real parsing. If the python code uses http or a local interface it's your decision.

I'm also not really pleased with udev calling a systemd service. Because you need an extra layer (systemd) which is not necessary. I did this because udev seams to be in a chrooted environment and therefore us not able to created the mountpoint folders. Maybe there is a better way.

escoand:
The standalone udev rule (not working because of the chroot) can be found here: https://www.axllent.org/docs/view/auto-mounting-usb-storage/

escoand:
OK, maybe there is a simple restart solution: https://github.com/rbrito/usbmount/issues/2

escoand:
Got a working solution with the udev mount rules. Now you can just replace autoplay.sh with a python script for better asset handling.

escoand:
Maybe we need a clean job for "external" assets if the usb stick was removed when shut down...

rsrini7:
I just tried the autoplay. 

1. when i keep the usb and start rpi, its not booting at all.
2. not showing in viewer any files from usb when it is mounted.
3. It is deleting the existing entries while unblug the pen drive. (some usb port)

escoand:
1. What does "not booting at all" mean?
2. I've just tested plugging while running. I would say while booting the autoplay script was called to early and screenly was not running. This could get better when inserting directly into the screenly database with the python script. Currently the webservice is called.
3. Yes, the asset handling script could be better. But there was an idea to replace it with a Python script. So I did not focus on this. My goal was to trigger the script at all.

So, can you try it when screenly is up and running?

rsrini7:
"Not booting at all" means - i dont have monitor, so by default i enabled ssh during installtion and later installed the realvnc to see the output of the screen. When i insert the USB and try to boot, i am not able to get the IP address of the PI. 

Anyhow, after booting and inserting the USB, i could see the background running service autoplay udev which is trying to mount and unmount. But nothing happens in the viewer screen. In one USB port, the screen was blank for some time and then started displaying the manually uploaded pic. 

I am not sure, how you done this. I am trying to show a demo to my company in a week. If works will update. Thanks for your idea and explanation.

escoand:
The process has 3 steps, maybe you can debug it in your system:
1. mkdir /media/LABEL
2. mount the filesystem there
3. call the script for asset handling
Unplugging is vice versa.

1 and 2 should hopefully work without problems as they were copied from udev and RPi documentations. If so you could call the script by hand and maybe see some output:
```/usr/local/bin/autoplay.sh add sdb1``` or how your dev is called

antonmolodykh:
great idea! However, now when I run the script, I get `Exception: Invalid URL. Failed to add asset.`

escoand:
Currently the shell script is just proof-of-concept - for me it worked. I think the best approach would be a python script using lib/assets_helper.py. This would be much more stable than the current solution.

escoand:
What do you think about this quick-and-dirty solution?

Ohh wait, the imports do not work from this directory.
I'm not the python expert - what do you think is the correct location for the file and the imports?

escoand:
Additionally this is a security question: a system service is calling as root some user libraries.

vpetersson:
Ping @rusko124 - what do you think here? Perhaps we can resurrect this now with the new Celery worker?

rusko124:
@vpetersson Sure! I'll care about it after #987 PR

escoand:
Fine, nice to read that. My progress in this case stopped because I was busy with other things...

vpetersson:
@escoand I think @rusko124 is working on this in paralell in #1010. Could you guys collaborate on that?

escoand:
Already realized this. I try to give comments to his PR.

PR #873, opened 2018-07-26
Use autoremove parameter from ansible instead of command.

strongly-typed:
Since ansible 2.1 this parameter is available.
https://docs.ansible.com/ansible/latest/modules/apt_module.html


vpetersson:
Nice one @strongly-typed. This wasn't available when we wrote the initial code.

Issue #871, opened 2018-07-25
Stuck on Screenly Splash Screen with experimental update


bhagemann:
#### Overview of the Issue

Installed a new image of Sprint7 on SD card.  Booted up on RaspPi.  Configured wi-fi and enabled SSH.  Rebooted the Pi via SSH.  Ran the bash line to upgrade the card to the latest everything (experimental version, upgrade system and so on).  Rebooted.  Got blue Screenly screen which it says on.  Doesn't begin any slideshow.  Can't log in with web front end (get ngnix bad gateway message).  CAN connect with SSH, reboot it... does the same thing.

### Environment

* Raspberry Pi Hardware Version:  Tried on 3 and new 3+, exactly same thing happens
Tried with 2 SD cards, different brands
Tried with 2 power supplies

This only started happening about a week ago or so when I had the experimental version running fine mostly, and ran the bash script to see if there were any development updates it would pick up.  Since then I can't get the latest dev/experimental version running.

vpetersson:
@antonmolodykh can you please take a look?

netjordanlee:
+1

Also getting this issue.

mscreations:
I've run into this issue as well. I was trying to see if the experimental version would work on my Pi Zero, and I found that the device boots and I can ssh into it, but none of the docker containers start up automatically. Issuing a `docker start screenly-ose-websocket && docker start screenly-ose-server && docker start screenly-ose-viewer` does take care of it though.

vpetersson:
@mscreations Interesting find. We have a new update for the Pi Zero in the pipeline, but I don't have an ETA for this yet. The other issue is a bit stranger and needs to be investigated.

antonmolodykh:
I confirm the problem. I don't understand why this is happening, because I did not change anything with regard to launching containers. I'll let you know when I fix it

mscreations:
The weird part is after i manually start the containers the first time, it works fine afterwards. I have had no issues with it starting afterwards. I do however have an issue where the whole pi locks up. No ssh, no web server response, and the viewer stops.

acruzmx:
+1 .. Same here, except that when I do docker start screenly-ose-websocket the container does not exist... Any idea?

vpetersson:
@acruzmx Did you try it with the latest changes to the experimental branch?

acruzmx:
Hi, thanks for asking. I was using the latest update. I ended up reinstalling the old one. 
[...]

PR #875, opened 2018-08-03
Fixed: containers don't start after experimental update

antonmolodykh:
I do not know why this problem arose, I did not change the code to run the containers (this worked several weeks ago).
anyway at the moment the containers are created and are in the "stop" state, they don't start after reboot.

after this PR, systemd services will stop during an update, the containers will be started immediately after pulling.


antonmolodykh:
Refs #871

PR #881, opened 2018-08-21
hyphenate utf8 for older emacs on jessie

umeboshi2:
I had to fix a problem with a screenly install yesterday.  The emacs on the rpi complained about the "utf8" coding, but "utf-8" seems to work well.


antonmolodykh:
Thanks, @umeboshi2 
@vpetersson LGTM

PR #883, opened 2018-08-22
Rm wsgiref

umeboshi2:
It appears that wsgiref is not being used.  Sorry, yesterday's PR forgot to branch from the previous PR.

I have a friend who is installing these into restaurants and other places in the city.  I am probably going to be more involved in the code, since it's gaining popularity here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,5 +26,4 @@ sh==1.08
 six==1.11.0
 uptime==2.0.2
 werkzeug==0.10.4
-wsgiref==0.1.2
 youtube_dl>=2017.7.2


antonmolodykh:
Thanks, @umeboshi2 
LGTM

vpetersson:
Thanks @umeboshi2!

umeboshi2:
@vpetersson Thanks!  It wouldn't install in a py3 venv, and when I went back to look at the code, it wasn't being used.

vpetersson:
@umeboshi2 I'd be surprised if the code base runs in Python 3. probably a fair bit of change required.

umeboshi2:
@vpetersson yes, it seems that there is quite a bit to be changed.  I was under the impression, when I saw six in the reqs.txt that it was being used to help with the transition, but it seems to be an unused requirement.  I went ahead and added travis intregration to my fork, and I will add a py3.5 environment to test against on the ci when I get more tests to pass.  Much of the way of coding for both 2 and 3 is a matter of taste, and there are many different "recipes" for futurizing the code.  If you have the time to look through the "six" branch of my fork, please let me know how you would like the code to look.  I have been using both ```__future__``` and ```six```, when only one strategy may be preferable.

PR #885, opened 2018-08-23
Allow the user to specify the ip address for the device in screenly. …

jishac:
…This allows screenly to be used in "offline" mode.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -78,6 +78,8 @@ def get_node_ip():
         that is being used as the default gateway.
         This should work on both MacOS X and Linux."""
         try:
+            if settings['my_ip'] != '':
+                return settings['my_ip']
             address_family_id = max(list(gateways()['default']))
             default_interface = gateways()['default'][address_family_id][1]
             my_ip = ifaddresses(default_interface)[address_family_id][0]['addr']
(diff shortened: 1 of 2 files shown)


jishac:
Screenly is intended to work in an "offline" mode according to https://support.screenly.io/hc/en-us/articles/115002570873-Can-I-use-Screenly-OSE-in-offline-mode-

Screenly's hotspot can be disabled by creating the file .screenly/initialized and the network interfaces manually configured for local access.

However, the system will not operate without a default gateway set.  This small modification is intended to allow the user to specify a local address as the one to use for the device.  No checks are made against the user supplied address.

antonmolodykh:
I don't quite understand why this is necessary. I mean that in fact, if there is no default gateway, you see "Unable to resolve local IP address." on your screen, and after that Screenly works as usual, that is, it displays local content. right?

jishac:
It did not move past the splash screen, and more importantly I was unable to access the screenly web API to manage or upload content.  I would get a 500 error from the server.

antonmolodykh:
>It did not move past the splash screen

this is because there are no local assets in your playlist (that is, the playlist is not empty, but all assets are web pages)

> I would get a 500 error from the server.

I'm not sure that this is so. the server must be started in any case. I'll check it out again tomorrow, but I'm sure it's true.
anyway, in my opinion your PR makes sense, but I can't decide how good this idea is. What do you think, @vpetersson ?

PR #886, opened 2018-08-24
Upgrade pep8 to pycodestyle

umeboshi2:
It seems that pep8 has been renamed to pycodestyle.  Also, pycodestyle seems to be 
a little more nitpicky than pep8, but using pep8 is deprecated.
https://pep8.readthedocs.io/en/release-1.7.x/

I started this branch upon encountering the ugly code in rtmplite in the tests directory, and noticed that pep8 was being used to check the code.  I have been using flake8 in my python projects, which performs more extensive linting checks, such as unused imports.  It helped warn me about grep and netstat being unused in the ```lib/utils.py``` file, as a result of yesterday's merge using code from netifaces.

In order to help "sixify" the rtmp code (I am not having nearly as many problems with the screenly code), I decided to use ```autopep8``` to help clean it up a bit.  The ```rtmplite``` repo looks like it's been mostly abandoned.  The autopep8 tool has a few features that help to "futurize" the code for python3, and it seemed to be easier to use that, rather than manually digging into that mess.

I am not sure how y'all feel about the styling changes.  I tried to break the lines only when it seemed the best approach, and placed ```# noqa``` comments at the end of the long lines that I thought were probably better left alone.

One of the newer errors flagged by pycodestyle is bare except statements.  Changing those to ```except BaseException``` silences the checker, but doesn't really provide the proper fix.  I changed a few of the bare excepts to be excepting BaseException, but decided to start ignoring those as well.  ```E722``` was added to the ignore option in the ```.travis.yml``` file, however my preference is to not ignore these issues, but actually except the most common error that may occur, and fail on unexpected errors, dealing with those as they arise.  The bare excepts say, "it doesn't matter what goes wrong," which may not always be the case.  This could potentially provide a false positive that the code is working, when something is really broken.

I decided to futurize ```bin/migrate.py``` even though this branch isn't focused on "sixifying" the code, since that seems to be the only place in the screenly code that the print statement is being used.

I'm sorry about the size of this PR.  I integrated travis-ci in my fork to help test this as I was working on it.  The past few commits have been successful on travis, so I decided to go ahead and open this PR, then get back to sixifying the rest of the code.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -28,14 +28,16 @@ HTTP_OK = xrange(200, 299)
 # Travis can run.
 try:
     from sh import omxplayer
-except:
+# FIXME find an expected exception
+except Exception:
     pass
 
 # This will work on x86-based machines
 if machine() in ['x86', 'x86_64']:
     try:
         from sh import ffprobe, mplayer
-    except:
+    # FIXME find an expected exception
+    except Exception:
         pass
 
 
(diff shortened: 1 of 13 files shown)


vpetersson:
@rusko124 Could you please take a look at this PR?

lib/utils.py
```diff
@@ -72,24 +71,18 @@ def validate_url(string):
     """
 
     checker = urlparse(string)
-    return bool(checker.scheme in ('http', 'https', 'rtsp', 'rtmp') and checker.netloc)
+    valid_schemes = ('http', 'https', 'rtsp', 'rtmp')
+    return bool(checker.scheme in valid_schemes and checker.netloc)
 
 
 def get_node_ip():
         """Returns the node's IP, for the interface
```
rusko124: 8 spaces instead of 4, thus, the line length is greater than 79

PR #887, opened 2018-08-26
Is simple http server needed

umeboshi2:
It looks like the spec runner is using server.py and working well.  I don't think that using the simpleHTTPserver is needed.  On the chance that it is actually needed, we can use :8081 in the ```.travis.yml``` and ```phantomjs-runner.js``` files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ To simplify development of the server module of Screenly OSE, we've created a Do
 Assuming you're in the source code repository, simply run:
 
 ```
-$ docker run --rm -ti \
+$ docker run --rm -it \
     --name=screenly-dev \
     -e 'LISTEN=0.0.0.0' \
     -p 8080:8080 \
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ script:
   - find . ! -path "*/rtmplite/*" -name \*.py -exec pep8 --ignore=E402,E501,E731 {} +
   - nosetests -v -a '!fixme' --with-isolation
   - ansible-playbook --syntax-check -i ansible/localhost ansible/site.yml
+  - python -m SimpleHTTPServer 8081 &
   - sleep 3
   - bash static/spec/runner.sh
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ To simplify development of the server module of Screenly OSE, we've created a Do
 Assuming you're in the source code repository, simply run:
 
 ```
-$ docker run --rm -ti \
+$ docker run --rm -it \
```
vpetersson: This is the same thing :)

umeboshi2: I'm sorry about that.  I did that just after I added a travis hook to my local fork, since travis has to wait for a push to start running.  I forgot it was there when I made the PR.  Also, when I tried :8081, I found that it worked ok, but I'm not sure if the ```server.py``` background process was killed by travis.  When ```python -m SimpleHTTPServer 8080 &``` fails to start because 8080 is already used, the ```server.py``` background process is killed by travis at the end of the run, but when I tested with :8081 for just the static server, it seems that the static server process was killed, but the ```server.py``` process wasn't killed.  The ```rtmp.py``` process was killed on both attempts.

Thanks for making this project available!  Y'all have already taught me quite a few things that I have not been doing myself.  I will probably start using ```splinter``` in the future.  The last time I attempted to setup browser testing, it was in a nodejs environment, and I wasn't really happy with it.  I like using python instead.

vpetersson: Glad you like it and thanks for the contributions!

umeboshi2: @vpetersson , I should've mentioned that ```python -m SimpleHTTPServer 8080 &``` only works on python2 and ```python -m six.moves.SimpleHTTPServer 8080``` will not work with either python version on my local machine.  I still don't know what the strategy is for the screenly code to be py3 capable, but using an explicit ```python2``` command should work as long as it is available on travis.

vpetersson:
> I still don't know what the strategy is for the screenly code to be py3 capable, but using an explicit python2 command should work as long as it is available on travis.

The idea is to move to Python 3  at some point, but we would have to finalize the dockerization prior to that, since older versions of Rasbian only ships with Python 2.7.

Issue #316, opened 2015-10-01
Add Google Analytics for ose.demo.screenlyapp.com


vpetersson:
I want to be able to track visitors on the [demo site](http://ose.demo.screenlyapp.com). 

The Google Analytics ID is: UA-37846380-3

Here's what I suggest we do: 
- We introduce a new variable in `settings.py` that is called something like `DEMO`
- We use something like `bool(os.getenv(DEMO, False))` such that we can set it with an environment variable
- In the templates, we add a conditional check for `DEMO`. If it is true, we add the Google Analytics snippet in the footer.

vpetersson:
I found an easier way to do this. Turns out there the file ` /.dockerenv` is always present in Docker. As such, the following snippet should do:

```
def is_running_in_docker():
    return os.path.isfile('/.dockerenv')
```

PR #893, opened 2018-09-03
Demo ga

rusko124:
I added Google analytics if the Screely was launched in the Docker container. This is intended for ose.demo.screenlyapp.com
The Google Analytics ID is: UA-37846380-3

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -225,9 +225,4 @@ def is_demo_node():
     Check if the environment variable IS_DEMO_NODE is set to 1
     :return: bool
     """
-    is_demo_node_env = os.getenv('IS_DEMO_NODE', False)
-    if isinstance(is_demo_node_env, str) and is_demo_node_env == '1':
-        return True
-    else:
-        return False
-
+    return string_to_bool(os.getenv('IS_DEMO_NODE', False))
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -218,3 +218,11 @@ def template_handle_unicode(value):
     if isinstance(value, str):
         return value.decode('utf-8')
     return unicode(value)
+
+
+def is_running_in_docker():
+    """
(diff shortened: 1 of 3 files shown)


vpetersson:
@rusko124 Since we're now moving the general workflow to Docker, let's tweak this instead. Instead of checking for if the file exist, let's instead check for the environment variable `IS_DEMO_NODE=1`. If that is set, then inject the GA snippet. If it is set to 0 or not set at all, skip it.

lib/utils.py
```diff
@@ -220,9 +220,15 @@ def template_handle_unicode(value):
...
-def is_running_in_docker():
+def is_demo_node():
     """
-    Check if running in the docker
+    Check if the environment variable IS_DEMO NODE is set to 1
     :return: bool
     """
-    return os.path.isfile('/.dockerenv')
+    try:
+        if os.environ['IS_DEMO_NODE'] == '1':
```
vpetersson: Use `os.getenv('IS_DEMO_NODE', False)` instead as we want to avoid try/except whenever possible.

lib/utils.py
```diff
@@ -222,13 +222,12 @@ def template_handle_unicode(value):
...
     :return: bool
     """
-    try:
-        if os.environ['IS_DEMO_NODE'] == '1':
-            return True
-        else:
-            return False
-    except KeyError:
+    is_demo_node_env = os.getenv('IS_DEMO_NODE', False)
+    if isinstance(is_demo_node_env, str) and is_demo_node_env == '1':
```
vpetersson: You can simplify this by using [this helper function](https://github.com/Screenly/screenly-ose/blob/master/lib/utils.py#L42-L43).

vpetersson:
LGTM!

vpetersson:
Refs #316.

Issue #816, opened 2018-05-04
Make hotspot and splash page adaptive


vpetersson:
I have a small screen that I use for testing. It has something like 680x480 resolution. However, when I use this screen, I cannot make out the passphrase for the hotspot. I can see the SSID, but not the SSID.

vpetersson:
Still not adaptive enough: 

![img_1380](https://user-images.githubusercontent.com/357664/41776034-22694f7e-761e-11e8-980e-de87d6abd804.JPG)

antonmolodykh:
![image](https://user-images.githubusercontent.com/25369651/42134026-7da747b8-7d55-11e8-99e5-35e908761d56.png)

vpetersson:
I was wrong. The resolution in the image is 800x480.

PR #898, opened 2018-09-07
Hotspot and splash page adaptive.

rusko124:
#816


vpetersson:
@rusko124 Can you also provide a screenshot for 1920x1080, which is the default resolution?

vpetersson:
LGTM!

PR #899, opened 2018-09-11
Make upload text files like html

rusko124:
When i try upload html page, i had error message like "Bad file format".

### DIFF
--- a/server.py
+++ b/server.py
@@ -819,8 +819,7 @@ class FileAsset(Resource):
         req = Request(request.environ)
         file_upload = req.files.get('file_upload')
         filename = file_upload.filename
-
-        if guess_type(filename)[0].split('/')[0] not in ['image', 'video']:
+        if guess_type(filename)[0].split('/')[0] not in ['image', 'video', 'text']:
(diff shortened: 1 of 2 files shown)


vpetersson:
I don't know if we want to support this to be honest. It's not going to work well in most cases, so I think it's better to throw an error.

vpetersson:
@rusko124 This PR doesn't seem to do what it describes. Also, see my comment.

rusko124:
@vpetersson i made mistake this changes for https://github.com/Screenly/screenly-ose/pull/915.

server.py
```diff
@@ -819,8 +819,7 @@ def post(self):
         req = Request(request.environ)
         file_upload = req.files.get('file_upload')
         filename = file_upload.filename
-
-        if guess_type(filename)[0].split('/')[0] not in ['image', 'video']:
+        if guess_type(filename)[0].split('/')[0] not in ['image', 'video', 'text']:
```
vpetersson: I don't like this. Let's remove the text.

Issue #817, opened 2018-05-04
"Re-run network detection" if there is no wifi


antonmolodykh:
We should handle the case when the device does not have wifi. For example, if a user with Pi 2 presses reset button and pulls out the cable, he will see a black screen after rebooting.

PR #915, opened 2018-09-26
If device not supported wifi

rusko124:
PR for #817 
When I tried to reproduce the issue, I found that some edits had already been made in this commit https://github.com/Screenly/screenly-ose/commit/c57937fcf2a3de004b88625f8d71b41dc424c4c5#diff-51ee9359c6f236a2e9710703c82b04be , but also we can add some change for rc.local file.


vpetersson:
Thanks!

Issue #848, opened 2018-07-05
ScreenlyWebView not dismissing alert dialog between assets


abrenoch:
It appears in the latest beta build if an `alert` dialog is shown on a website, it will persist until the device has restarted.

To reproduce simply add a website that shows an `alert` dialog.

 I discovered this trying to debug a locally hosted website, and was not able to get it to go away without needing to reboot (tried to kill the web view process and things began acting up).

I'm running the 'My Michelle' release on a pi 3 b+

rusko124:
@vpetersson 
I confirm that this problem exists, i tried find solves for this problem but at the moment I can not offer a correct and beautiful solution. As far as I know, Screenly is planning to aborted the uzbl browser, so does it make sense now to look for a solution?

vpetersson:
@rusko124 

> As far as I know, Screenly is planning to aborted the uzbl browser, so does it make sense now to look for a solution?

Yes, this is the planned replacement for `uzbl` so it makes sense to spend time on finding a solution here.

vpetersson:
@brandfocus It's already in the works. You can try it out by switching to the experimental branch.

brandfocus:
@vpetersson There seem to be compile issues with Resin on the experimental branch...

`[31m[Error][39m        Error: The command '/bin/sh -c curl -s https://bootstrap.pypa.io/get-pip.py | python &&     pip install --upgrade -r /tmp/requirements.viewer.txt &&     pip install --upgrade -r /tmp/requirements.server.txt &&     pip install --upgrade -r /tmp/requirements.websocket_server.txt' returned a non-zero code: 1
[31m[Error][39m    Not deploying release.`

vpetersson:
Interesting. I don't think we tested the experimental release on Resin yet.

vpetersson:
Should be fixed. Please confirm @brandfocus.

PR #923, opened 2018-10-04
Some fixes for stuck when alert, confirm or prompt

rusko124:
This fixes for #848 issue.
If screenly viewer has a web page asset with an alert, confirmation or promt, the viewer will stuck until we close the interactive box.
These fixes allow you to insert a js code that overwrite alerts, confirms and prompts in most cases, through the event in the config file uzbl. However, there is a little chance that this will not help, and then we catch the event when the browser focus has losted, and at the end of the asset's duration we close the page.

### DIFF
--- a/ansible/roles/screenly/files/uzbl-config
+++ b/ansible/roles/screenly/files/uzbl-config
@@ -1,2 +1,6 @@
 set show_status = 0
-set ssl_verify = 1
+set ssl_verify  = 1
+set on_event    = request ON_EVENT
+
+@on_event   LOAD_START         js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
+@on_event   LOAD_COMMIT        js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
--- a/ansible/roles/screenly/files/uzbl-config
+++ b/ansible/roles/screenly/files/uzbl-config
@@ -1,5 +1,4 @@
 set show_status = 0
-set ssl_verify  = 1
 set on_event    = request ON_EVENT
 
 @on_event   LOAD_START         js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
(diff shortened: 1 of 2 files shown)


ansible/roles/screenly/files/uzbl-config
```diff
@@ -1,2 +1,6 @@
 set show_status = 0
-set ssl_verify = 1
+set ssl_verify  = 1
```
vpetersson: @rusko124 I think this need a way to control this variable since we need to be able to support self-signed certificates.

rusko124: @vpetersson  I checked in the settings we have the use_ssl parameter.
And also we have this code
```
uzbl_rc = 'set ssl_verify = {}\n'.format('1' if settings['verify_ssl'] else '0')
with open(HOME + UZBLRC) as f:  # load uzbl.rc
    uzbl_rc = f.read() + uzbl_rc
```
in server.py
So we can just delete this set line from config.

vpetersson: Got it!

vpetersson:
@rusko124 From what I can tell, all this PR really changes is `browser_event` -> `str(browser_event)` (in addition to renaming "browser" to "browser_event".

rusko124:
@vpetersson this little changes for https://github.com/Screenly/screenly-ose/pull/923/commits/759481a3657e5cfbf0f87f3eea99ef746ca47d4c or are you saying that this whole PR does not make sense?

vpetersson:
> this little changes for 759481a or are you saying that this whole PR does not make sense?

I don't know if it does, or not. But is this the intended change? If it solves a problem, clearly it makes sense.

rusko124:
> > this little changes for [759481a](https://github.com/Screenly/screenly-ose/commit/759481a3657e5cfbf0f87f3eea99ef746ca47d4c) or are you saying that this whole PR does not make sense?
> 
> I don't know if it does, or not. But is this the intended change? If it solves a problem, clearly it makes sense.

Yes, it is this the intended change and also it solves a problem. But what do you think about this solution?

vpetersson:
If it is solve the intended problem, that's great. I just thought it was somewhat of a strange fix.

Issue #924, opened 2018-10-05
Errors associated with 2 installation sources ansible.


rusko124:
#### Overview of the Issue

When i execute "bash <(curl -sL https://www.screenly.io/install-ose.sh)" command in latest version screenly, i get error in ansible-playbook:
```
TASK [screenly : Install pip dependencies] *****************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [screenly : Create default assets database if does not exists] ****************************************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ImportError: cannot import name shell_loader
fatal: [localhost]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
	to retry, use: --limit @/home/pi/screenly/ansible/site.retry

PLAY RECAP *************************************************************************************************************************************************************************************************************************************
localhost                  : ok=48   changed=13   unreachable=0    failed=1   
```
The reason for this error is that we have 2 sources of installation of ansible.
-  [screenly bash script](https://www.screenly.io/install-ose.sh)
-  [requirements.txt](https://github.com/Screenly/screenly-ose/blob/master/requirements.txt)

Ansible playbook have a pip install requirements with ansible.
Since the ansible establishes another ansible, we have errors.

I propose to remove this pip task from the ansible, and in the bash script change
```
sudo pip install ansible==2.5.3
``` 
to 
```
sudo pip install -r /home/pi/screenly/requirements.txt
```
or just delete ansible in requiements.txt

Also wrote about this problem [here](https://www.raspberrypi.org/forums/viewtopic.php?f=41&t=12396&view=unread&sid=a5b959c42ec6f7dcabda190c511fdd87#unread)
@vpetersson what you think about this?

### Environment

* Raspberry Pi Hardware Version: Raspberry Pi 3
* Screenly OSE Version: [latest image](https://github.com/Screenly/screenly-ose/releases/tag/v0.15)

Issue #878, opened 2018-08-15
Solve security vulnerabilities


vpetersson:
[Snyk](snyk.io) reported a number of vulnerabilities. We should update and patch them.

<img width="903" alt="screen shot 2018-08-15 at 2 16 15 pm" src="https://user-images.githubusercontent.com/357664/44149973-d9fb505a-a095-11e8-9fa8-73277d06c701.png">
<img width="887" alt="screen shot 2018-08-15 at 2 16 10 pm" src="https://user-images.githubusercontent.com/357664/44149974-da1c2b90-a095-11e8-9026-2a4ec5b33e9e.png">
<img width="878" alt="screen shot 2018-08-15 at 2 16 03 pm" src="https://user-images.githubusercontent.com/357664/44149975-da3ab8bc-a095-11e8-9bb6-1b0991c1449a.png">
<img width="897" alt="screen shot 2018-08-15 at 2 15 59 pm" src="https://user-images.githubusercontent.com/357664/44149976-da5c9c2a-a095-11e8-9b04-caa50b1bd81c.png">

PR #927, opened 2018-10-08
Remove ansible from requirements.txt

rusko124:
At the moment I was trying to install screenly-ose(with https://www.screenly.io/install-ose.sh) on
RASPBIAN LITE, but it failed.
The reason is this problem #924.
At the moment users also have this problem if choose develop or production branch in script( expiremental branch haven't ansible in requirements files).


vpetersson:
@rusko124 Can you explain why this fix works? You're simply moving Ansible from the regular requirements file to the dev one.

rusko124:
@vpetersson 
Oh sure. This [script](https://www.screenly.io/install-ose.sh) before the start of the ansible-playbook, installs the ansible with the specified version. Then the ansible itself installs another version of the ansible from the requirements.txt, which leads to various errors.
Previously, there were no such errors, since the ansible in script and the ansible in requirements.txt were the same version.
Adding ansible to requirements.dev.txt is necessary for travis cl. Also travis cl no longer indicates the vulnerability of the ansible dependencies(#878).

rusko124:
> A better solution would be to make them consistent IMHO.
> 
I do not quite understand why you need to install 2 times ansible, if ansible participates only for installation. And the requirements.txt itself directly for work screenly's processes.
> Yes, that's ecause it doesn't detect it. Not that it isn't still an issue.
>
If we don't add ansible to requirements.dev.txt, we also get a travis cl error.

PR #931, opened 2018-10-09
Add pause for screenly-viewer container in ansible-playbook

rusko124:
#### Overview of the Issue

When performing a task with creating a screely-ose-viewer container in during the work of the ansible-playbook(experimental branch). the container starts up immediately. It blocks the work of the ansible-playbook. (screenly-viewer running with splash page and also we can't use keyboard) 
These changes pause working of this container until the system is rebooted.
In official ansible documentation we can read about paused tag for docker_container, but this tag not worked there.
If changes state in task, It also does not give the desired result. Since it will be necessary to create services for running containers with system.

#### Reproduction Steps

Steps to reproduce this issue, eg:

* Run ansible-playbook in experimental branch
* Wait create screenly-viewer task

### DIFF
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -124,6 +124,9 @@
       - /home/pi/.screenly:/home/pi/.screenly
       - /home/pi/screenly_assets:/home/pi/screenly_assets
 
+- name: Pause screenly-viewer container
+  command: docker pause screenly-ose-viewer
+
 - name: Create screenly-websocket-server container
   docker_container:
     name: screenly-ose-websocket


vpetersson:
Don't we need to subsequently resume it in the final phase?

rusko124:
@vpetersson
I think about this. But if playbook was run in install script, and we resume in final phase, we also can not reach the end in the script.

vpetersson:
I don't understand why that is a locking operation. Maybe it has been fixed in a newer version of Ansible.

vpetersson:
@rusko124 Isn't this still an issue though?

rusko124:
@vpetersson it's hard to say. So if we start ansible-playbook through ssh, it's work fine.
If we use terminal with ctrl+alt+f1, in build process will start screenly-ose-viewer, and it will be on top of the current installation. So we don't see the process, and when will it end.

vpetersson:
That makes sense. If you pause it, will it automatically start after reboot?

rusko124:
Yes, it will.

vpetersson:
Ok then it's fine.

Issue #841, opened 2018-06-22
Bug in new environment


vpetersson:
#### Overview of the Issue

A paragraph or two about the issue you're experiencing.

#### Reproduction Steps

* Add the asset http://video.blendertestbuilds.de/download.blender.org/peach/trailer_1080p.mov as a URL

You will end up with:

![screenshot from 2018-06-22 13-26-04](https://user-images.githubusercontent.com/357664/41776527-03b6b2f4-7620-11e8-90fc-555edc2e4d76.png)

### Environment

* Raspberry Pi Hardware Version: Raspberry Pi 3 Model B
* Screenly OSE Version: Experimental branch

antonmolodykh:
it seems this link is not valid at this time. however I tested it using another video (http://techslides.com/demos/samples/sample.mov) of the same format and it works fine

PR #935, opened 2018-10-10
Bug associated with determining the type of file

rusko124:
I was able to reproduce this issue #841.
error_1
The reason is the incorrect definition of the file mimetype in js code.
This edits allows to take the default value, if js could not determine the type of file.
Also i have this error when upload file without file extension.
error_3
For this situation i made edits in server python file.


vpetersson:
LGTM Thanks.

Issue #933, opened 2018-10-10
There is no possibility to add a page with requiring js


rusko124:
#### Overview of the Issue

The problem is described in detail https://support.screenly.io/hc/en-us/community/posts/360017643434-URL-error-for-asset.
I reproduced this problem and can confirm it.
![error](https://user-images.githubusercontent.com/17593028/46716156-b9d08500-cc84-11e8-8a05-141b67b5e2fd.png)
When screenly check on validate this url, send GET request and receive the 404 status code, this is due to the fact that javascript is disabled.
@vpetersson Do we need to solve such situations? If yes, then I can suggest using something like this https://github.com/kennethreitz/requests-html

vpetersson:
I think that overcomplicates things. Basically, that's a bug in the webiste/server rather than in Screenly. It should return a 20X. Perhaps a better option is to add an option for something like "Skip asset check."

rusko124:
Got it!

vpetersson:
Closing out. Please ping the original reporter to ensure it worked.

PR #936, opened 2018-10-10
Added option 'Skip asset check'

rusko124:
This option solves the problem described here #933, when user can't add asset due to server error.
For example take https://dashboard.ambientweather.net/devices/public/7034051c8592fabc11be520a5a001203 asset.
screen_1
**Without skip option:**
screen_2
**With skip option:**
screen_

### DIFF
--- a/server.py
+++ b/server.py
@@ -282,10 +282,8 @@ def prepare_asset(request):
     return asset
 
 
-def prepare_asset_v1_2(request, asset_id=None):
-    req = Request(request.environ)
-
-    data = json.loads(req.data)
+def prepare_asset_v1_2(request_environ, asset_id=None):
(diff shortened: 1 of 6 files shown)


vpetersson:
@rusko124 I wasn't thinking about a global option, but rather a tickbox when adding an asset.

rusko124:
single
Now it's only when add asset

vpetersson:
Can you:

 * Change that to a checkbox
 * Make it a bit smaller

rusko124:
edits

vpetersson:
Yes, that's fine too. I was more thinking the alignment between the checkbox and the text, but that's solved in this design so I'm happy with that.

PR #906, opened 2018-09-19
Patch for compiling experimental build on Resin

brandfocus:

Issue #904, opened 2018-09-14
Errors when compiling experimental branch on Resin


brandfocus:
There seem to be issues when compiling Resin on the experimental branch...

```
[31m[Error]�[39m Error: The command '/bin/sh -c curl -s https://bootstrap.pypa.io/get-pip.py | python && pip install --upgrade -r /tmp/requirements.viewer.txt && pip install --upgrade -r /tmp/requirements.server.txt && pip install --upgrade -r /tmp/requirements.websocket_server.txt' returned a non-zero code: 1 �[31m[Error]�[39m Not deploying release.
```

brandfocus:
https://github.com/Screenly/screenly-ose/pull/906

PR #940, opened 2018-10-19
Support for experimental branch build on resin.io

rusko124:
These edits allow build experimental branch on resin.
Build problems was discuss #906 and also #904.


rusko124:
@brandfocus Can you try to build on resin with these edits?

Issue #721, opened 2018-01-05
Unable to interact with the webpage using the keyboard


domlwilliams:
I'm unable to interact with the webpage when using the keyboard. I can see the mouse working and know the keyboard is working as it works when the dev tools is open.
I'm trying to view a site which is behind a login and can only think of manually logging in each time the pi is started.

vpetersson:
Not sure about this one. The keyboard should work. @antonmolodykh any ideas here?

antonmolodykh:
I don't know what's the matter, but I confirm that the keyboard does not work

vpetersson:
I want to make this an option that is disabled by default.

putneyj:
I am also seeing this issue. Is there some way of re-enabling keyboard input? I am also trying to access sites that are behind a log-in screen.

vpetersson:
@antonmolodykh do we have a solution for this?

antonmolodykh:
@vpetersson  I have not found a solution for this yet, but I think that it will not matter soon, as we will replace uzbl

vpetersson:
The more I think about this, I think this is the wrong approach. A better solution would be to use VNC or similar instead.

jonpfeiler:
I am still having an issue with this. Is there any way to enable the keyboard when directly accessing screenly on the pi?

ediphy-azorab:
This still appears to be an issue. Any workarounds available?

nv2k:
> The more I think about this, I think this is the wrong approach. A better solution would be to use VNC or similar instead.

I'm having the same problem. Tried VNC and it's not working either.
Any thoughts?

PR #952, opened 2018-11-02
Ability to use the keyboard as needed

rusko124:
Currently, Screenly-OSE does not have the ability to interact with the page using the keyboard. As we can see in this question #721, it creates some inconvenience for users.
This commit allows you to enable the use of the keyboard via the settings page.
When switching this parameter, the script analyzes which config currently uses the uzbl-browser and when you change "Use keyboard" restarts the browser with the desired config.
keyboard_support
@vpetersson What do you think about it?

### DIFF
--- a/ansible/roles/screenly/files/uzbl-config-keyboard-enable
+++ b/ansible/roles/screenly/files/uzbl-config-keyboard-enable
@@ -1,0 +1,65 @@
+set show_status = 0
+
+# === Core settings ==========================================================
+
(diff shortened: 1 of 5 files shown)


vpetersson:
I'm not too sure about this. Keyboard input in uzbl is all but intuitive. It's not as straight forward as most users would expect it to be. Hence, I think this is somewhat of a confusing feature to add. I much rather add VNC-style input in the experimental branch.

Issue #94, opened 2013-03-07
Make name unique


vpetersson:
I think it's pretty confusing that multiple items can have the same name. Let's make the name unique. If you try to add or edit an asset, an error message should appear.

rusko124:
Resolved. Can we close this?

PR #954, opened 2018-11-12
Unique name to asset

rusko124:
These edits for #94.
A sequence number is added if an asset with the same name already exists.

no_duplicate


vpetersson:
LGTM.

Issue #818, opened 2018-05-10
Feature request - display asset duration


JonnyTech:
I have many still image assets on a random loop. They are all set to display for 10 seconds each. While changing the duration time for them I am losing track of what each is set to (it is a long list!). Can I request that the listing displays the duration of each asset.

vpetersson:
Reasonable feature request.

PR #955, opened 2018-11-12
Add column 'duration' to asset row

rusko124:
These edits for #818.
Also "duration" column, can be done as an option in the settings.

display_asset_duration


rusko124:
I think we can do option selection fields that are needed in the settings. This will enable hide for example Start date and End date, but show only the duration.
What are you think? @vpetersson

vpetersson:
@rusko124 please add the (time) unit for the duration too.

rusko124:
col_duration

rusko124:
duration

vpetersson:
Nice!

PR #959, opened 2018-11-22
resize2fs_once valid url

rusko124:


rusko124:
Also I can offer to completely remove
```
# Install filesystem resizer
wget -O /etc/init.d/resize2fs_once https://github.com/RPi-Distro/pi-gen/blob/master/stage2/01-sys-tweaks/files/resize2fs_once
chmod +x /etc/init.d/resize2fs_once
systemctl enable resize2fs_once
sed -i '$s/$/ init=\/usr\/lib\/raspi-config\/init_resize.sh/' /boot/cmdline.txt
```
as we use pi-gen to build.

vpetersson:
Let's do whatever they do upstream.

PR #967, opened 2018-11-23
Revert "Supporting unicode for player name"

vpetersson:
Reverts Screenly/screenly-ose#962


vpetersson:
@rusko124 this broke the player if player_name is set to null.

rusko124:
@vpetersson Can't reproduce the error
succ

Issue #969, opened 2018-11-27
Stops playing or crash after loading webpage


TheCazz:
Hi I have notice something strange with the latest and the version before that some webpage make screenly just stop working. It show the webpage all the time and I have to do a reboot,

Some page make it crash and instead of show the page it show screenly start screen with the ipaddress and then go to the imgge until is time to show the webpage again it crash and show the screenly start screen.

But some webpage works fine and I can't think of why??

I have also see when it have problem to show the default clock or weather page it stop working too.

I run with a Raspberry Pi B+ with the latest screenly OSE image that I have download from github

rusko124:
Hi, @TheCazz !
Can you provide the url of the web pages you use, please?

TheCazz:
Hi again and thanks for the very fast replay
I Think I understand the problem (Maybe) and that is I'm from sweden and we have å ä ö and in the error log I can see "ascii codec can't decode"
I can give you the whole screenly-viewer but I is going to be a very long one :)

I did try with this page to see what kind of meal a local school is going to have.
https://skolmaten.se/asenskolan/

The other pages that I have problem is on a local server so is not easy for me to show them to you

TheCazz:
[errorlog.txt](https://github.com/Screenly/screenly-ose/files/2619342/errorlog.txt)

Here is the copy and paste log

vpetersson:
@TheCazz looks like you're spot on. We should definitely support unicode characters. Assigning this to the coming sprint.

TheCazz:
Did try with the webpage that stop at the webpage and show nothing more then just the webpage but the log does not say so much 

Nov 27 10:36:46 raspberrypi python[597]: Sleeping for 10
Nov 27 10:36:56 raspberrypi python[597]: Generating asset-list...
Nov 27 10:36:57 raspberrypi python[597]: Showing asset http://addresstothewebpage.php?info=http://pagetotherssinfo/rss/days/ (webpage)

It maybe does not like that I have two http address to connect to the webpage??
The first address is to the PHP server on the local server and the second one is to a webpage that have a RSS feed

TheCazz:
Ahh nice thanks. 
Looking for the update :)

TheCazz:
Just curious about how is going. I mean I'm curious to try and like to know if it fix in the latest development version or how I can see what is new in the development version. New to this so sorry if it sound strange :)

ealmonte32:
@TheCazz -

[...]

PR #970, opened 2018-11-27
Support unicode for title of webpage asset

rusko124:
Refs #969


vpetersson:
@rusko124 can you add some tests for this?

rusko124:
@vpetersson Added, Can you check?

Issue #973, opened 2018-11-28
Server Error: database is locked


Cryjack:
Hi,
we've got the problem shown below while loading the page and even while saving changes. There are 490 active assets used. Is this too much for SQLite? Is there a possibility to change to MySQL or something else?

![image](https://user-images.githubusercontent.com/11329556/49141510-0cc6d000-f2f7-11e8-9933-0c173d1039cc.png)

rusko124:
Hi @Cryjack !
Can you tell which version of Screenly you are using, please?

Cryjack:
Hi rusko124,

yes of course. Sorry.

Just updated today to the latest productive version. But the issue exists even before.

vpetersson:
My guess would be that it's a SD card corruption issue. I'd recommend that you do the following:

 * Export a backup (Settings -> Backup)
 * Get a new SD card
 * Flash out the latest version to the new SD card
 * Restore the backup

Cryjack:
Hi @vpetersson 

the system is working fine, its just the overview of the assets and editing. We have 3 systems and all of them have the same issue. The cards are just 2 months in use so I assume that curruption issues are not the problem.

Cryjack:
Ah, another thing is, that backups don't work with this amount of assets. I assume this could be a setting in webserver (maxfilesize or something).

Cryjack:
The page loads and all assets are shown after an amount of time, but the message is shown every time again.

vpetersson:
Interesting. I'd be keen to solve that. Could you provide some steps to reproduce this?

rusko124:
@vpetersson @Cryjack I will deal with this problem and give you information about it soon

Cryjack:
@rusko124 
Okay great, thank you. If I can do something to help just let me know.

rusko124:
@Cryjack Can you update your Screenly and tell me if this has helped you?

Cryjack:
@rusko124 I've updated on dev-branch and can confirm that the issue is gone. Great job! Thank you very much!

PR #975, opened 2018-11-29
Optimize save ordering

rusko124:
The problem is described here #973 
I was able to reproduce it.
I've added more than 200 assets and this has had a huge impact on performance.
The problem was the ` save_ordering() ` function.
With such a volume of assets, the speed of execution this function ranged from 5-20(!) seconds, which caused the "database is locked" error.

After my edits, the execution speed of the function is about 0.1 seconds.

### DIFF
--- a/lib/assets_helper.py
+++ b/lib/assets_helper.py
@@ -130,12 +130,12 @@ def delete(conn, asset_id):
 
 def save_ordering(db_conn, ids):
     """Order assets. Move to last position assets which not presented in list of id"""
-    assets = read(db_conn)
 
-    for play_order, asset_id in enumerate(ids):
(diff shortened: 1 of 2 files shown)


vpetersson:
@rusko124 Check the code quality feedback

rusko124:
@vpetersson 
Yes, I checked, there are warnings for SQL queries, not the quality of the code. 
SQL queries are similar to those that were previously.
https://app.codacy.com/app/vpetersson/screenly-ose/file/26525013689/issues/source?bid=6090834&fileBranchId=9974861#l17

vpetersson:
Alright, i guess we can leave them for now.

Issue #977, opened 2018-11-30
Backup not working for large amount of Assets


Cryjack:
#### Overview of the Issue

When starting a Backup for a large amount of Assets (>300) the Backup stops working and an empty Messagebox appears. The Backup is not created.

#### Reproduction Steps

Steps to reproduce this issue, eg:

* Press "Settings"
* Press "Get Backup"
* Errormessage like in Screenshot appears

![image](https://user-images.githubusercontent.com/11329556/49280559-ca35fc80-f48a-11e8-998c-9ea3937beadb.png)

### Environment

* Raspberry Pi Hardware Version:
Raspberry Pi 3B+
* Screenly OSE Version:
latest dev (master-branch)

rusko124:
@Cryjack Also checked with assets more than 300, and this works for me.
Are you may not have enough free space?
To better understand the problem, can you show the logs, please?(`journalctl -u screenly-web`)

rusko124:
@Cryjack Got it, than can you provide information about free space on sd and size of directory "screenly_assets", please? (without screenly-backup.tar.gz)

Cryjack:
@rusko124 

"df -h" shows:
filesystem size used avail use% mountedon
/dev/root       7.3G  2.6G  4.5G  37% /

"du -hs /home/pi/screenly_assets" shows:
506M    /home/pi/screenly_assets

As far as I can see, no space problem.

rusko124:
Hi, @Cryjack ! Can you update and tell me if this has helped you?

Cryjack:
@rusko124 
I updated and checked the version.

The Webpage stays the same und does not change. But I can see, that the backup tar was created just fine in the filesystem. There is just no output on Webpage.

rusko124:
@Cryjack it's strange, the changes above should have helped.
An error without text still appears?

Cryjack:
@rusko124 

Yes, the same messagebox without content appears.

Cryjack:
The error on the website occurs while not finished creating the backupfile (still grows).

rusko124:
@Cryjack Hm, can you provide last commit hash, please?
You can also see the error in the web browser console(it should be associated with a timeout).
I can also offer you to open the Screenly web page in incognito mode,  there is a possibility that the JS script is cached.

Cryjack:
@rusko124 
No problem, here are the logs.

Here is the commit hash:
commit df48d332a2611ee2388825b87777da78fceb4e14
[...]

PR #980, opened 2018-12-03
Set large timeout for backup

rusko124:
Discussion of the issue #977

These edits make fix when it was not possible to download a backup if .tar is generated more than 30 seconds.
Now the timeout is 30 minutes.


vpetersson:
LGTM.

Issue #876, opened 2018-08-13
Deprecate cronjobs in favor of worker


vpetersson:
As pointed out in #874, some features will not work in resin/dockerized environment. To solve this, we need to introduce a worker that can run inside a docker container.

I suggest that we use [Celery](http://www.celeryproject.org/) and port all cronjobs to run there..

Marco-LB:
Hello, is it possible to have a tutorial? instructions how to create a command to execute on startup? hdmi-cec on startup?

vpetersson:
@Marco-LB sure, take a look [here](https://www.screenly.io/blog/2017/07/02/how-to-automatically-turn-off-and-on-your-monitor-from-your-raspberry-pi/).

Marco-LB:
thanks for the quick reply, but cron and crontab know how to use them, but celery is the problem

vpetersson:
Well, this isn't implemented yet. That's what this ticket is tracking.

Marco-LB:
can this post help? [https://github.com/resin-io-playground/cron-example](url)

vpetersson:
We know how to do it. It's fairly straight forward. We just haven't implemented it yet. It is scheduled for the next sprint.

Marco-LB:
OK, Thanks for your information

rusko124:
This issue was solved in [Celery](https://github.com/Screenly/screenly-ose/tree/celery) branch(PR https://github.com/Screenly/screenly-ose/pull/987#issuecomment-450341787).
The subsequent discussion, I recommend to move here #1081

@vpetersson I think we can close this.

PR #897, opened 2018-09-06
Fix backup and recover

rusko124:
Fixed backup and recover for resin.io and Docker.

#800

Issue #734, opened 2018-01-26
Simplify upgrade flow


vpetersson:
One pain point for a fair amount of users is the upgrade process. It would be desirable to be able to run the upgrade from within the user interface.

vpetersson:
# Overview

The goal of this task is to be able to upgrade Screenly OSE from within the user interface. However, it quickly get complicated when we factor in the security elements. Hence, this overlaps with #876.

## Requirements

* The goal is that both viewer and server shall run as a non-privilegued (and without ability to run `sudo` for arbitrary commands). We should assume this for the sake of the upgrade process.
* Since the upgrade will require root to run, we need to somehow run this process through another runner.
* The process runner should only be able to run whitelisted commands (and not arbitrary commands)
* The process runner should be owned by root. 
* The system should be compatible with both a dockerized system and a non-dockerized system.
  * We can ignore a resin/balena-environment, since the upgrades are handled differently there.
* It should be possible to launch the upgrade from within the user interface and see the progress (as well as potential errors).

## Possible solution

* We implement a task runner with Celary as described in #876. This task runner needs to run directly on the host system through `systemd` (even in the case of a dockerized environment because of the `sudo` solution below)
* We alter `/etc/sudoers` such that the user `pi` requires password to run arbitrary commands.
  * **Note:** This will also require us to run `tvservice` and similar through this mechanism.
* We create a new script called `/usr/local/sbin/upgrade-screenly.sh`, which is a hard coded script that only runs the installer. This file should have the permission 0700 and owned by root to improve security. 
  * It is essential that this script does not simply call on a script in `~/screenly` that is owned by pi, as that would allow an attacker to execute any command as root.
  * We should also probably fetch this script directly from Github from the master branch instead of copied from the `~/screenly` folder to improve security further.
* We add a rule to `/etc/sudoers` that allows the user pi to run `/usr/local/sbin/upgrade-screenly.sh` without a password.
* With all those pieces in place, we can now create a Celary task that calls on `upgrade-screenly.sh` through `sudo`, which in turn executes the upgrade script.

vpetersson:
@sublimino - I'm actually really curious about your feedback here.

rusko124:
This issue was solved in [Celery](https://github.com/Screenly/screenly-ose/tree/celery) branch(PR https://github.com/Screenly/screenly-ose/pull/987).
The subsequent discussion, I recommend to move here #1081

@vpetersson I think we can close this.

Issue #121, opened 2013-04-14
Basic system controls via the Screenly web interface


git-pi:
I was unable to get into the SSH interface of Screenly today (for some reason), though I could still hit the web interface successfully and ping both interfaces (eth0 and wlan0) perfectly well..  Rather than hauling out a keyboard and monitor, it would have been nice to have a "Restart Screenly" and/or "Shut Down Screenly" buttons on the main interface page that would launch the appropriate Linux commands in the background.

As an aside, Screenly had stopped displaying after 18 hours or so, hence my need to restart it.  Don't know what version I'm currently on - see other message regarding that...  ;-)

riccla0321:
+1 to the password protected web interface.

xbmcnut:
+1. Reboot, password protected back-end an build number displayed in 'System Info' would be greatly appreciated.

danbsfc:
This would be fantastic! Having buttons for Shutdown & Reboot, along with password protection would make it a virtually perfect piece of kit. Got 2 stations running Screenly OSE now and highly recommend it to anyone who asks.

krzimmer:
Being able to power cycle the pi via web interface would be a very nice feature.  As far as web interface protection, I'm using iptables to allow access to only trusted hostnames.  This is a fine security mechanism for now, but I would still like basic username:password authenication.  I don't know how to implement this in an elegant way.  My current plan to add basic authentication is to install Apache.  Apache would at as an authenication proxy to the Screenly webserver.  I would then block direct access to the Screenly webserver.

markdegrootnl:
+1

plittlefield:
+1

toussa:
+1

vpetersson:
Blocked by #876.

rusko124:
This issue was solved in [Celery](https://github.com/Screenly/screenly-ose/tree/celery) branch(PR #1011 ).
The subsequent discussion, I recommend to move here #1081

@vpetersson I think we can close this.

PR #987, opened 2018-12-05
Added celery worker

rusko124:
Discussion here #876
This is just a sketch for how Screenly can work with Celery.
At the moment, no tasks and support for Docker and Resin.io


vpetersson:
@rusko124 good start, but we're missing a few things #734 in the skelleton design.

rusko124:
I will combine this question #734, as it depends on Celery.
In this large commit, I used notes from https://github.com/Screenly/screenly-ose/issues/734#issuecomment-441737456
I'll leave comments about those or other points in the code.

The problem is updating when we enable password prompt for sudo. This is solved by running the update with sudo, but without it we will get errors. The update script was also corrected.

I understand that we still need changes, but now we can update through the web interface(just only for master branch).

ansible/roles/screenly/tasks/main.yml
```diff
@@ -70,6 +70,14 @@
     job: "/usr/local/bin/screenly_utils.sh cleanup"
     user: pi
 
+- name: Copy upgrade_screenly.sh
```
vpetersson: This is an attack vector. A user could simply add any arbitrary comand into this file and now they can execute anything as root.

ansible/roles/system/tasks/main.yml
```diff
@@ -241,3 +240,18 @@
   file:
     path: /var/swap
     state: absent
+
+- name: Remove pi-nopasswd file
```
vpetersson: 👍

ansible/roles/system/files/010_pi-nopasswd
```diff
@@ -0,0 +1 @@
+pi ALL=(ALL) NOPASSWD: ALL
```
vpetersson: This is wrong. We should require password. We then need to whitelist selected few commands.

vpetersson: Fair enough, but we still need another one then for the locked down state.

rusko124: The task above just deletes this file (010_pi-nopasswd file for disabling sudo password in raspbian by default)

ansible/roles/screenly/tasks/main.yml
rusko124: Updated

rusko124:
So that I added above.
- I changed celery broker from rabbitmq to redis
- Now the update script that is used through the web interface is taken directly from the repository
- Added docker-compose for celery health in Docker environment.
- Added support for celery in resin.io
- Added authentication to run system commands through the web interface. This will provide greater security. This is also needed for this #121. You cannot run these commands without enabling this option.

bin/upgrade_screenly.sh
```diff
@@ -0,0 +1,90 @@
...
+    export DOCKER_TAG="latest"
+    BRANCH="master"
+#elif [ $1 = "experimental" ]; then
+#  export DOCKER_TAG="experimental"
+#  BRANCH="experimental"
+#elif [ $1 = "production" ]; then
+#    export DOCKER_TAG="production"
+#    BRANCH="production"
+else
+  exit 1
```
vpetersson: Maybe add some logging here?

vpetersson:
Looking good, @rusko124. What's left in the initial scope?

rusko124:
@vpetersson 
Um, I think we just need to work on it:
- run_upgrade.sh for manual start
- upgrade_screenly.sh to update via web interface

Will we use run_upgrade.sh script through sudo or something else better?

Dockerfile.worker
```diff
@@ -0,0 +1,17 @@
+FROM python:2.7-stretch
+MAINTAINER Viktor Petersson <vpetersson@screenly.io>
```
vpetersson: This is a new file. You'd be the maintainer ;)

rusko124:
List of changes in the last commit:
- Instead of 2 files updates now is one
- Manual run of the update for the user remained unchanged
- Setting a required sudo password is now directly in the script itself, since it is not possible to do this in the ansible.
- Some edits

rusko124:
@vpetersson Could you look it up when you'll have time? And if everything is as you would like to see it then I will work on the quality of the code.

server.py
```diff
@@ -40,20 +42,50 @@
 from lib.utils import validate_url
 from lib.utils import is_demo_node
 
-from settings import auth_basic, CONFIGURABLE_SETTINGS, DEFAULTS, LISTEN, PORT, settings, ZmqPublisher
+from settings import auth_basic, auth_system, CONFIGURABLE_SETTINGS, DEFAULTS, LISTEN, PORT, settings, ZmqPublisher
 
 HOME = getenv('HOME', '/home/pi')
 DISABLE_MANAGE_NETWORK = '.screenly/disable_manage_network'
+CELERY_RESULT_BACKEND = getenv('CELERY_RESULT_BACKEND', 'redis://localhost:6379/0')
```
vpetersson: Is there a scenario where `CELERY_RESULT_BACKEND` and `CELERY_BROKER_URL` are different? If not, I suggest we consolidate them to `CELERY_BROKER`

rusko124: Hmm I think that variables likely will not different, but the result and broker have a different purposes. So I think we don't need to merger them into one variable, it can be confusing.

Dockerfile.worker
rusko124: Got it :)

rusko124:
@vpetersson 
Could you look at this commit https://github.com/Screenly/screenly-ose/pull/987/commits/9c6888da425a5b57d8e7cccde62de8f6267944f2 ?
There I have combined a few issues:
- https://github.com/Screenly/screenly-ose/issues/876
- https://github.com/Screenly/screenly-ose/pull/897#issuecomment-419362352
- Added stop and pause commands for the viewer
- Removed screenly_utils.sh 
- Added support for periodic tasks to celery

vpetersson:
@rusko124 Let's createa  new branch so that we can test it easier. It's a pretty large change so we should thread lightly.

PR #997, opened 2018-12-22
Added pause command

frankdeath:
Added a pause command that allows a user to pause & unpause the slideshow.  The implementation doesn't modify the duration of any assets.


vpetersson:
@rusko124 Could you please take a look. 

Also, @frankdeath I'd like to hold this until we've landed #987 since it does a pretty big overhaul on relevant components. In short, these commands would have to be moved to the celery worker.

frankdeath:
> Also, @frankdeath I'd like to hold this until we've landed #987 since it does a pretty big overhaul on relevant components. In short, these commands would have to be moved to the celery worker.

That sounds very reasonable.  This pull req isn't urgent and I don't have strong feelings about the specifics of the implementation.  I just needed a way to pause a slideshow and thought someone else might find my modifications useful.

vpetersson:
@rusko124 ping.

Issue #995, opened 2018-12-18
Assets not showing 


DaveEps:
Hi i am new on here and fairly new to Screenly !
i had screenly up and running on an old Raspberry pi for a few months 
and it worked so well with my website , i thought i would upgrade to the latest Pi ,
as the old Pi was very slow .
my main asset is a link to my gallery on my website , which just shows images to any customers coming in.
but for some reason it will not load on the new one ?
so i have tried numerous website testing them and some load and some dont !
for example , Facebook home page worked but,
facebook with the extension to my page wouldnt load ?

sorry if this is a duplicate of an other issue but i have searched for other similar problems but cant find anyone with the same .
i have attached a log of when i turn the asset on .
be much appreciated if someone could help .
thanks in advance 
[screenly.pdf](https://github.com/Screenly/screenly-ose/files/2689866/screenly.pdf)

vpetersson:
Hard to say, but please fill out the steps for reproducing the issue as per the template.

DaveEps:
Thanks for the reply back ,
also sorry i deleted that part thinking i didnt need it !!

i have Raspberry Pi 3 Model B+ which i installed the newest version of screenly from NOOBS_v2_8_2 .

For example to reproduce what i have done .

*added asset - webpage = https://www.facebook.com/
*turn on asset - waited for like a minute
*facebook homepage shows up on screen (works perfect)
*turn off asset

*added asset - webpage = https://www.facebook.com/EPSbn9/
*turn on asset - waited for like a minute
*screen flashes white over and over again like its almost not connecting to the website / or website is taking to long to respond .
*turn off asset- stops flashing and goes back to blank page with screenly logo.

to note : https://www.instagram.com/?hl=en - worked but when i added my instagram page 
https://www.instagram.com/epsbn9/?hl=en it didnt work 

this also doesnt work with my website http://expandedpoly.co.uk/ - 
but i have just found out i have a problem with my site ! which is getting fixed as we speak .

i have also added several other webpages to test which worked fine ; 
https://www.ebay.co.uk/
http://www.xmasclock.com/
https://www.instagram.com/?hl=en
https://github.com/Screenly/screenly-ose/issues

if you need any other info that i might of missed let me know .
and thanks in advance for the help .
Dave

DaveEps:
Just to note : during installation of screenly ,
the raspberry Pi turned off twice due to lack of power ! i also had a flashing lighting bolt sign in the corner of the screen .
[...]

PR #1008, opened 2019-01-08
SIGALRM when viewer stuck

rusko124:
Thanks to this issue #995, I found next problem
Reproduce steps:
- Activate asset with this url https://www.instagram.com/epsbn9/?hl=en
- Disable asset above
- Try activate asset above one more time.
- Then we'll see that we're stuck on black_page.html

These edits have:
Send a signal SIGALRM in 10 seconds when we started to receive data from the browser process. If the browser process does not receive any data during this time, an exception will be thrown out and the viewer will continue to work.

### DIFF
--- a/lib/errors.py
+++ b/lib/errors.py
@@ -1,0 +1,2 @@
+class SigalrmException(Exception):
+    pass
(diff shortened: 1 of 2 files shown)


vpetersson:
Thanks, @rusko124!

PR #1010, opened 2019-01-08
Auto-mount usb and add usb assets

rusko124:
This PR continuation #869 PR.

I'm interested in removing usb of assets, in special cases, when the usb was turn off and assets remained.
@vpetersson Can you help with it?


vpetersson:
@rusko124 I think we need to think about the attack vectors a bit more here. 

Let's think about the following scenarios:

 * What happens if someone malicious plugs in a USB stick into the device? Will it just add these assets and play them? 
 * What happens if someone adds a USB stick with 5000 assets? Do we copy them to disk? If so, it could eat up all available disk space. 
 * If we do not copy them to disk, what happens if someone removes the USB stick? Do we then leave those records in the database?  (Think about what will happen to the database without any kind of vacuuming process).

escoand:
>     * What happens if someone malicious plugs in a USB stick into the device? Will it just add these assets and play them?
If we think about somebody malicious plugging something in, I would say images on USB sticks are the least critical thing...
 
>     * What happens if someone adds a USB stick with 5000 assets? Do we copy them to disk? If so, it could eat up all available disk space.
My previous solution used the absolute path to the file and so not copying them. If I'm right the code here is doing the same.

>     * If we do not copy them to disk, what happens if someone removes the USB stick? Do we then leave those records in the database?  (Think about what will happen to the database without any kind of vacuuming process).
When unplugging we can clean up the database because we know the removed path and have this in the database. But I think we need some kind of housekeeping which cleans old entries if USB sticks have been removed when powered off.

ansible/roles/screenly/tasks/main.yml
```diff
@@ -82,6 +82,41 @@
...
+    src: /lib/systemd/system/systemd-udevd.service
+    dest: /etc/systemd/system/systemd-udevd.service
+
+- name: Configure systemd-udevd service
+  lineinfile:
+    dest: /etc/systemd/system/systemd-udevd.service
+    regexp: '^MountFlags='
+    line: 'MountFlags=shared'
+
+- name: Installs udev restart service
```
escoand: You've forgotten the same thing like me: activating the service.
Just add to ansible/roles/screenly/vars/main.yml

server.py
```diff
@@ -106,6 +106,25 @@ def shutdown_screenly():
     sh.sudo('shutdown', 'now', _bg=True)
 
 
+def append_usb_assets(mountpoint):
```
escoand: I don't know celery but do we need `@celery.task` here as well?

server.py
```diff
@@ -447,6 +466,33 @@ def get(key):
...
+
+    if not filetype:
+        return
+
+    if filetype not in ['image', 'video']:
+        return
+
+    return {
+        'asset_id': uuid.uuid4().hex,
+        'duration': int(get_video_duration(filepath).total_seconds()) if "video" == filetype else 10,
```
escoand: shouldn't we use the default duration for images and so null here?

server.py
rusko124: You're right. I lost it when I was solving merge conflicts. Thanks!

vpetersson:
> > ```
> > * What happens if someone malicious plugs in a USB stick into the device? Will it just add these assets and play them?
> > ```
> 
> If we think about somebody malicious plugging something in, I would say images on USB sticks are the least critical thing...

Not sure I agree. Picture the device sitting unsupervised in a retail environment. Now, that's clearly not an ideal situation, but a real possibility. It would then be trivial for a malicious user to show innapropriate images on the screen. Hence I think we need some kind of failsafe here, such as activating the import in the UI or at the very least a toggle that is set to disabled by default for auto imports. 

Alternatively, we could generate a random "key" in the user interface (that could be bundle into a file say 'screenly.ini` that has the key in the payload). Then, unless this key file is present on the drive, Screenly will ignore the USB drive.
 
> > ```
> > * What happens if someone adds a USB stick with 5000 assets? Do we copy them to disk? If so, it could eat up all available disk space.
> > ```
> 
> My previous solution used the absolute path to the file and so not copying them. If I'm right the code here is doing the same.

Gotcha. That mitigates that issue.

> > ```
> > * If we do not copy them to disk, what happens if someone removes the USB stick? Do we then leave those records in the database?  (Think about what will happen to the database without any kind of vacuuming process).
> > ```
> 
> When unplugging we can clean up the database because we know the removed path and have this in the database. But I think we need some kind of housekeeping which cleans old entries if USB sticks have been removed when powered off.

Gotcha. Ok let's think about some kind of periodic vacuuming script that we can run through Celery then on a daily basis.

escoand:
> Hence I think we need some kind of failsafe here, such as activating the import in the UI or at the very least a toggle that is set to disabled by default for auto imports.
Sounds good to me and simple for the user to handle.

I head an USB keyboard or even more malicious gadgets in mind. At this point we don't have to think about some inappropriate pictures.

Hmm, when I think about this maybe your second solution is better...

vpetersson:
> I head an USB keyboard or even more malicious gadgets in mind. At this point we don't have to think about some inappropriate pictures

Right, which is why we're starting to lock things down. In the future, we'll drop permission on the runtime user as well moving to framebuffer such that there is no X11/Xorg session to drop out into.

rusko124:
@vpetersson @escoand 
So guys, edits:
- Added periodic task cleaning usb assets(every hour)
- Added feature to add USB assets by file with key on USB
- Added tests
- Some fixes

Screenshots:
screen_1
screen_2

tests/celery_tasks_test.py
```diff
@@ -1,25 +1,85 @@
...
+
+        settings['usb_assets_key'] = generate_perfect_paper_password(20, False)
+        settings.save()
+
+        key_data = {"screenly": {"key": settings['usb_assets_key']}}
+        with open(self.key_file, 'w') as f:
+            yaml.dump(key_data, f)
+
+    def tearDown(self):
+        remove(self.key_file)
```
rusko124: Need to add `remove(self.cleanup_asset_file)`

vpetersson:
Very nice, @rusko124. Good job! Could you take a look too @escoand?

escoand:
How can i give it a try easily?

vpetersson:
@escoand You can check it out locally:
https://help.github.com/articles/checking-out-pull-requests-locally/

escoand:
Good work, seems to work very well. But would you really want to use a yaml file?
It's not described in the gui what to do with it. A simple file as txt and the code in it would be simpler for non-experienced users.

escoand:
And I think a notification in the gui and on screen when the key is missing would be also nice.

escoand:
What do you think?

rusko124:
>Can the yaml file be expanded to activate the assets as well? Maybe there are more options which could be helpfull.

I think the yaml file extension is well, but if you have suggestions, I'll be glad to hear it
>And I think a notification in the gui and on screen when the key is missing would be also nice. And also how many assets were added and that they have to be activated.

It would be convenient, but I think it's need to put off for later.

@vpetersson We need your opinion

escoand:
> I think the yaml file extension is well, but if you have suggestions, I'll be glad to hear it

Currently I have this in mind:
```yaml
activate: true
duration: 10s | 5m | infinite
```

rusko124:
@vpetersson 
>That seems like a reasonable feature. Could you guys provide a full example of the YAML?

usb_assets_key.yaml
```
screenly:
  key: "d9uaHYGfGfZ97pmzZwx3"
```
usb_assets

vpetersson:
Looks good to me, @rusko124.

escoand:
Any progress in mainlining this? 2 more options for the yaml file came to my mind so you have (nearly) full control without connecting to the device.

So my proposal would be (the first would be my recommended default):
```yaml
screenly:
  key: string
  activate: true | false
  duration: 10s | 5m | infinite
  play_for: week | day | month | year | forever | manual
  begin_date: timestamp | string
  end_date: timestamp | string
  copy: false | true (copy files to internal memory, but this will fill up the device sometimes)
```

rusko124:
@escoand looks good!
I'll be back to work on Screenly OSE soon, and I'll continue to work on this request.

rusko124:
@escoand @vpetersson What you think?

usb_assets_key.yaml
```
screenly:
  key: "KJjjxytgaWTRwvZ82Hc8"
  activate: false
  copy: false
  start_date: 1552586400
  end_date: 1553191200
  duration: 10
```

usb_1
usb_2

escoand:
Can we use a more user friendly time format? Unix timestamps are hard to type by hand

escoand:
With this it would be perfect for me.

rusko124:
@escoand Sure, It won't be a problem, but I'd like to wait for any comments from @vpetersson

vpetersson:
@rusko124 / @escoand  I think this needs to be documented too.

PR #1011, opened 2019-01-08
Added some system commands for the web interface

rusko124:
Discussion #121

Added some system commands for the web interface.
Such as:
- reboot 
- shutdown

system_controls


ealmonte32:
@rusko124 -

Wouldn't the following make the users have to physically go and power cycle the Pi to turn it back on?
```
@celery.task
def shutdown_screenly():
    """Background task to shutdown Screenly-OSE."""
    sh.sudo('shutdown', 'now', _bg=True)
```

There should be a warning to users that they are shutting down the Pi and not just screenly-viewer..

vpetersson:
@ealmonte32 Yes, you are right. This would require physical intervention. I agree that it would make sense to have a secondary warning. Could you create a different ticket for this?

BavoB:
Newbie question: does the 'Reboot Screenly' command reboot just the Screenly program running on the Raspberry Pi or does it reboot the whole OS.
I'm asking because I get rather frequent remarks that the video output does not work and would like to try the reboot command to see if that works + it's something that the user can easily be directed to do him-herself...

ealmonte32:
> Newbie question: does the 'Reboot Screenly' command reboot just the Screenly program running on the Raspberry Pi or does it reboot the whole OS.
> I'm asking because I get rather frequent remarks that the video output does not work and would like to try the reboot command to see if that works + it's something that the user can easily be directed to do him-herself...

It reboots the whole OS/pi.

BavoB:
Thanks, are you aware of other measures that can be taken against the fact that the TV does show black image even with the correct HDMI input is chosen?
It always happens when I’m not there of course.
I presume the TV is woken up from sleep state or even off state. The Raspberry is on all the time.

Regards,
Bavo

ealmonte32:
> Thanks, are you aware of other measures that can be taken against the fact that the TV does show black image even with the correct HDMI input is chosen? It always happens when I’m not there of course. I presume the TV is woken up from sleep state or even off state. The Raspberry is on all the time. Regards, Bavo

Have you tried anything mentioned in the raspberry pi video documentation with regards to HDMI compatibility and troubleshooting? https://www.raspberrypi.org/documentation/configuration/config-txt/video.md

```
config_hdmi_boost
Configures the signal strength of the HDMI interface. The minimum value is 0 and the maximum is 11.
The default value for the original Model B and A is 2. The default value for the Model B+ and all later models is 5.
If you are seeing HDMI issues (speckling, interference) then try 7. Very long HDMI cables may need up to 11, but values this high should not be used unless absolutely necessary.
This option is ignored on the Raspberry Pi 4.
```

You can try some of the things mentioned there.

Also, a monitor showing black image vs turning off are two different things, I would need more details, and logs, it might be a totally different issue than with the HDMI. You might simply be having the issue with memory consumption by having some websites that cause the python process to be killed and not restarted.
If that is the case, I would recommend you run the installer again, but until it is approved by viktor you need to either edit the viewer.py yourself or simply run this fork version mentioned in my comment here https://github.com/Screenly/screenly-ose/issues/1355#issuecomment-696388611 that includes all fixes, such as the memory one, select the master/dev branch, and see if that helps.

Issue #971, opened 2018-11-27
Add support for displaying the Raspberry Pi version/model in the user interface


vpetersson:
I'd like to add the ability to display the Raspberry Pi version in the system info under System Info.

We can get the hardware version from here:

```
$ grep Revision /proc/cpuinfo
Revision        : a01041
```

Then we simply need to parse this, and turn it into the model number using [this list](https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md).

dig1talmad:
While you're at it, could you display the version of Screenly in System Info too?

vpetersson:
Good idea @dig1talmad.

qbicdesign:
yay

qbicdesign:
can we get MAC address(es) added too please?

vpetersson:
@qbicdesign Yes shouldn't be very hard. I guess we need to display both for WiFi and Ethernet.

rusko124:
Raspberry Model / Screenly Version / MAC address were added to the system info page.
@vpetersson I think we can close this.

PR #1015, opened 2019-01-09
Added 'Raspberry Model' and 'Screenly version' for system info page

rusko124:
Discussion #971
system_info

### DIFF
--- a/lib/diagnostics.py
+++ b/lib/diagnostics.py
@@ -5,6 +5,7 @@ import os
 import sh
 import socket
 import sqlite3
+import re
 import utils
 from pprint import pprint
 from uptime import uptime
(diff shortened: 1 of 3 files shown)


vpetersson:
@rusko124 Wouldn't this break in the new Celery branch? How do you want to solve that?

rusko124:
> @rusko124 Wouldn't this break in the new Celery branch? How do you want to solve that?

In Screenly Version column is set current_branch@short_git_hash
One edit if the `master` branch then it will be set to `development` text(For more understanding).

I do not know what exactly to set in this column, this is just an example for now. Could you help me with this? @vpetersson

bcherup:
Will this used in experimental ?

rusko124:
@vpetersson  
What needs to be added or changed to progress here?

PR #1037, opened 2019-02-24
Add installer for libraspberrypi0 and ifupdown, also chmod vchiq

JeffResc:
These steps were necessary in order to get screenly to function correctly when installing it on Balena Cloud. I've decided to PR these changes in an attempt to share my fixes with everyone else that uses screenly.


vpetersson:
Looks good. Thanks @JeffResc.

PR #1039, opened 2019-03-01
Fixes: viewer freezes and restarts in some cases.

rusko124:
@vpetersson @ealmonte32
This [PR](https://github.com/Screenly/screenly-ose/pull/923)  for the problem when browser stuck at alert() prompt () etc. events
Unfortunately, this has given other problems, such as freezing and restarting the browser with non-standard encodings, since viewer read all the events of the browser.

These changes will revert PR above, but it will resolve issues for most users.

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -45,7 +45,6 @@ INTRO = '/screenly/intro-template.html'
 
 current_browser_url = None
 browser = None
-browser_focus_lost = False
 
 VIDEO_TIMEOUT = 20  # secs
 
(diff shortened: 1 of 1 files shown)


ealmonte32:
@rusko124 
sorry to even bring this up, but while testing for some other [issue](https://github.com/Screenly/screenly-ose/issues/1222) this came up again on the latest master branch, so i just wanted to bring it up..

running a `tail -f /var/log/syslog` yields no sort of the viewer checking if it actually loaded the next asset or not, and thus requires a manual `systemctl restart screenly-viewer.service` for it to start working again..

i know you said screenly will be at some point moved to qt web view so this might be a waste of time to look into, but i remember i [committed ](https://github.com/Screenly/screenly-ose/pull/1095/files)the following to the master branch on april of this year:
`@on_event   LOAD_ERROR    exit`
but i dont see it on the config-screenly file when i just checked.. i dont think you guys merged the change..

i have added it to my config-screenly file to further test, and reason for doing it is because since lots of users are still on the master branch with uzbl, and they might encounter this, maybe that exit command on load error event might help..

image

PR #1046, opened 2019-03-07
Allow PLAYER_NAME environment variable

JeffResc:
This pull request has two proposed changes for balenaCloud devices:
1. Allows the environment variable ```PLAYER_NAME``` to be set in the balenaCloud administration panel.
2. If ```PLAYER_NAME``` is not set in the balenaCloud administration panel, the player name will default to ```BALENA_DEVICE_NAME_AT_INIT```, or the device's name when it is first initialized on balenaCloud.

### DIFF
--- a/bin/start_resin.sh
+++ b/bin/start_resin.sh
@@ -26,6 +26,12 @@ if [ -n "${MANAGEMENT_USER+x}" ] && [ -n "${MANAGEMENT_PASSWORD+x}" ]; then
     sed -i -e "s/^user=.*/user=${MANAGEMENT_USER}/" -e "s/^password=.*/password=${MANAGEMENT_PASSWORD}/" /data/.screenly/screenly.conf
 fi
 
+if [ -n "${PLAYER_NAME+x}" ]; then
+    sed -i -e "s/^player_name=.*/player_name=${PLAYER_NAME}/" /data/.screenly/screenly.conf
+else
+    sed -i -e "s/^player_name=.*/player_name=${BALENA_DEVICE_NAME_AT_INIT}/" /data/.screenly/screenly.conf
+fi
+
 sed -i "/\[Service\]/ a\Environment=RESIN_UUID=${RESIN_DEVICE_UUID}" /etc/systemd/system/screenly-web.service
 
 systemctl start X.service


bin/start_resin.sh
```diff
@@ -26,6 +26,12 @@ if [ -n "${MANAGEMENT_USER+x}" ] && [ -n "${MANAGEMENT_PASSWORD+x}" ]; then
     sed -i -e "s/^user=.*/user=${MANAGEMENT_USER}/" -e "s/^password=.*/password=${MANAGEMENT_PASSWORD}/" /data/.screenly/screenly.conf
 fi
 
+if [ -n "${PLAYER_NAME+x}" ]; then
+    sed -i -e "s/^player_name=.*/player_name=${PLAYER_NAME}/" /data/.screenly/screenly.conf
```
vpetersson: Let's not hardcode `/data/[...]`. I have a feeling that will break multi-container environments. See https://www.balena.io/docs/learn/develop/runtime/#persistent-storage

JeffResc: @vpetersson This is how it existed already on line 26, should I update both of them?

JeffResc: @vpetersson This should be updated to ```/var/lib/docker/volumes/<APP ID>_resin-data/_data```?

vpetersson: > This is how it existed already on line 26, should I update both of them?

Yes I know - that file pre-dates the multi-container setup.

> This should be updated to /var/lib/docker/volumes/<APP ID>_resin-data/_data?

I'm not sure. I would imagine that there is some sort of environment variable that we can tap into that would give you the data path. If not, it might be worthwhile porting on the Balena forum.

JeffResc: Just read up on this [document](https://www.balena.io/docs/learn/develop/runtime/#balenaos-v2120-and-above), and it says that the `/data` directory is automatically linked to the data directory that the container is in. Without having multiple containers, those directories don't exist and only the `/data` directory is accessible, at least on the instance that I'm currently testing on (latest versions of all softwares).

vpetersson:
@JeffResc can you rebase this after the latest merge?

Issue #1054, opened 2019-03-21
Suggestion: Dark(er) splash screen?


madsrh:
I've always found the [pure white splash screen](https://camo.githubusercontent.com/5eed8b940f5dba9776024d7c73f992bfd00caa6c/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f3538313039362f3136323739322f64643534626533342d373763662d313165322d383037392d3938306138326439643730632e706e67) to be kind of eye hurting when booting up Screenly for the first time. This is an issue because Screenly is often used on very large screens. Why not use a "dark-theme" for this?

I haven't found the holy grail regarding the design yet, but here's a (dirty) mockup to illustrate what I mean: 

http://madsrosendahl.dk/screenly/

To be clear:
- Dark doesn't have to be black, just not eyehurting white #ffffff
- I'm not referring to the webinterface (for configuration), ONLY the content that is displayed initially on the digital sign itself.

![image](https://user-images.githubusercontent.com/3986894/54740475-d1108880-4bbb-11e9-84a1-f229616069cd.png)

rusko124:
I really like it!
@vpetersson

L4cko:
Hi @madsrh 

Looks ok. Remove the thin white line and make the logo smaller 20%

http://prntscr.com/n1dstw

vpetersson:
Thanks @madsrh. Take a look in the master branch (developer version) to see what you think.

PR #1060, opened 2019-03-22
Redesign: splash and hotspot pages

rusko124:
Ref #1054
hotspot
responsive
splash_page

PS: At the moment I don't have a real device to see how it looks. But later I'll ping you @vpetersson 

UPD:
Colors in Firefox
firefox


rusko124:
photo_2019-03-23_13-26-46
photo_2019-03-23_13-26-59

@vpetersson

Issue #1043, opened 2019-03-07
No database migration on Balena upgrade


axlroden:
#### Overview of the Issue

I did a pull and push to my balena environment, and there were new fields in the database. 
The migrate.py script is not run and is not pointing at /data/.screenly but /home/pi/...

#### Reproduction Steps

Steps to reproduce this issue, eg:

* git pull
* git push balena master

### Environment

* Raspberry Pi Hardware Version: pi3
* Screenly OSE Version: from 47a0adb to f16d953

I had to manually run the migration script with an edited line:
Change:
`configdir = os.path.join(os.getenv('HOME'), '.screenly/')`
To:
`configdir = os.path.join(os.getenv('/data/.screenly/')`

vpetersson:
@axlroden Hmm no that isn't correct. `os.path.join(os.getenv('/data/.screenly/')` hard codes the path. I know that the Balena guys made some changes to this when they introduced the multi-container system. However, we need to make it dynamic such that it works both with and without Balena. The above change would break anything other than Balena.

axlroden:
Ignore what I did to fix the issue, the issue as I see it is there is no migration run on a container update.

vpetersson:
@axlroden Yes that is a valid bug.

PR #1068, opened 2019-04-02
Update Dockerfile.template to use balenalib base images

chrisys:
The `latest` in `resin` was jessie, but in `balenalib` it's `stretch` so we need to specify `jessie` explicitly.

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,4 +1,4 @@
-FROM resin/%%RESIN_MACHINE_NAME%%-debian
+FROM balenalib/%%BALENA_MACHINE_NAME%%-debian:jessie
 MAINTAINER Anton Molodykh <amolodykh@screenly.io>
 
 RUN apt-get update && \


chrisys:
@ealmonte32 does this one work OK for you (I only noticed your PR after submitting this one)?

ealmonte32:
@chrisys , I have not tried it, I've been trying different docker templates in order to deal with the new balenalib image.. I can give it a try and let you know.

ealmonte32:
@chrisys 
Most of the container images i create to test keep getting the "failed to get d-bus connection" error, including using the repo:jessie you mentioned. I've only gotten the resin container image to work with new balenaOS on a pi zero, not pi3.

I am new to creating docker/moby container images, and I've read and sort of understand we need to rewrite the dockerfile/app that does not depend on systemd or dbus, but it is just going over my head right now.
Some people have said to use the --privileged=true flag, but I dont know where in the dockerfile that would go, I already tested different environment flags on the dockerfile to run screenly as not the root account, but to no avail.

Can you point me in the right direction?

rusko124:
@chrisys doesn't work for me

@vpetersson I think we need add update balena's template to current sprint.
Dockerfile.template is not working at the moment.

chrisys:
@ealmonte32 @rusko124 thanks for trying it. Having looked at it more closely now it's likely that the issues are being caused by the changes to the initsystem in the latest base images.

I guess the current`Dockerfile.template` was produced before balenaCloud supported multicontainer; I wonder now (without knowing the inner workings of Screenly) if it would be possible to run all of the required services in their own containers?

Docs regarding initsystem changes: https://www.balena.io/docs/reference/base-images/base-images/#installing-your-own-initsystem

vpetersson:
@chrisys Good catch. Yes, when Balena moved to multi-containers, a lot of things changed. #1043 touches this too. We need to schedule a task in the next sprint to sort this out.

chrisys:
@vpetersson OK, I'll close this in favour of issue #1080

chrisys:
One last thing, I made a fork which will deploy on balenaCloud today, until the multi-container implementation is ready: https://github.com/chrisys/screenly-ose

ealmonte32:
Just a reminder cause I saw this while reading this: https://docs.docker.com/engine/reference/builder/#cmd

The current Dockerfile.template has CMD twice, it cannot be used more than once or else the build process only uses the last CMD during run time.. 
I dont know if that only applies to "Dockerfile" files, and not "Dockerfile.template" files since I am still a bit confused about the difference but know that the template one can be given certain runtime arguments while the other one cant.

PR #1071, opened 2019-04-05
Update utils.py

ealmonte32:
Change in HTTP xrange since redirect responses are also OK.
Set request verification to true to stop urllib InsecureRequestWarning showing up in logs and users mistakenly thinking this is what causes the TLS handshake error..

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -23,7 +23,8 @@ from assets_helper import update
 
 arch = machine()
 
-HTTP_OK = xrange(200, 299)
+# 300 level HTTP responses are also ok, such as redirects, which many sites have and load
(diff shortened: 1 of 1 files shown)


lib/utils.py
```diff
@@ -23,7 +23,8 @@
 
 arch = machine()
 
-HTTP_OK = xrange(200, 299)
+# 300 level HTTP responses are also ok, such as redirects, which many sites have and load
+HTTP_OK = xrange(200, 399)
```
vpetersson: Should we throw [418](http://httpstat.us/418) in there too? ;)

ealmonte32: lol, no tea for me, only coffee

PR #1077, opened 2019-04-11
Update install.sh to include OSE version file

ealmonte32:
Added the echo to user selection options to create simple file that users and System info can use to show current version of screenly.


bin/install.sh
```diff
@@ -33,14 +33,14 @@ echo && read -p "Would you like to use the experimental branch? It contains the
 if [ "$EXP" != 'y'  ]; then
   echo && read -p "Would you like to use the development branch? You will get the latest features, but things may break. (y/N)" -n 1 -r -s DEV && echo
   if [ "$DEV" != 'y'  ]; then
-    export DOCKER_TAG="production"
+    export DOCKER_TAG="production" && echo "Screenly OSE version: Production" > ~/OSE_version.md
```
vpetersson: No need for `&&`, just add it to the line below. That makes it more readable.

PR #1084, opened 2019-04-19
Fix: backup recover for celery branch

rusko124:
Backup and recover was rework there #987
This fix contains the missing part of the code.

### DIFF
--- a/lib/backup_helper.py
+++ b/lib/backup_helper.py
@@ -32,4 +32,6 @@ def recover(file_path):
             if directory not in tar.getnames():
                 raise Exception("Archive is wrong.")
 
+        tar.extractall(path=getenv('HOME', '/home/pi'))
+
     remove(file_path)


lib/backup_helper.py
```diff
@@ -32,4 +32,6 @@ def recover(file_path):
             if directory not in tar.getnames():
                 raise Exception("Archive is wrong.")
 
+        tar.extractall(path=getenv('HOME', '/home/pi'))
```
vpetersson: @rusko124 this would break when running on Balena, wouldn't it?

rusko124: @vpetersson  In Balena cloud we use the environment variable `HOME=/data`, so this should work.

vpetersson: Great - we should probably replace that with SRLY_STORAGE or something but fine for now.

PR #1089, opened 2019-04-26
Fix: wifi-connect and wifi reset

rusko124:
We have issues about that people can not reset the current point of wifi or when user change the location of the device it does not show a point(balena) to connect to a new access wireless point.
This PR solve these problems.

General tests were reproduced:
- Connected to the point, after IP was displayed. Then, the current connection was reset through `http://ip/settings`. After reboot we can see the hotspot page.
- Successfully connected to the wireless point. Turn off the Wi-Fi router. After reboot we can see the hotspot page.

### DIFF
--- a/ansible/roles/network/files/wifi-connect.service
+++ b/ansible/roles/network/files/wifi-connect.service
@@ -2,13 +2,13 @@
 Description=Wifi Connect
 Wants=network-online.target
 After=network-online.target
-ConditionPathExists=!/home/pi/.screenly/initialized
 
 [Service]
 WorkingDirectory=/home/pi/screenly
 User=pi
 Type=oneshot
 
+ExecStartPre=/bin/rm -f /home/pi/.screenly/initialized
 ExecStart=/usr/bin/python /home/pi/screenly/start_resin_wifi.py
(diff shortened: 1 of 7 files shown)


ansible/roles/network/files/wifi-connect.service
```diff
@@ -2,13 +2,13 @@
...
 Wants=network-online.target
 After=network-online.target
-ConditionPathExists=!/home/pi/.screenly/initialized
 
 [Service]
 WorkingDirectory=/home/pi/screenly
 User=pi
 Type=oneshot
 
+ExecStartPre=/bin/rm -f /home/pi/.screenly/initialized
```
vpetersson: Wouldn't this cause the wifi to be resetted at *every* run?

rusko124: > Wouldn't this cause the wifi to be resetted at _every_ run?

No, because the script checks if there is an active point or network connection. After script execute `ExecStartPost=/usr/bin/touch /home/pi/.screenly/initialized`.
It was checked manually, after restart everything works fine.

This part is necessary that the viewer not to skip the `hotspot` page.

PR #1090, opened 2019-04-26
Fix: snyk vulnerabilities

rusko124:
This PR solves `snyk` vulnerabilities.
Tested through custom run_upgrade script.


vpetersson:
Good fix! Thanks, @rusko124.

PR #1095, opened 2019-04-29
Update uzbl-config

ealmonte32:
adding the load_error event as a way to see if it is a good workaround for the TLS error on UZBL and allows assets to continue to load even if the browser encounters error.

### DIFF
--- a/ansible/roles/screenly/files/uzbl-config
+++ b/ansible/roles/screenly/files/uzbl-config
@@ -1,5 +1,6 @@
 set show_status = 0
-set on_event    = request ON_EVENT
+set on_event = request ON_EVENT
 
-@on_event   LOAD_START         js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
-@on_event   LOAD_COMMIT        js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
+@on_event   LOAD_START    js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
+@on_event   LOAD_COMMIT   js alert=function(e){return true};confirm=function(e){return true};prompt=function(e){return true};
+@on_event   LOAD_ERROR    exit


rusko124:
Hi @ealmonte32 !
Is the viewer still gets stuck at page with error? I am sure that Screenly is working well for now.
This was due to the fact that we did not handle `LOAD_ERROR` event inside the code, but now we have refused to handle any events in general. #1039
So the viewer should work without problems now.

vpetersson:
@ealmonte32 Are you able to reproduce the issue w/out this patch with the latest disk image (published this weekend)?

ealmonte32:
> Hi @ealmonte32 !
> Is the viewer still gets stuck at page with error? I am sure that Screenly is working well for now.
> This was due to the fact that we did not handle `LOAD_ERROR` event inside the code, but now we have refused to handle any events in general. #1039
> So the viewer should work without problems now.

@rusko124 
The viewer would get stuck as others reported and confirmed, and once the browser gets the TLS error, no other assets would be able to continue to load, this line has fixed the issue for me where other assets are able to continue to load even if the browser experiences the error.
Your fix at #1039 was focused on viewer.py, and this is directly telling the browser what to do rather than how the viewer finds the browser and then trying to restart it if it finds it dead etc..
I've been using the production version and dev version that already included #1039 and the error was still happening with the specific websites that people reported.

ealmonte32:
> @ealmonte32 Are you able to reproduce the issue w/out this patch with the latest disk image (published this weekend)?

@vpetersson 
I have not touched the new image as I got back from vacation yesterday.
But remember that all the issues dealing with TLS handshake error that were sent to Sprint 9 were being tested and tried during the time before the image was created, and the tickets/issues were closed without an actual fix for the specific TLS handshake error on the sites users reported.
And I had tried various methods of testing like using the Sprint 8 image, or installing raspbian lite then install script, upgrading to developer method or running both developer and production on two different Pi's, and the error would occur on all versions.
This commit is simple and should not affect anything else.
I just need users that were having the TLS error to test and see if this helps, but without including it in the config-screenly file is hard to show users how to add that line, so it is much easier to include it in the installation and thats it.

rusko124:
This is strange because I can suggest steps to reproduce:

- Add url `https://dasdaxzczsdfafadaxc` (with skip asset check)
error
- Uzbl events when this url is loaded(part of):
```
EVENT [1374] VARIABLE_SET progress.output str '[=>        ]10%'
EVENT [1374] GEOMETRY_CHANGED '800x600+135+105'
EVENT [1374] GEOMETRY_CHANGED '800x600+135+105'
EVENT [1374] GEOMETRY_CHANGED '800x600+135+105'
EVENT [1374] LOAD_ERROR 'https://dasdaxzczsdfafadaxc/' 2 'Error resolving \'dasdaxzczsdfafadaxc\': Name or service not known'
EVENT [1374] REQUEST_STARTING 'about:blank'
EVENT [1374] FOCUS_GAINED
EVENT [1374] REQUEST_STARTING 'https://dasdaxzczsdfafadaxc/favicon.ico'
EVENT [1374] LOAD_PROGRESS 10
```
error_page
- Then I added this asset to my active playlist. I can confirm that the viewer does not stuck. We just see a page with an error during the specified duration.

ealmonte32:
> * Then I added this asset to my active playlist. I can confirm that the viewer does not stuck. We just see a page with an error during the specified duration.

Hmm.. yes I see how what you are doing and saying makes sense, you are making the browser force a load error event to see if it stays stuck, but I can't explain why the TLS error one does get browser stuck.

I am testing now again by removing the load_error on event to see if I am able to reproduce "Unable to load page - TLS handshake error".
I will post back here as soon as it happens or with any updates regarding this.

rusko124:
FYI, I also had the error "Unable to load page - TLS handshake error", but with the [latest](https://github.com/Screenly/screenly-ose/releases/tag/0.17) updates the viewer was no hang. This issue was in the [previous](https://github.com/Screenly/screenly-ose/releases/tag/0.16) release in my case.

vpetersson:
@rusko124 should we close or merge this?

ealmonte32:
@rusko124 
Ok, I will try the latest release without changing anything to see how it handles the assets that have given everyone the same TLS error and would stay stuck on that screen.

rusko124:
@vpetersson For now, Yes, I think we can close.
@ealmonte32 If you'll find a error with viewer stuck and steps to reproduce, please ping me.

ealmonte32:
> FYI, I also had the error "Unable to load page - TLS handshake error", but with the [latest](https://github.com/Screenly/screenly-ose/releases/tag/0.17) updates the viewer was no hang. This issue was in the [previous](https://github.com/Screenly/screenly-ose/releases/tag/0.16) release in my case.

@rusko124 
When you say "the latest updates", do you mean I need to install the Sprint 9 image, or is running the `run_upgrade.sh` supposed to work as well? because I just ran the upgrade script, reinstalled screenly, chose development version, I still got the TLS error, and it seems to be still in old system because the System Info page of web dashboard does not show the new info like mac address, screenly version, raspberry model, and uzbl is still the browser..

Issue #1100, opened 2019-05-02
Viewer crashes when NetworkManager is not in use


joeladria:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

Just upgraded to the latest production release, and I am not currently a user of NetworManager, however the main loop in viewer.py calls nm-cli even when it was not activated in the install. 

Erasing all code in main() regarding network manager seems to resolve this issue for me. 

#### Reproduction Steps

Steps to reproduce this issue, eg:

* Install OSE using the install script, and opt to NOT use NetworkManager
* viewer does not play content

### Environment

* Raspberry Pi Hardware Version: Model 3B+ Revision: 1.3 Ram: 1 GB Sony UK

* Screenly OSE Version: production@618864d

ealmonte32:
@joeladria 
See my response for fix here:
https://github.com/Screenly/screenly-ose/issues/1102#issuecomment-489695705

rusko124:
Should be resolved in the last commits.
I think we can close this

vpetersson:
@joeladria Please test the development branch and let us know if the issue still persists.

Issue #1094, opened 2019-04-27
Fix dependency issue


vpetersson:
When creating the latest disk image, I ran across the following error:

```
ERROR: requests 2.20.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.1 which is incompatible.
ERROR: cryptography 2.6.1 has requirement cffi!=1.11.3,>=1.8, but you'll have cffi 1.7.0 which is incompatible.
```

We should address this going forward.

gstoelen:
During installation of the latest build I also ran into this error : 
/home/pi/.local/lib/python2.7/site-packages/requests/__init__.py:91: RequestsDependencyWarning: urllib3 (1.25.1) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)

It's refusing to start the Screenly services. 

This comes from the logs : 

May 02 13:45:21 ZNCTPI001 systemd[1]: screenly-viewer.service: Service hold-off time over, scheduling restart.
May 02 13:45:21 ZNCTPI001 systemd[1]: Stopped Screenly Viewer.
May 02 13:45:21 ZNCTPI001 systemd[1]: Starting Screenly Viewer...
May 02 13:45:21 ZNCTPI001 systemd[1]: Started Screenly Viewer.
May 02 13:45:24 ZNCTPI001 python[2248]: /usr/local/lib/python2.7/dist-packages/requests/__init__.py:91: RequestsDependencyWarning: urllib3 (1.25.1) or chardet (3.0.4) doesn't match a supported version!
May 02 13:45:24 ZNCTPI001 python[2248]:   RequestsDependencyWarning)
May 02 13:45:26 ZNCTPI001 python[2248]: Viewer crashed.
May 02 13:45:26 ZNCTPI001 python[2248]: Traceback (most recent call last):
May 02 13:45:26 ZNCTPI001 python[2248]:   File "/home/pi/screenly/viewer.py", line 516, in <module>
May 02 13:45:26 ZNCTPI001 python[2248]:     main()
May 02 13:45:26 ZNCTPI001 python[2248]:   File "/home/pi/screenly/viewer.py", line 481, in main
May 02 13:45:26 ZNCTPI001 python[2248]:     wireless_connections = nmcli_get_connections('wlan*', 'ScreenlyOSE-*', active=True)
[...]

PR #1101, opened 2019-05-03
Fix: work with nmcli

rusko124:
These edits helps for users w/out network-manager
Refs #1100, https://github.com/Screenly/screenly-ose/issues/1094#issuecomment-488649624

But as I said here https://github.com/Screenly/screenly-ose/issues/1094#issuecomment-488661111
https://github.com/balena-io/wifi-connect uses NetworkManager. In other words, the user will not be able to configure the wireless network.

Sum up: Yes , this PR will help users without NetworkManager, but I think we need to make NetworkManager by default.
@vpetersson Need your opinion.

UPD: Our latest image already includes NetworkManager.

### DIFF
--- a/ansible/roles/network/tasks/main.yml
+++ b/ansible/roles/network/tasks/main.yml
@@ -1,3 +1,8 @@
+- name: Installs NetworkManager
+  apt:
+    name: network-manager
+    state: latest
+
 - name: Check if screenly-network-manager files exist
   stat:
     path: /usr/sbin/screenly_net_mgr.py
@@ -36,20 +41,15 @@
 
 # Use resin-wifi-connect if Stretch
 
-- debug:
-    msg: "Manage network: {{ manage_network }}"
-
 - name: Disable dhcpcd
   command: systemctl disable dhcpcd
   when:
     - ansible_distribution_major_version|int >= 9
-    - manage_network|bool == true
 
 - name: Activate NetworkManager
   command: systemctl enable NetworkManager
   when:
     - ansible_distribution_major_version|int >= 9
-    - manage_network|bool == true
 
 - name: Check if resin-wifi-connect required version exist
   stat:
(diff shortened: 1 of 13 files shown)


lib/utils.py
```diff
@@ -120,7 +120,11 @@ def nmcli_get_connections(pattern=None, pattern_ignore=None, fields=None, active
     if active:
         params.append('--active')
 
-    out = sh.nmcli('-t', '-f', ','.join(fields), 'c', 'show', *params)
+    try:
+        out = sh.nmcli('-t', '-f', ','.join(fields), 'c', 'show', *params)
```
vpetersson: Should we not use dbus for this instead?

ansible/roles/network/tasks/main.yml
```diff
@@ -120,60 +115,33 @@
...
+    dest: /var/lib/polkit-1/localauthority/10-vendor.d/org.freedesktop.NetworkManager.pkla
+
+- name: Set ResultAny to yes
+  replace:
+    regexp: '^ResultAny=.*'
+    replace: 'ResultAny=yes'
+    dest: /var/lib/polkit-1/localauthority/10-vendor.d/org.freedesktop.NetworkManager.pkla
+
+- name: Copy org.freedesktop.NetworkManager.pkla to 50-local.d
+  command: cp -f /var/lib/polkit-1/localauthority/10-vendor.d/org.freedesktop.NetworkManager.pkla /etc/polkit-1/localauthority/50-local.d
```
vpetersson: Shoulnd't we use the native copy feature for this?

rusko124: It makes sense, but in this case it does not work unfortunately.

vpetersson: Got it. All good then. Let's fix the CI run before we can land it tho.

PR #1133, opened 2019-05-24
Fix: check_update

rusko124:
Edits: 
- Code reuse
- Fixes error in balena

### DIFF
--- a/server.py
+++ b/server.py
@@ -86,8 +86,7 @@ def is_up_to_date():
         latest_sha = None
 
     if latest_sha:
-        branch_sha = diagnostics.get_git_hash()
-        return branch_sha.stdout.strip() == latest_sha
+        return diagnostics.get_git_hash() == latest_sha
 
     # If we weren't able to verify with remote side,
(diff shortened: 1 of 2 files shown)


vpetersson:
@rusko124 Take a closer look at this if you're touching the git stuff. There are more places where we can improve this (i think i counted three different places when I looked).

Moreover, we still need this for Balena, so we need a fallback logic (perhaps writing the hash to disk during the Docker build process?) such that we can view the version there too.

rusko124:
>Take a closer look at this if you're touching the git stuff. There are more places where we can improve this (i think i counted three different places when I looked).

Fixed all I found.

>Moreover, we still need this for Balena, so we need a fallback logic (perhaps writing the hash to disk during the Docker build process?) such that we can view the version there too.

It's a little heavier than it looks.
>NOTE: You don't need to worry about ignoring .git as the builders already do this by default. 

The note from https://www.balena.io/docs/learn/develop/dockerfile/

Issue #1124, opened 2019-05-16
Refactor templates to use includes


vpetersson:
Right now there is a lot of repeating in the templates. We should adopt a DRY approach and remove a lot of the repeated code.

rusko124:
Can we close this?

Issue #1080, opened 2019-04-12
Migrate to multi-container on Balena


vpetersson:
Some time ago, Balena introduced [multi-container](https://www.balena.io/docs/learn/develop/multicontainer/) support. This is a great improvement as it follows the Docker best practice of only running one service per container. We likely need three containers:

 * Viewer
 * Server
 * Celery

vpetersson:
@rusko124 I guess this is still blocked by the lack of celery container, right?

rusko124:
@vpetersson Yes, you're right

PR #1139, opened 2019-05-30
Edits for celery branch

rusko124:
Edits:
- Merge with `master` branch
- Reworked docker-compose for dev (`./docker/start.sh up`)
- Added celery container for balenaCloud #1080
- Split all dockerfiles to multiple dockerfiles(all in `./docker`) #1124 

Tests:
- Backup and upload in balena version: worked 
balena_finished
celery


bin/install.sh
```diff
@@ -57,27 +57,20 @@ EOF
     echo && read -p "Would you like to use the development branch? You will get the latest features, but things may break. (y/N)" -n 1 -r -s DEV && echo
     if [ "$DEV" != 'y'  ]; then
       export DOCKER_TAG="production"
+      echo "Screenly OSE version: Production" > ~/OSE_version.md
```
vpetersson: Production isn't a version. it's a channel. The version is the hash.

docker-compose.yml
```diff
@@ -1,39 +1,65 @@
-version: "3"
+version: "2"
```
vpetersson: Why are we downgrading?

rusko124: Balena is work with this version.
`docker-compose.dev.yml` uses `version: 3`

vpetersson: Got it. Fair.

bin/install.sh
rusko124: Got it, then we need to edit this point in the master too.

vpetersson: Yep. If that's from master then we need to fix it there too.

PR #1143, opened 2019-06-04
Allow connection to websocket when server is behind reverse proxy.

MaicoTimmerman:
Currently the view index will generate an URL for the websocket, which is based on the physical address of the machine the server is running on. However when the server is running behind a reverse proxy like apache or nginx, we need to correct hostname in the request.

This pull request changes the generation of the websocket URL to use the hostname of the http request, which will be the ip if running natively without proxy and the hostname of the reverse proxy when being served through apache or nginx.

### DIFF
--- a/server.py
+++ b/server.py
@@ -19,6 +19,7 @@ from hurry.filesize import size
 from mimetypes import guess_type
 from os import getenv, makedirs, mkdir, path, remove, rename, statvfs, stat
 from subprocess import check_output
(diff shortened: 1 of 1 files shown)


vpetersson:
Thanks for the contribution @MaicoTimmerman. Looks sensible at a glance. @rusko124 could you please take a look?

rusko124:
@vpetersson look good to me

PR #1144, opened 2019-06-05
Edit: version of screenly in install.sh

rusko124:
Ref https://github.com/Screenly/screenly-ose/pull/1139#discussion_r288925166
```
pi@raspberrypi:~ $ cat version.md 
Screenly version: master@76c881cc12a7ed93cf74547abec7961fe3a4bc18
PRETTY_NAME="Raspbian GNU/Linux 9 (stretch)"
```

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -102,7 +102,7 @@ sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*' 
 
 cd ~/screenly && git rev-parse HEAD > ~/.screenly/latest_screenly_sha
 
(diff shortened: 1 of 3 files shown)


bin/install.sh
```diff
@@ -108,6 +102,8 @@ sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*'
 
 cd ~/screenly && git rev-parse HEAD > ~/.screenly/latest_screenly_sha
 
+echo -e "Screenly version: $(git rev-parse --abbrev-ref HEAD)@$(git rev-parse HEAD)\n$(cat /etc/os-release | grep 'PRETTY_NAME')" > ~/version.md
```
vpetersson: The latter part will likely fail in a dockerized version. Also, `lsb_release` is probably a lot cleaner than `cat`-ing the file.

vpetersson:
@rusko124 Still need to switch to the short git hash.

PR #1145, opened 2019-06-06
Fix: database migration on Balena

rusko124:
Issue #1043


rusko124:
`run_setup` function running only in server container, other containers running after server container

vpetersson:
Ah ok. Good then. THanks.

Issue #1149, opened 2019-06-10
Add new backend support for WoTT


vpetersson:
Once #1148 has been implemented, let's build to build the [WoTT Agent](https://github.com/WoTTsecurity/agent) integration as the first reference implementation. Think of the WoTT agent as Last Pass for IoT, meaning a credential manager.

The integration would work as follows (documentation needed too):

* The user would install the WoTT agent onto the device and claim the device in the WoTT dashboard.
* The user would create two different credentials in the WoTT dashboard, one for the username, and one for the password.  These must have predetermined names, as we will rely on these paths later.
* We need to add a new auth backend that supports WoTT. All it essentially need to do is to read the credentials from the paths generated above. This module then needs to check the credentials entered by the user against these credentials.

PR #1154, opened 2019-06-11
Edits: hide integrations button if is not balena application

rusko124:
Edits:
- Hide `integrations` button in header if Screenly running not on balena


vpetersson:
The idea was to add more integrations (such as #1149).

Issue #1142, opened 2019-06-04
localisation of date display


pooh22:
Please reopen #599 (Localise date format)
It would be really nice if the date could be displayed in a more intuitive form for the people using it. In the Netherlands, we're really used to day-month-year format, not the US format month/day/year.

rusko124:
Hi @pooh22 !
Feature was added. Try to upgrade and let me know your opinion.

pooh22:
Thanks @rusko124 It will be a while before I have time to test it, but it is much appreciated!

PR #1157, opened 2019-06-12
Feature: date format setting

rusko124:
Ref #1142

date_forma

### DIFF
--- a/server.py
+++ b/server.py
@@ -102,14 +102,15 @@ def template(template_name, **context):
     but also injects some global context."""
 
     # Add global contexts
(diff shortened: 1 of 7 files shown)


vpetersson:
@rusko124 looks like there are still some code quality issues.

rusko124:
This is because script does not see the variables in the index template.
изображение

vpetersson:
> This is because script does not see the variables in the index template.
> изображение

Fair enough.

vpetersson:
Nice job. Merging.

Issue #1148, opened 2019-06-10
Refactor auth backend to make it dynamic


vpetersson:
There's currently limited support for [auth](https://github.com/Screenly/screenly-ose/blob/master/docs/http-basic-authentication.md) in Screenly OSE. It would be nice to make this more modular such that we can add more backends in the future. 

As such, let's refactor the way that auth is done right now and do the following:

 * Introduce a new settings in the dashboard where a user can pick the auth backend.
 * Refactor the auth code (such as [this](https://github.com/Screenly/screenly-ose/blob/master/settings.py#L45-L46) to make it modular and support multiple backends.
 * Refactor the auth checks to pick the auth backend from a given system (currently only the bult-in, but design it to support multiple ones).

PR #1165, opened 2019-06-17
add new backend support for wott

### LINKED ISSUE #1149: Add new backend support for WoTT

eugenosm:
closes #1149 
based on #1148 (PR #1162)


eugenosm:
and it need to do something with permission to acces to credentials file

vpetersson:
@a-martynovich Could you review the last part of this?

auth.py
```diff
@@ -0,0 +1,234 @@
...
+            raise ValueError("Must supply current password to change username")
+        if not current_pass_correct:
+            raise ValueError("Incorrect current password.")
+
+        if not self._fetch_credentials():
+            raise ValueError("Can not read WoTT credential file or login credentials record is incorrect")
+
+        # self.settings['user'] = self.user
+        # self.settings['password'] = self.password
+        # self.settings['openpass'] = self.openpass
```
a-martynovich: There're no settings, so `update_settings` shouldn't do anything.

auth.py
```diff
@@ -0,0 +1,234 @@
...
+        return 'auth_basic.html', {'user': self.settings['user']}
+
+    def authenticate(self):
+        realm = "Screenly OSE {}".format(self.settings['player_name'])
+        return Response("Access denied", 401, {"WWW-Authenticate": 'Basic realm="{}"'.format(realm)})
+
+    def update_settings(self, current_password):
+        auth_backend = self.settings['auth_backend']
+        current_pass_correct = True if auth_backend == '' \
+            else self.settings.auth_backends[auth_backend].check_password(current_password)
```
a-martynovich: I should remove this code. Auth class should only care if its password is correct, it shouldn't care about other backends.

auth.py
```diff
@@ -0,0 +1,236 @@
...
```
a-martynovich: @eugenosm Please rebase your branch on top of #1162 . In my PR this line [looks different](https://github.com/Screenly/screenly-ose/pull/1162/files#diff-01a06b801eb7962fc5be15ff8f1cd899R97).

a-martynovich:
@vpetersson what should happen if we can't load WoTT credentials? Hide the backend from the dropdown? Act as if auth is disabled? Show error when switching to WoTT backend (and disallow switching)?

vpetersson:
> @vpetersson what should happen if we can't load WoTT credentials? Hide the backend from the dropdown? Act as if auth is disabled? Show error when switching to WoTT backend (and disallow switching)?

That's an interesting use case.I don't think we should disable auth, as that's a potential attack vector. I'd suggest throwing a 50x maybe?

a-martynovich:
I meant what if we can’t load wott credentials when we have other backend selected? Should we allow switching to wott backend in Settings (but show error after saving settings)? Or should we hide the option?

vpetersson:
> I meant what if we can’t load wott credentials when we have other backend selected? Should we allow switching to wott backend in Settings (but show error after saving settings)? Or should we hide the option?

Oh, i was purely talking about *if* WoTT was selected. If WoTT isn't selected, then it doesn't really matter. In that case we should probably just log it.

a-martynovich:
I’m asking about th UI. Should we allow switching to wott backend in Settings (but show error after saving settings)? Or should we hide the option?

vpetersson:
After a discussion on Slack.

 * If the file (normally ` /opt/wott/credentials/screenly.json` unless overridden in config) is absent, prevent save and display “Can’t load /opt/wott/credentials/screenly.json”.
 * If the folder `/opt/wott` is absent, then assume the WoTT agent is not installed.

These are somewhat naive assumptions, but will do the trick for now.

auth.py
```diff
@@ -0,0 +1,228 @@
...
+from functools import wraps
+import hashlib
+import os.path
+import json
+import re
+
+from flask import request, Response
+
+WOTT_CREDENTIALS_PATH = '/opt/wott/credentials'
+WOTT_SCREENLY_CREDENTIAL_NAME = 'screenly_credentials'
```
korvyashka: If it is used only to be appended with ".json" - I guess it could be named WOTT_SCREENLY_CREDENTIAL_FILE

auth.py
```diff
@@ -0,0 +1,228 @@
...
+        if os.path.isfile(wott_credentials_path):
+            with open(wott_credentials_path, "r") as credentials_file:
+                credentials = json.load(credentials_file)
+                login_record = credentials.get('login_credentials', '')
+                if not login_record:
+                    return False
+                login_record = login_record.split(':', 1)
+                if len(login_record) == 2:
+                    self.user, password = login_record
+                    self.password = '' if password == '' else hashlib.sha256(password).hexdigest()
```
korvyashka: Seems password is stored in raw form in file - I'm not sure if it is fine, but I guess that is wott cred format.

auth.py
```diff
@@ -0,0 +1,228 @@
...
+        return True
+
+    def authenticate(self):
+        pass
+
+    def check_password(self, password):
+        return True
+
+
+class BasicAuth(Auth):
```
korvyashka: This seems to be the same code as in https://github.com/Screenly/screenly-ose/pull/1162 and I guess will be removed after requested by you rebase.

auth.py
```diff
@@ -0,0 +1,228 @@
...
+                if not login_record:
+                    return False
+                login_record = login_record.split(':', 1)
+                if len(login_record) == 2:
+                    self.user, password = login_record
+                    self.password = '' if password == '' else hashlib.sha256(password).hexdigest()
+                    return True
+
+        self.user = ''
+        self.password = ''
```
korvyashka: Few minor comments here.
1) Can't it be done stateless, without need of storing user and password in the object? Just by reading them from cred during the check? Probably cached with cache decorators f.e?
2) I believe this function can be simplified by common practices:
 - Return early
 - Reverse if and return early
 - Moving these 2 statements to top of the func.

a-martynovich: 1. It can, but why? We already have username/password in settings (basic auth) loaded into memory from screenly.conf

auth.py
a-martynovich: Exactly so.

vpetersson: It's actually the way that the basic auth is done in Screenly OSE too. It's not good, but that's a problem for another day.

auth.py
a-martynovich: No, the logic implies that this name will be used in WoTT Credentials UI in the "Name" field.

auth.py
a-martynovich: 2. done

a-martynovich:
Follow-up changes:
* WoTT option will be hidden if there's no /opt/wott directory.
* Server will respond with 503 if credentials file suddenly disappears while WoTT backend is active.

If credential file is missing when switching to WoTT backend in settings an error will be shown (it was already done by @eugenosm ).

PR #1166, opened 2019-06-20
add integration tab extension for wott

### LINKED ISSUE #1155: Add "Integration tab" extension for WoTT
We need to add a new section to the "Integration tab" if WoTT is detected and installed and populate it with the "Device ID" from WoTT.

eugenosm:
closes #1155

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -229,9 +229,11 @@ def authorized(orig):
     :return: Response
     """
     from settings import settings
+
     @wraps(orig)
     def decorated(*args, **kwargs):
         if not settings.auth:
             return orig(*args, **kwargs)
         return settings.auth.authenticate_if_needed() or orig(*args, **kwargs)
+
     return decorated
--- a/auth.py
+++ b/auth.py
@@ -220,9 +220,15 @@ class WoTTAuth(BasicAuth):
     @property
     def template(self):
         return None
-def authorized(orig):
-    from settings import settings
 
+
+def authorized(orig):
+    """
+    Annotation which initiates authentication if the request is unauthorized.
+    :param orig: Flask function
+    :return: Response
+    """
+    from settings import settings
     @wraps(orig)
     def decorated(*args, **kwargs):
         if not settings.auth:
--- a/auth.py
+++ b/auth.py
@@ -220,8 +220,6 @@ class WoTTAuth(BasicAuth):
     @property
     def template(self):
         return None
-
-
 def authorized(orig):
     from settings import settings
 
--- a/auth.py
+++ b/auth.py
@@ -14,6 +14,7 @@ WOTT_USER_CREDENTIALS_PATH = os.path.join(WOTT_CREDENTIALS_PATH, LINUX_USER)
 WOTT_SCREENLY_CREDENTIAL_NAME = 'screenly'
 
 
+
 class Auth(object):
     __metaclass__ = ABCMeta
 
--- a/auth.py
+++ b/auth.py
@@ -14,7 +14,6 @@ WOTT_USER_CREDENTIALS_PATH = os.path.join(WOTT_CREDENTIALS_PATH, LINUX_USER)
 WOTT_SCREENLY_CREDENTIAL_NAME = 'screenly'
 
 
-
 class Auth(object):
     __metaclass__ = ABCMeta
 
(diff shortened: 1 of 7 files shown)


lib/utils.py
```diff
@@ -329,3 +331,30 @@ def is_balena_app():
     :return: bool
     """
     return bool(getenv('RESIN', False)) or bool(getenv('BALENA', False))
+
+
+def is_wott_integrated():
```
vpetersson: @eugenosm this won't work well across runtime environments (such as Docker). It's also very hackish. Just do if `WOTT_PATH` exists, then assume the agent is installed.

lib/utils.py
```diff
@@ -329,3 +331,30 @@ def is_balena_app():
...
+    """
+    :return: WoTT Device id of this device
+    """
+    metadata_path = os.path.join(WOTT_PATH, 'metadata.json')
+    if os.path.isfile(metadata_path):
+        with open(metadata_path) as metadata_file:
+            metadata = json.load(metadata_file)
+        if 'device_id' in metadata:
+            return metadata['device_id']
+    return '-- can not read device id --'
```
vpetersson: This should be logged, not just returned. Also, should be "Could not read WoTT Device ID".

templates/settings.html
```diff
@@ -174,48 +174,31 @@ <h1 class="page-header">
...
                         </div>
                     </div>
-                    <div class="form-group row" id="user_group">
-                        <label class="col-2 col-form-label">User</label>
+                    {% if context.need_current_password %}
+                    <div class="form-group row" id="curpassword_group">
+                        <label class="col-2 col-form-label">Current Password</label>
                         <div class="col-3">
-                            <input class="form-control" name="user" type="text" value="{{ context.user }}">
+                            <input class="form-control" name="curpassword" type="password" value="">
```
vpetersson: "curpassword" -> "current-password"

vpetersson:
@a-martynovich could you please review this one?

lib/utils.py
eugenosm: ok

lib/utils.py
eugenosm: fixed

templates/settings.html
eugenosm: applied

auth.py
```diff
@@ -14,6 +14,7 @@
 WOTT_SCREENLY_CREDENTIAL_NAME = 'screenly'
 
 
+
```
a-martynovich: this causes our build to fail with PEP8 error

templates/settings.html
```diff
@@ -187,7 +187,7 @@ <h1 class="page-header">
                     <div class="form-group row" id="curpassword_group">
                         <label class="col-2 col-form-label">Current Password</label>
                         <div class="col-3">
-                            <input class="form-control" name="curpassword" type="password" value="">
+                            <input class="form-control" name="current-password" type="password" value="">
```
a-martynovich: this should also be changed in `settings.py:settings_page()`

auth.py
eugenosm: fixed

templates/settings.html
eugenosm: fixed

auth.py
```diff
@@ -220,14 +220,7 @@ def _check(self, username, password):
     @property
     def template(self):
         return None
-
-
 def authorized(orig):
-    """
```
a-martynovich: why was this removed?

docs/wott-authentication.md
```diff
@@ -6,15 +6,15 @@ Access to Screenly web configuration interface can be restricted via an [*HTTP b
 ## Configuration
 
 After installing [*wott-agent*](https://www.wott.io/) and claiming the device in [*wott-dashboard*](https://dash.wott.io).
-Create credential with `name` `screenly_credentials` and `key` `login_credentials` contains colon separated username and password ( for example `pi:raspberry` ).
-In `owner` field enter linux user under which Screenly-ose ran on your device ( f.ex. `pi`) 
+Create credential with `name` `screenly` and `key` `login_credentials` contains colon separated username and password ( for example `pi:raspberry` ).
```
a-martynovich: key `login`

eugenosm: yeah....i miss that

auth.py
eugenosm: restored

Issue #1091, opened 2019-04-27
Upgrade to Python 3


vpetersson:
When we created Screenly OSE, Python 3 was still in its early days. Things have changed, and Python 2.7 is reaching EOL next year. It's time to start upgrading our code base to Python 3.

ekevoo:
Hello, I have ported a different project from 2 to 3 in the past and I had zero problems, though it was definitely much smaller. In #1045 you mention that it's partly due to dependencies. Have you already identified which dependencies are affected?

vpetersson:
@ekevoo We'd love help with the upgrade to Python 3. Things have likely changed sin the initial comment in 1045, so I don't know if it is still the case. My gut feeling tells me that by now everything will work, and we will just need to refactor the code.

ekevoo:
I might be able to help, but I just installed Screenly for the first time yesterday and haven't looked at the code yet. I guess the two big initial questions are

1. What's the situation with unicode strings and byte strings? Are they marked (u"" and b"" or are there usually `from __future__ import unicode_literals`)?
2. I notice the installer spends quite some time and effort clearing out any potential `python3` and manually pulls `pip`, which is what made me come look for this issue; is there an easy way to disable that? I don't want to have to wrestle with it for days.

vpetersson:
> What's the situation with unicode strings and byte strings? Are they marked (u"" and b"" or are there usually from __future__ import unicode_literals)?

That's a good question - @rusko124?

> I notice the installer spends quite some time and effort clearing out any potential python3 and manually pulls pip, which is what made me come look for this issue; is there an easy way to disable that? I don't want to have to wrestle with it for days.

Sure, that's done in [here](https://github.com/Screenly/screenly-ose/blob/master/bin/install.sh).

ekevoo:
So I'm working at this today and I'm kind of puzzled on amf.py. I'm not sure what this does but it's clearly heavily encoding-related. I saw very few references to py2's str which is equivalent to py3's bytes.

vpetersson:
@rusko124 could you chime in here?

ekevoo:
@vpetersson Hey are you still interested in my help? My previous pull request is super stale and no longer usable, but I would appreciate if we could have a 30~60min meeting just to help me set up a testing environment so I can start over and build the PR in a way that the CI tools will be appeased.

vpetersson:
[...]

PR #1180, opened 2019-07-07
Assisted 2to3 run

ekevoo:
This is a first-pass attempt to address issue #1091.

I have run most `2to3` fixes on the codebase. I skipped `ws_comma` (seems pointless), `import` (sometimes breaks when `__name__ == '__main__'`), and `basestring` (we want `bytes` instead of `str` in those cases). About half of the rules yielded no changes, which is a good sign.

I did not actually *run* the code. I see there are three active branches and I don't know which one is the best one to do this work on, so I did the low hanging fruit effort and just walked through what `2to3` does, using my previous upgrading experience.

Therefore, this PR is mostly informational and probably incomplete- please check the diffs in the individual commits to see the scope of this migration effort. Something like this is *definitely* the first step in addressing #1091.


ekevoo:
I don't know why I wrote `fix -m` in the last one… it was `2to3 -f` as well :)

I checked Travis, most errors are due to the new py3 metaclass syntax. The other one is a style problem which is trivial to fix.

vpetersson:
Very nice @ekevoo! Let's get those tests working and then @rusko124 can do a review.

rusko124:
Hey @ekevoo !
Thanks for the work! Can you update the python version for `.travis.yml`, please? First of all I would like to ci passed.

ekevoo:
HFS this is Ubuntu Precise! It has python 3.3! Both of which went out of support 2 years ago.

I'll keep trying some more, but honestly, I can't embrace all this techincal debt :(

ekevoo:
Uh, I don't know where I got precise from. It's trusty, still a bit old but still in support. Python 3.0-3.4 are all kinda not used a lot anymore, and 3.5 is going down soon. I'm going to upgrade past xenial to bionic and see what happens.

ekevoo:
alright, proper runtime errors now!

i know what to do about the bytes/str ones, which i'll get around to some other day, but no promises about the ones that parse html. 🤷‍♂

vpetersson:
Awesome. Good job @ekevoo.

ekevoo:
...and this is all I got for now, which is clearly not enough. I'm not sure how to test the code because I've just been using the prebuilt image and I don't really have a spare pi I can take off and mess with. How would you suggest me to test this using a local amd64 Debian/Ubuntu VM?

ekevoo:
Alright! At least I got python 3.7 running now, which is important because that's the default python 3 version for raspbian buster. I added some end-of-life dates to `.travis.yml` while at that.

This is all I know to do. 🤷‍♂

ekevoo:
Huh? The PR used to only be rejected by Travis and now all three are rejecting it? :(

None of the things Codacy is pointing out were made by me. And I'm not permitted into snyk, but judging by the open branches by dependency-bot I have a good guess what it says. I'll wait for further commits to resolve those two.

vpetersson:
@rusko124 you have any insight on this?

rusko124:
PR Quality check should be fixed here #1208 PR
Snyk check doesn't relate with this PR(will be fix soon)
Thus we need to fix only travis CI.
FYI #1208 has changes for `.travis.yml`, please pull up the changes after merge the PR

ekevoo:
I'm having difficulty to find the time to finish this up. In the meantime can you please reply to my question at https://github.com/Screenly/screenly-ose/pull/1180#issuecomment-515709251

vpetersson:
@ekevoo We definitely appreciate all the effort you've put in on this. Is there anything we can do to assist you over the last hurdle? @rusko124 also please look at @ekevoo comment he referred to.

ekevoo:
That's all I can do without the ability to test the code. 🤷‍♂️

vpetersson:
Thanks @ekevoo. Looks like there is a merge conflict at the moment. Could you merge in master? Thanks.

ekevoo:
I had just done exactly that 😭 but fine, it's small enough.

The two (now three) merges that I did since the initial 2to3 run didn't go through 2to3 again. Many of its transformations aren't idempotent so we can't just re-run it on every merge and further merges all require manual adjustments. However you can follow the error messages and adjust.

vpetersson:
@rusko124 could you help out here?

PR #1185, opened 2019-07-10
Run nosetests from the Dockerfile 

### LINKED ISSUE #1181: Lack of developer documentation and difficulty running tests
I'm attempting to adapt screenly for a custom use-case, learning the code, etc.  I hit a wall running the tests on a local ubuntu 19.04 install.  Eventually, through looking through the travis-ci config, I was able to install enough dependencies to be able to run the tests, but hit a few roadblocks.  I'd like to highlight what those are for the next person and see if there is any low-hanging fruit that I could possibly help address.

Ideally, I'd like a way to reproduce the Travis-CI test environment in a local docker image, which can serve as documentation for what's needed to run the tests locally, outside of travis's own infra, and also be able to hack on it in a more local, but still isolated context, such as a pipenv virtual python environment.

I would be happy to contribute to documenting or improving these issues, and I'm not sure if it is desirable or inline with the project direction, so seeking some guidance.  What should be split out into a separate issue, what can be documented better, etc.

  - Lack of an isolated dev environment: In my brief foray into python, I've noticed there are a number of tools that are partial solutions to dependency management. Pipenv seems the best of these, which brings together pyenv, virtualenv, and deterministic dependency management similar to npm lock files.  I would propose switching from the requirements.txt to a Pipfile, and have generated one locally using pipenv's CLI with some success.  It's a necessity for developers working on more than one project at a time to minimize the shared dependencies needed and their interaction.

  - Not clear what python version the project requires.  I see the project is still on 2.7 (noticed the use of print statements), but adding a .python-version or Pipenv file would make it clearer.

  - The tests are really slow: specifically the splinter/selenium based tests require extra setup (I tried using a newer version of splinter, which tipped me to needing geckodriver, which I had to install manually), and the test running seems to wait for over a minute after the browser exits on an up-to-date firefox 66 (travis references 45).  I can't tell if I'm doing something wrong, or if I just need to turn off those tests, but I see it's much faster on travis-ci.  The issues were not googleable.  I think the root issue is too much dependency on an old OS and its deps (travis-ci and ubuntu 14.04).  IMO, the fix for this is more isolation, allowing the tests to be run in docker, with a headless firefox or XVFB setup.  This dockerfile can also serve as documentation for the minimal dependencies needed to work on the project. I attempted to alter the supplied Dockerfile, but hit so many problems that I wanted to ask here first.

  - The tests require files to be copied into paths in the user's home directory.  IMO the fix for this is to use env-vars to override the $HOME-relative paths to in-repo files only during a test scenario.

  - The tests require a daemon to be running. I would normally expect calling the nosetests CLI command to spin up whatever services are needed and tear them down at the end. This seems especially easy given the DB is just a sqlite file.  

Thanks for reading, please let me know what's desirable for the project or if there are any easy fixes to these issues I've missed.

gtrak:
With these changes, I am able to run the docker tests with a single command.

Before merging, I need to ensure the selenium version changes still work with the current Travis CI config, or also make the necessary changes there.

Fixes #1181

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:18.04
+FROM debian:stretch
 MAINTAINER Viktor Petersson <vpetersson@screenly.io>
 
 RUN apt-get update && apt-get -y install \
         build-essential \
         curl \
-        firefox \
+        firefox-esr \
         ffmpeg \
         git-core \
         libffi-dev \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,42 @@
-FROM debian:stretch
+FROM ubuntu:18.04
 MAINTAINER Viktor Petersson <vpetersson@screenly.io>
 
-RUN apt-get update && \
-    apt-get -y install \
+RUN apt-get update && apt-get -y install \
         build-essential \
         curl \
+        firefox \
         ffmpeg \
         git-core \
         libffi-dev \
+        libsdl2-2.0-0 \
         libssl-dev \
         lsb-release \
         mplayer \
         net-tools \
+        net-tools \
         procps \
         python-dev \
         python-gobject \
-        python-imaging \
+        python-gobject-2 \
         python-netifaces \
+        python-pil \
+        python-setuptools \
         python-simplejson \
         sqlite3 \
+        xvfb \
     && \
     apt-get clean
+
+# Gecko Driver
+ENV GECKODRIVER_VERSION 0.24.0
+RUN curl -L -o /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER_VERSION/geckodriver-v$GECKODRIVER_VERSION-linux64.tar.gz \
+  && rm -rf /opt/geckodriver \
+  && tar -C /opt -zxf /tmp/geckodriver.tar.gz \
+  && rm /tmp/geckodriver.tar.gz \
+  && mv /opt/geckodriver /opt/geckodriver-$GECKODRIVER_VERSION \
+  && chmod 755 /opt/geckodriver-$GECKODRIVER_VERSION \
+  && ln -fs /opt/geckodriver-$GECKODRIVER_VERSION /usr/bin/geckodriver \
+  && ln -fs /opt/geckodriver-$GECKODRIVER_VERSION /usr/bin/wires
 
 # Install Python requirements
 ADD requirements.txt /tmp/requirements.txt
@@ -41,6 +57,7 @@ COPY . /home/pi/screenly
 RUN chown -R pi:pi /home/pi
 
 USER pi
+
 WORKDIR /home/pi/screenly
 
 EXPOSE 8080
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -1,6 +1,6 @@
 nose==1.2.1
 mock==1.0.1
 pep8==1.7.0
-selenium==2.53.6
+selenium==3.141.0
 setuptools==34.4.0
-splinter==0.7.4
+splinter==0.11.0
--- a/run_docker_tests.sh
+++ b/run_docker_tests.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+docker build . -t screenly-ose-dev
+
+docker run --rm -it screenly-ose-dev bash -c '
+  mkdir -p ~/.screenly ~/.config/uzbl/ ~/screenly_assets
+  cp ansible/roles/screenly/files/uzbl-config ~/.config/uzbl/config-screenly
+  cp ansible/roles/screenly/files/screenly.conf ~/.screenly/
+  cp ansible/roles/screenly/files/screenly.db ~/.screenly/
+  cp ansible/roles/screenly/files/screenly_utils.sh /tmp/screenly_utils.sh
+  echo -e "[local]\nlocalhost ansible_connection=local" > ansible/localhost
+  curl https://www.screenly.io/upload/ose-logo.png > /tmp/image.png
+  curl https://www.screenly.io/upload/ose-logo.png > ~/screenly_assets/image.tmp
+  curl https://www.screenly.io/upload/big_buck_bunny_720p_10mb.flv > /tmp/video.flv
+  export DISPLAY=:99.0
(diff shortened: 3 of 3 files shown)


gtrak:
One of the tests is still failing, I'm not sure why:
```======================================================================
FAIL: test_add_asset_streaming (splinter_test.WebTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/screenly/tests/splinter_test.py", line 218, in test_add_asset_streaming
    self.assertEqual(len(assets), 1)
AssertionError: 0 != 1
-------------------- >> begin captured logging << --------------------
```

vpetersson:
If i'm not mistaken, this one spins up a temporary RTSP server, but due to how Docker works, it will not allow the connection.

gtrak:
I get the same error when I run with --net=host, and a different one if I turn off the rtsp daemon, so I don't think the issue is specifically about docker networking.

gtrak:
I did some more investigation.  One thing I noticed is rtmp.py is just super slow in this scenario for some reason:
The current selenium test timeout is 10s on the last step. If I curl, it finishes in 55.  If I increase the last test step timeout to 60, the test still fails, and it also fails at 120.
```
pi@46a6d9cffd97:~/screenly$ curl rtmp://localhost:1935/app/video.flv -o out.flv
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 10.0M    0 10.0M    0     0   185k      0 --:--:--  0:00:55 --:--:--  199k
```
I'm not sure where to look next, it seems the rtmp server project is really old and inactive.  I ran the test manually with firefox and my best guess is the asset check (works fine with skip_asset_check) is failing at the url_fails check: 

```def url_fails(url):
    """
    If it is streaming
    """
    if urlparse(url).scheme in ('rtsp', 'rtmp'):
        if arch in ('armv6l', 'armv7l'):
            run_omxplayer = omxplayer(url, info=True, _err_to_out=True, _ok_code=[0, 1])
            for line in run_omxplayer.split('\n'):
                if 'Input #0' in line:
                    return False
            return True
        else:
            run_mplayer = mplayer('-identify', '-frames', '0', '-nosound', url)
            for line in run_mplayer.split('\n'):
                if 'Clip info:' in line:
                    return False
            return True
```

```
mplayer -identify -frames 0 -nosound rtmp://localhost:1935/app/video.flv
Creating config file: /home/pi/.mplayer/config
MPlayer 1.3.0 (Debian), built with gcc-7 (C) 2000-2016 MPlayer Team
do_connect: could not connect to socket
connect: No such file or directory
Failed to open LIRC support. You will not be able to use your remote control.

Playing rtmp://localhost:1935/app/video.flv.
libavformat version 57.83.100 (external)
[tcp @ 0x7f757d46b180]Connection to tcp://localhost:1935 failed (Connection refused), trying next address
```

ffprobe from the get_video_duration function has a similar error.

These are also interesting:
```do_connect: could not connect to socket
 connect: No such file or directory
```

vpetersson:
This is because you cannot use 'localhost' inside a Docker container.

gtrak:
I have demonstrated curling against the localhost rtmp server above (it
binds to 0.0.0.0) and I'm pretty sure I've done so in the past. Running
with --net=host would remove any meaningful differences.

vpetersson:
@gtrak Sorry, my bad. You're right. I believe @rusko124 wrote these tests. Could you chime in?

rusko124:
Hey @gtrak !

@vpetersson Do we need to update the dockerfile to ubuntu 18.04?

Judging by this #1185  we want to update the Travis Ci  template to 18.04, but at the moment it is not technically possible as Travis does not support 18.04 https://docs.travis-ci.com/user/reference/overview/

Also some Travis Ci logs:
```
ERROR: test_add_asset_streaming (splinter_test.WebTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Screenly/screenly-ose/tests/splinter_test.py", line 200, in test_add_asset_streaming
    with Browser() as browser:
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/splinter/browser.py", line 64, in Browser
    return driver(*args, **kwargs)
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/splinter/driver/webdriver/firefox.py", line 78, in __init__
    **kwargs
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/selenium/webdriver/firefox/webdriver.py", line 164, in __init__
    self.service.start()
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/selenium/webdriver/common/service.py", line 83, in start
    os.path.basename(self.path), self.start_error_message)
WebDriverException: Message: 'geckodriver' executable needs to be in PATH. 
```

vpetersson:
> Do we need to update the dockerfile to ubuntu 18.04?

Ubuntu 18.04 is the current LTS release. 16.04 is also an LTS and got a few more years of lifetime so if there is a big amount of difference in work between 16.04 and 18.04, let's stick to 16.04.

rusko124:
@vpetersson I mean that currently we have `FROM debian:stretch`, ubuntu we use only for travis

vpetersson:
Ah i see - well, we should probably stick to Debian, as it's closer to Raspbian, rather than moving to Ubuntu.

gtrak:
Travis is supporting 18.04 imminently:
https://travis-ci.community/t/ubuntu-18-04-1-lts-bionic-beaver/1270/23

I think we'll have to figure out this rtmp issue sooner or later, but if it
works on 16.04, I'm happy to stop thinking about it. I don't expect Ubuntu
vs Debian to make much of a difference in terms of compatibility, but am
personally more familiar with Ubuntu.

gtrak:
I tested both `FROM debian:buster` and `FROM debian:stretch` to see if there's a difference, they both fail the same rtmp-related test as the ubuntu image.

How should we proceed?

I would recommend sticking to stretch in the docker image since we're worried about raspbian compatibility, and updating travis to 16.04 at least.  

But we still need to fix the rtmp test. Notably, I traced that code down, and the binary it calls into  special-cased by arch, so already different from raspberry pi regardless of distro.

gtrak:
It appears to me we need to determine if RTMP is broken on either the server or client side, then work around it, but since I have little context on how that's supposed to work, I'd like to avoid digging into it if someone has more knowledge :-).

gtrak:
Looks like we want buster after all #1187

vpetersson:
@gtrak Adding support for Buster doesn't necessarily mean we need to use that as the test image.

Issue #1168, opened 2019-06-24
Issue upgrading on NOOBS


vpetersson:
#### Overview of the Issue

When upgrading on a NOOBS installation (to development release), the upgrade fails with the error:

```
TASK [network : Download resin-wifi-connect release] **********************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'not resin_wifi_version_file_exist' failed. The error was: error while evaluating conditional (not resin_wifi_version_file_exist): 'resin_wifi_version_file_exist' is undefined\n\nThe error appears to be in '/home/pi/screenly/ansible/roles/network/tasks/main.yml': line 87, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Download resin-wifi-connect release\n  ^ here\n"}

PLAY RECAP ****************************************************************************************************************************************
localhost                  : ok=50   changed=9    unreachable=0    failed=1    skipped=15   rescued=0    ignored=0
```

#### Reproduction Steps

Steps to reproduce this issue, eg:

* Install using Noobs
* run `bin/run_upgrade.sh`
* Select the development release

### Environment

* Raspberry Pi Hardware Version: Model 3 B+
* Screenly OSE Version: Sprint 9 Completed

rusko124:
Fix merged, Issue can be close.

PR #1190, opened 2019-07-12
Fix: network tasks of ansible

rusko124:
Ref: #1168


vpetersson:
Will this work with Noobs?

rusko124:
Yes, PR was tested on Noobs

Issue #1244, opened 2019-10-27
Rabbitmq keeps stopping and restarting


trustees629:
Just updated my Pi zero's to bd67ec7 and rabbitmq keeps stopping and restarting. I'm running on Balena. Has anyone else had this issue? 

Also, screenly-celery shows as downloaded but nothing else happens.

vpetersson:
@trustees629 Can you provide some logs for this?

trustees629:
Yes I can but I'm not exactly sure how to pull them from Balena. Can you help?

vpetersson:
@trustees629 You can get the container logs directly from their interface by selecting the right container.

trustees629:
Ok, I can get into screenly-websocket, screenly-viewer, and screenly-server. I looked for screenly_viewer.log and it's nowhere to be found. Thoughts?

ChuckIT:
Hi I am experiencing the same problem using Balena OS 2.43.0+rev1 on an RPi3 Screenly is only running an announcement video of still images less than two minutes in length. It runs OK for several days but then hits the RabbitMQ restart loop and the device goes offline. 
system log is at https://pastebin.com/9BymhpuE

vpetersson:
@rusko124 any clues here?

trustees629:
Here's a snip from my logs.

04.11.19 12:00:33 (-0500) <screenly-viewer> Showing asset important.jpg (image)
04.11.19 12:00:33 (-0500) <screenly-viewer> Sleeping for 10
04.11.19 12:00:38 (-0500) Restarting service 'rabbitmq sha256:3b74d2cb896b533ef4ce1c734c725caff5a3b2a7b2ddf4a81e12f275d1d2b5dd'
04.11.19 12:00:43 (-0500) Service exited 'rabbitmq sha256:3b74d2cb896b533ef4ce1c734c725caff5a3b2a7b2ddf4a81e12f275d1d2b5dd'
04.11.19 12:00:46 (-0500) <screenly-viewer> Generating asset-list...
04.11.19 12:00:46 (-0500) <screenly-viewer> Showing asset wifi.png (image)
04.11.19 12:00:46 (-0500) <screenly-viewer> Sleeping for 10
04.11.19 12:00:58 (-0500) <screenly-viewer> Generating asset-list...
04.11.19 12:00:58 (-0500) <screenly-viewer> Showing asset drinksbefore.png (image)

vpetersson:
[...]

PR #1249, opened 2019-11-06
Edits: docker-compose.dev.yml

rusko124:
Ref #1244

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -51,8 +51,6 @@ services:
     image: redis:alpine
     logging:
       driver: none
-    ports:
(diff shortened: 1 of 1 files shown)


vpetersson:
Let's only expose services that needs to be expose and refactor this to use docker networking.

rusko124:
@vpetersson I cleared the compose file from `expose`. And configured the other services to use only need ports.

vpetersson:
> @vpetersson I cleared the compose file against `expose`. And configured the other services to use only need ports.

No, we're still exposing Redis. Redis is only used internally, so it doesn't need to be exposed.

PR #1252, opened 2019-11-08
Edits: docker-compose.yml

rusko124:
Ref #1244

Edits: 
- Containers are isolated(Instead of network_mode: host)

Problems:
- The splash page shows the docker IP, not the device IP
- Swagger also refers to the wrong IP(need to edit the IP manually)
swagger1
swagger2

@vpetersson I need your review and opinion


vpetersson:
Looks good! Merging.

PR #1269, opened 2019-11-22
Update utils.py

ealmonte32:
fix for "Unable to resolve local IP address" for when using wlan..

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -85,9 +85,9 @@ def get_node_ip():
     that is being used as the default gateway.
     This should work on both MacOS X and Linux."""
     try:
-        time.sleep(15)
         default_interface = gateways()['default'][AF_INET][1]
         my_ip = ifaddresses(default_interface)[AF_INET][0]['addr']
+        time.sleep(5)
         return my_ip
     except (KeyError, ValueError):
         raise Exception("Unable to resolve local IP address.")
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -8,6 +8,7 @@ import re
 import requests
 import string
 import sh
+import time
 
 from datetime import datetime, timedelta
 from distutils.util import strtobool
@@ -84,6 +85,7 @@ def get_node_ip():
     that is being used as the default gateway.
     This should work on both MacOS X and Linux."""
     try:
+        time.sleep(15)
         default_interface = gateways()['default'][AF_INET][1]
         my_ip = ifaddresses(default_interface)[AF_INET][0]['addr']
         return my_ip


vpetersson:
@ealmonte32 It's generally considered bad practice to add random sleep functions. Instead of sleeping, let's change this and make it wait for a condition instead.

ealmonte32:
i was reading and yes that is correct, we want to use a wait function I was just using a quick alternative.. i dont know how to edit a pull request once it has been submitted.. i use the web interface more than the git in termimal so i had to delete my forked branch in order to delete this PR.. 🤦‍♂

vpetersson:
>  i dont know how to edit a pull request once it has been submitted

Normally you fork the repo and then create your own branch there where you can push your changes to. These updates are then automatically reflected in the PR.

There are pretty decent UIs for Git, like Git Tower if you prefer that.

vpetersson:
@rusko124 Could you provide some pointers here on how to improve this w/out having to use `sleep()`?

rusko124:
I can suggest passing retry variable and wait timeout between iterations variable to get_node_ip function. 
For example:
```
def get_node_ip(retry=3, timeout=0.5):
    """Returns the node's IP, for the interface
    that is being used as the default gateway.
    This should work on both MacOS X and Linux."""
    for _ in range(retry):
        try:
            default_interface = gateways()['default'][AF_INET][1]
            my_ip = ifaddresses(default_interface)[AF_INET][0]['addr']
            return my_ip
        except (KeyError, ValueError):
            time.sleep(timeout)
    raise Exception("Unable to resolve local IP address.")
```

ealmonte32:
@rusko124 
Thanks, but I still see you use the time.sleep() function there..
Can you tell me why me putting the time.sleep(5) before returning `my_ip` (as shown below) is not good on this case?

```
def get_node_ip():
    """Returns the node's IP, for the interface
    that is being used as the default gateway.
    This should work on both MacOS X and Linux."""
    try:
        default_interface = gateways()['default'][AF_INET][1]
        my_ip = ifaddresses(default_interface)[AF_INET][0]['addr']
        time.sleep(5)
        return my_ip
    except (KeyError, ValueError):
        raise Exception("Unable to resolve local IP address.")
```

I have implemented yours and just tested 3 times by resetting wifi every time and so far i do not get the exception, so it works thanks, will create new PR for it.

P.S. wanted to tell you when I click Reset Wifi on webUI and screenly is restarted, there is a white screen background showing the ad-hoc AP info instead of the screenly logo background one.. have you seen this before? want me to create ticket for it?

Issue #1270, opened 2019-11-24
[Feature] Add memory info under System info


Gifford47:
Hi!
it would be useful, if you could see under the tab System info beside "load average" and "free space" also the occupied memory in percent.

vpetersson:
Seems like a sensible feature.

PR #1272, opened 2019-11-27
Memory settings

rusko124:
Ref #1270

Screenshot:
изображение


vpetersson:
@rusko124 Please use "Closes #X" in the description, as that will automatically close the related issue. Right now it's just referencing.

ealmonte32:
@rusko124 
Do you think it is good to use human readable format for memory?
Sample: https://github.com/giampaolo/psutil/blob/master/scripts/meminfo.py

PR #1274, opened 2019-11-27
Update utils.py

ealmonte32:
fix for wlan connections that were raising: `Exception("Unable to resolve local IP address.")`

_should we import time module?_

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -79,16 +79,20 @@ def validate_url(string):
     return bool(checker.scheme in ('http', 'https', 'rtsp', 'rtmp') and checker.netloc)
 
 
-def get_node_ip():
+def get_node_ip(retry=3, timeout=0.5):
     """Returns the node's IP, for the interface
     that is being used as the default gateway.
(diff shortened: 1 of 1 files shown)


vpetersson:
@ealmonte32 check the comment from the Codacity quality review.

ealmonte32:
if it is because i did not import the time module then i wonder how it still works without it being imported on the test device.. 🤔 

also, i thought you did not want to use time.sleep() so how come this is fine?
i thought when i added a simple sleep for 5 seconds or so before the return of `my_ip` that it would suffice and not cause any adverse side effects..

vpetersson:
> if it is because i did not import the time module then i wonder how it still works without it being imported on the test device.. 🤔
> 
> also, i thought you did not want to use time.sleep() so how come this is fine?
> i thought when i added a simple sleep for 5 seconds or so before the return of `my_ip` that it would suffice and not cause any adverse side effects..

Well, I'm not keen on it, but if @rusko124 is saying it's OK - then I'm OK with it. :)

PR #1281, opened 2019-12-27
Update main.yml

ealmonte32:
Remove cups related packages since screenly is not used for printing and these packages start services and daemon and is a waste of resources.


ealmonte32:
image

vpetersson:
Good find, @ealmonte32.

PR #1286, opened 2020-01-13
Update install.sh

ealmonte32:
Cleaning up unused packages by running autoremove...

I noticed so many leftover packages from earlang and other lib packages when going from Screenly Image to EXP branch and then to DEV branch, I realize this is better than putting all the names in ansible after switching branches.. of course the package that requires these dependencies would need to be removed through ansible during installation depending which branch you choose, but again, it seems better to add one main package there and let the autoremove take care of rest, unless you think it is better to add all packages specifically in the ansible playbook.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -165,6 +165,7 @@ sudo -E ansible-playbook site.yml $EXTRA_ARGS
 
 sudo apt-get autoclean
 sudo apt-get clean
+sudo apt autoremove -y
 sudo find /usr/share/doc -depth -type f ! -name copyright -delete
 sudo find /usr/share/doc -empty -delete
 sudo rm -rf /usr/share/man /usr/share/groff /usr/share/info /usr/share/lintian /usr/share/linda /var/cache/man


vpetersson:
Sensible @ealmonte32. Added it for review.

rusko124:
@ealmonte32 Since we don't update ansible in requirements.txt this could be cause an error while ansible-playbook is working. Did you run playbook with the changes?

ealmonte32:
> @ealmonte32 Since we don't update ansible in requirements.txt this could be cause an error while ansible-playbook is working. Did you run playbook with the changes?

I just changed the version here to 2.9.2 because of it being latest version at that moment, and i also removed it from requirements so that it doesnt cause error while ansible playbook running.. so this PR is literally just to update ansible version during the install before calling the playbook.

ealmonte32:
closing this but it doesn't make sense to have two sections for ansible, if it is required for the entire ansible-playbook, this makes ansible a main requirement for the installer, just like pip, so why keep track of ansible in both the install.sh and requirements.txt (yes we can make a script that installs the same version by looking into the requirements and grepping it, but cant think of a real reason it must stay in requirements.txt as well..)

PR #1287, opened 2020-01-13
Update requirements.txt

ealmonte32:
minor update for celery and ansible

PR #1298, opened 2020-02-09
Update requirements.txt

ealmonte32:
ansible is already called for install on `install.sh` so no need to set here as well as it can conflict while the playbook is running.. (as rusko mentioned)

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,6 +1,5 @@
 Mako==0.7.3
 MarkupSafe==1.0
-ansible==2.8.2
 celery==4.3.0
 certifi==2018.4.16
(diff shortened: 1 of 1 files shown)


vpetersson:
I'm torn about this one. Having it in two places makes no sense. However, it feels like it belongs in `requirements.txt` instead. 

What if we do something like this instead?
```
sudo pip install $(grep ansible requirements.txt)
```

I haven't tested this, but it's just an idea. We'd have to verify it both using `run-upgrade.sh` and  `install.sh` of course.

ealmonte32:
closing this one since this duplicate of https://github.com/Screenly/screenly-ose/pull/1287

PR #1299, opened 2020-02-09
Use alpine base images

escoand:
I tried to save some space and started to switch the containers to the alpine base images.

- [x] redis: 152MB --> 63MB
- [x] screenly-celery: 572MB --> 141MB
- [ ] screenly-server: 571MB
- [ ] screenly-viewer: 704MB
- [x] screenly-websocket: 343MB --> 272MB

### DIFF
--- a/docker/Dockerfile.celery.template
+++ b/docker/Dockerfile.celery.template
@@ -7,8 +7,8 @@ RUN apk --no-cache add \
         libcec
 
 # Install Python requirements
-RUN curl -s https://bootstrap.pypa.io/get-pip.py | python
-RUN pip install \
+RUN curl -s https://bootstrap.pypa.io/get-pip.py | python2
+RUN pip2 install \
     celery-redbeat==0.13.0
 
 COPY . /tmp/screenly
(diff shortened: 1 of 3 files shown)


vpetersson:
Awesome! I've been meaning to do this for some time.

escoand:
I was wondering why celery, server and viewer install all the requirements. Not all of the required debian packages exist for alpine.

ealmonte32:
@escoand Have you tested this with a pi zero w? this is the hardware that with its limited resources will benefit the most.. I always wanted to test the alpine image as well, will comment with feedback once I test.
This, combined with memory limits I added to the services, the pi zero w (and others) will def not hang, @vpetersson if this works as we hope, we can close that "black screen" issue once and for all, you can actually close it after both service PR's get approved since users wont see black screen anymore.

escoand:
Redis was really simple and the state is finished I would say. The websockets image should be finished as well, just did some small tests. The celery image is definitely work-in-progress. The image is building but there are exceptions at the moment. Currently I'm busy with other projects so it's sleeping.

My biggest problem are the mentioned python requirements. For me it doesn't make sense that 3 images need all of them. But I don't know which of them are needed for which kind of image, so it's time consuming trial and error.

escoand:
I've some more changes locally but I'm not at the point to push them.

vpetersson:
Good stuff, guys! I would just chime in and be mindful of the arch. I know that *some* upstream images are only available for ARMv7 (all modern Pis), whereas the Zero is still on ARMv6.

PR #1304, opened 2020-02-21
Update screenly-viewer.service

ealmonte32:
To prevent screenly-viewer from freezing the pi because of uzbl using too much memory when loading certain assets, we control the memory limits of the service. I was undecided as to which signal the KillMode should be set to, either default control-group or mixed, but since mixed accomplishes what control-group does in the end while sending different signals, I thought it'd be best to use that method.

### DIFF
--- a/ansible/roles/screenly/files/screenly-viewer.service
+++ b/ansible/roles/screenly/files/screenly-viewer.service
@@ -5,9 +5,19 @@ After=matchbox.service screenly-web.service
 [Service]
 WorkingDirectory=/home/pi/screenly
 User=pi
-MemoryLimit=60%
+
+# CPU Limits
+CPUAccounting=yes
+CPUQuota=100%
+CPUQuotaPeriodSec=60
+
+# Memory Limits
 MemoryAccounting=yes
+MemoryLimit=60%
+
+# Kill Signal control
 KillMode=control-group
+SendSIGHUP=yes
 
 Environment=DISPLAY=:0.0
 
--- a/ansible/roles/screenly/files/screenly-viewer.service
+++ b/ansible/roles/screenly/files/screenly-viewer.service
@@ -10,12 +10,13 @@ User=pi
 CPUAccounting=yes
 CPUQuota=100%
 
-# Memory Limits
+# Memory Limits as stopgap to prevent system lockup due to resource exhaustion
 MemoryAccounting=yes
-MemoryLimit=65%
+MemoryLimit=60%
 
 # Kill Signal control
 KillMode=control-group
+KillSignal=SIGKILL
 SendSIGHUP=yes
 
 Environment=DISPLAY=:0.0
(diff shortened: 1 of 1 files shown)


ealmonte32:
needed to lower memory allowance due to celery needing more.. also decided to go with default control-group

ealmonte32:
Just wanted to show all of the heavy assets that are being used for this test, and it is properly killing the service once the viewer (uzbl) exhausts the resources, we should have done this so long ago, thanks to you for the cgroups idea, I used to have to get up all the time when the pi froze because of memory, and had to physically pull the plug on it.. now I am 99% sure it wont do that, but we need to have precise control on the memorylimit and any other services because we have to take into consideration their total combined memory usage and not just the viewer (uzbl)..

image

ealmonte32:
Just an update, it's been 13+ hours now and screenly is still running, no hangs, no frozen system, if this memory limit was not on, the system would freeze/hang within 30mins or less.
I am now very confident in saying that this is a very good addition to the viewer service.

ealmonte32:
@vpetersson 
This is an update to again confirm that current cgroup memory limits as set on this and the celery service PR have allowed me to fill up the Assets with websites that as you can see from the memory error logs and oom-killer would have frozen the Pi within the hour and most likely required the user to physically go and reset the Pi.

Here are the screenshots from my tests, @rusko124 please test and approve soon so that we can tell any user that is experiencing any memory related issue to test the development version and we can get even more feedback.

image
image
image
image
image

ealmonte32:
To explain why such changes:
These limits are needed because when the uzbl browser and other services/processes consume all memory, the Pi becomes unusable and hangs, and requires a hard reset, most of the time requiring physical attention.

Observations:
Celery service uses more than 100MB of memory right from the start with about 10 assets..
Screenly Web socket service uses about 20MB..
Screenly Web service uses about 50MB..

This leaves the Viewer service very little left for the browser, so this has been the only sort of recovery control I have been able to come up with so that screenly and certain assets dont render the Pi useless after a few heavy assets are added..

I dont know what the best number that applies to all users is for LimitCPU, but just to be safe, I think if an asset spends more than 300 seconds (5mins) in CPU at a Running state (R), then that usually means leak or something wrong, and unless you manually restart that service or kill that process, it wont recover, and it is usually the uzbl-core.. again, we are talking specifically for the screenly-ose version that uses uzbl in its current form.. these limits and tweaks may not be needed down the road for qt web view, but I believe a Pi that can self-heal/recover when it encounters these issues by making use of the systemd/service is definitely a good thing.

I know we should focus on experimental and qt web view browser, but most are not using that yet, so this is why I've focused on master branch.

attachment: example of uzbl in Running state forever (i've seen it stuck there for 20+hours during some tests)
Screen Shot 2020-03-10 at 1 36 57 AM

ealmonte32:
So again, the good thing is that these changes are doing their job at making sure the pi doesnt freeze and become unusable, but overnight, I found the viewer.py process running, but uzbl-core dead, and the viewer did not restart the browser, so I wonder how we can easily adjust this part of the viewer.py code that checks for browser dead, to just maybe check in a similar way like
`pgrep -f uzbl-core`
and then `if [ $? != 0 ]; then pkill -f viewer.py` to restart the viewer which worked for me, which will just restart the viewer and browser and everything started running once again.. all a self-healing self-recover process.. there is probably a python way to do it without using `os.system('commands or run script and release with &')` but I'm just not experienced with python for that.. 😞

PR #1305, opened 2020-02-21
Update screenly-celery.service

ealmonte32:
Noticed celery would consume a lot of memory when Pi was left running screenly with many assets for many hours, so we can prevent this from eating up resources by also limiting the memory like with the viewer service. This time I left the KillMode as default.


ealmonte32:
Ugh, celery is such a memory hog... 50MB was not enough to even start it, when i looked at logs process kept being killed, so giving it 100M but this requires then lower screenly-viewer allowance , which I've now edited to 60%...

image

PR #1313, opened 2020-02-28
Update installer with whiptail dialogs

rchadgray:
I have updated the yes/no questions with whiptail dialogs.  I let the terminal show what was going on and displayed whiptail dialogs when a response from the user is needed.


bin/install.sh
```diff
@@ -82,19 +133,25 @@ if [ ! -f /etc/locale.gen ]; then
   sudo locale-gen
 fi
 
-
 sudo sed -i 's/apt.screenlyapp.com/archive.raspbian.org/g' /etc/apt/sources.list
-sudo apt-get update
+sudo apt update -y
```
vpetersson: `apt` will break in older versions of Debian (must use `apt-get`). I'm not sure how common this is, but I don't see much upside in using `apt` over `apt-get` (other than it is newer).

rchadgray: I updated that line with apt-get.  That was there in the original code.  I did not notice it.

rchadgray:
I have never used this install.sh file to install OSE.  So far my Raspberry Pi does not reboot and start using ScreenlyOSE like when i burn the image you provide to an SD card.

ealmonte32:
@rchadgray 
The install script needs to be updated to change all those `if statements` with `case` ones, for much cleaner code.
I wanted to work on that sometime in the upcoming weeks, but then as a side effect of the change to case statements, a lot of the install script needs to change to include functions, etc, when building the whiptail menu, thus becomes a bit more work, but it is better to use than if/else statements.

ealmonte32:
> I have never used this install.sh file to install OSE. So far my Raspberry Pi does not reboot and start using ScreenlyOSE like when i burn the image you provide to an SD card.

Did you create an Issue for this? that is better then posting on a PR that is about something else.

rchadgray:
> > I have never used this install.sh file to install OSE. So far my Raspberry Pi does not reboot and start using ScreenlyOSE like when i burn the image you provide to an SD card.
> 
> Did you create an Issue for this? that is better then posting on a PR that is about something else.

Let me do some more testing to be sure. Last time i tried with bash <(curl -sL https://www.screenly.io/install-ose.sh) it got stuck and killed my SSH session at this line:

Created symlink /etc/systemd/system/timers.target.wants/wott-agent-self-update.timer → /lib/systemd/system/wott-agent-self-update.timer.

I will test some more.  Sorry for posting in the PR section.  
Chad

PR #1317, opened 2020-03-04
Update install.sh

ealmonte32:
Many users report not knowing how to get a specific version of Screenly, thus changing the installer to show all three options seems like a better way to offer the different installations supported.


ealmonte32:
Not sure why this is failing... what is celery trying to do?

```
FAIL: test_result (celery_tasks_test.TestUpgradeScreenly)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Screenly/screenly-ose/tests/celery_tasks_test.py", line 32, in test_result
    self.assertEqual(self.upgrade_screenly_result, {'status': 'Invalid -b parameter.\n'})
AssertionError: {'status': '\x1b[3;J\x1b[H\x1b[2JInvalid -b parameter.\n'} != {'status': 'Invalid -b parameter.\n'}
- {'status': '\x1b[3;J\x1b[H\x1b[2JInvalid -b parameter.\n'}
?             ---------------------
+ {'status': 'Invalid -b parameter.\n'}
-------------------- >> begin captured logging << --------------------
celery.app.trace: INFO: Task server.upgrade_screenly[d28adb9b-5676-4fb7-8695-a8f9b37454e0] succeeded in 1.005017137s: {'status': 'eter.
'}
```

ealmonte32:
@rusko124  you need to remove the double quotes from travis firefox download URL, that is why it is throwing the error..

image

ealmonte32:
I would think before the whiptail installation gets finalized, we can use this install.sh file because of the easier selection of branches rather than the old style. This installation gives you all options at once rather than asking for one first and then another if that option wasnt selected.. like i originally said, most users would rather see all selections and pick one branch and know they picked that one rather than the if/else selection..

PR #1318, opened 2020-03-05
Update requirements.txt

ealmonte32:
pyOpenSSL==19.1.0 was missing and was causing production version install to fail.


ealmonte32:
Why is travis trying to install the requirements.dev.txt if I am committing this into the production branch..?

```
The command "pip install -r requirements/requirements.dev.txt" failed and exited with 1 during .
```

vpetersson:
> Why is travis trying to install the requirements.dev.txt if I am committing this into the production branch..?
> 
> ```
> The command "pip install -r requirements/requirements.dev.txt" failed and exited with 1 during .
> ```

If I'm not mistaken, the `requirements.dev.txt` is used for running some of the unit tests.

ealmonte32:
Realized requests[security] already installs pyopenssl along with cryptography and idna, so the reason for error must be this below, a little mess of packages and the dependencies they install while then calling different versions on the requirements.txt file...

While testing the requirements.txt by itself from production, I was getting this:

```
ERROR: requests 2.20.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.1 which is incompatible.
ERROR: cryptography 2.8 has requirement cffi!=1.11.3,>=1.8, but you'll have cffi 1.7.0 which is incompatible.
```
So I decided to start making one by one installation steps to see where we need to make the changes for the fix..

When installing ansible, this already installs certain dependency packages that are newer than what the requirements file includes, and thus the conflict starts, and then packages begin to be removed and installed, etc, wasting time during installer..

This screenshot is from installing ansible==2.8.2 , so as you can see, those packages that are already referenced and installed, could be removed from requirements because then you are reinstalling most and would need to keep track of which version ansible installs and put that same version in requirements..

But, since production version needs so many fixes already included in master branch, we need to merge the master branch already with production since current production installation doesnt work... and even after I make the installation work by editing and fixing the requirements, the screenly viewer doesnt start.

image

PR #1321, opened 2020-03-13
Update to viewer.py for "&" symbol parsing

ealmonte32:
fix for uzbl '&' symbol parsing..


ealmonte32:
Issue came up when users would put google slides presentation as asset, and the uzbl URL would replace the & in query string for two or more `amp;` words, and thus the `loop`, `delay` query of slide transition would not get properly passed..

PR #1341, opened 2020-07-02
Update installer with case option for branch and ansible pulled from req file

ealmonte32:


ealmonte32:
reason for /usr/share/info to not be deleted entirely but just the contents:
```
Reading package lists... Done
Building dependency tree
Reading state information... Done
0 upgraded, 0 newly installed, 0 to remove and 37 not upgraded.
1 not fully installed or removed.
After this operation, 0 B of additional disk space will be used.
Setting up install-info (6.3.0.dfsg.1-1+b1) ...
Not a directory: /usr/share/info.
```

ealmonte32:
i will edit the other update to install.sh and add this to it so there isn't a conflict and it takes care of two issues in one.

PR #1343, opened 2020-07-13
Update README.md

ealmonte32:
updates of incorrect links


vpetersson:
@ealmonte32 Thanks! There's a merge conflict, but also the support link should be updated to https://support.screenly.io/hc/en-us/categories/360002606694-OSE

ealmonte32:
Thanks, i'll close this since it looks like the file was updated already with the correct links.

PR #1346, opened 2020-07-13
Update install.sh

ealmonte32:
_force ansible dir downstream via override_

from experience, if users get errors during ansible installation, the entire installation fails, unless the force option is set here to override and pull down the fresh ansible dir from repo, they need to manually copy the entire ansible folder from github to their Pi using wget or curl.. that should not be, i dont see any reason why the master ansible repo should not be forced downstream on the device where screenly-ose is being installed.. if there is a reason i am missing please let me know, i cant think of one at the moment.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -167,7 +167,7 @@ sudo apt-get autoclean
 sudo apt-get clean
 sudo find /usr/share/doc -depth -type f ! -name copyright -delete
 sudo find /usr/share/doc -empty -delete
-sudo rm -rf /usr/share/man /usr/share/groff /usr/share/info /usr/share/lintian /usr/share/linda /var/cache/man
+sudo rm -rf /usr/share/man /usr/share/groff /usr/share/info/* /usr/share/lintian /usr/share/linda /var/cache/man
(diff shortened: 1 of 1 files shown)


ealmonte32:
some of these PR requests that are going stale should have been merged if no one objects to what it accomplishes.. such as this one, i still can't think of a reason why ansible directory should not be forced downstream from the github repo upon installations or reinstalls if it solves any potential error that might arise as i experienced when i was troubleshooting and needed to reinstall in-place.. remember all it does is force the ansible directory from github to install over anything stale or corrupt in that directory already, and if you are installing screenly then you obviously need the correct latest files from that ansible directory..

ealmonte32 (commented):
deleting the info directory creates an error when reinstalling, so instead, we simply delete the contents of the directory to avoid errors.

ealmonte32 (commented):
@rusko124  or @vpetersson 
please review...

ealmonte32:
went line by line with comparing the current master branch install.sh and the changes i made, everything should be good now including fixes.

ealmonte32:
created new PR with only current changes clearly visible and listed.

PR #1348, opened 2020-07-14
New: config for install.sh

rusko124:
How it works:
- Setup repository(with credentials if repo is private) and branch.
- Run script `./bin/run_upgrade.sh` or use web upgrade.
- Upgrade should pass w/out errors.


bin/install.sh
```diff
@@ -5,6 +5,10 @@ BRANCH_VERSION=
 MANAGE_NETWORK=
 UPGRADE_SYSTEM=
 
+if [ -f config ]; then
```
vpetersson, line 8: Let's rename this file to `.env` instead, as that has emerged as the standard.

PR #1349, opened 2020-07-14
Update viewer.py

ealmonte32:
using resource limits to prevent viewer from making pi unresponsive if memory leak occurs


ealmonte32:
this is so annoying, no undo buttons, no delete commit from the browser GUI, a simple mistake makes me have to create a whole new PR, ridiculous

PR #1350, opened 2020-07-14
Update viewer.py

ealmonte32:
setting memory resource limit on the viewer to prevent unresponsive pi when too much memory consumption occurs by either memory leak or too many assets

_the imports were just re-arranged by ascending order, nothing was removed_

### DIFF
--- a/viewer.py
+++ b/viewer.py
@@ -41,7 +41,7 @@ exec(setlimit)
 
 logging.info("The current memory resource limit is set as:")
 logging.info(setlimit)
-#
+# setting memory resource limits finalized
 
 __author__ = "Screenly, Inc"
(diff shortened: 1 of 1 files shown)


vpetersson:
@ealmonte32 I think this would be a lot easier (and more robust) to do in Docker. Here's the relevant [documentation](https://docs.docker.com/config/containers/resource_constraints/).

ealmonte32:
Yes definitely, I was looking into that but I realized that until we move the different parts to Docker, this would be one of the ways to prevent the viewer from rendering the RPi unresponsive if too many heavy assets are used and for some reason oom-killer was not properly restarting the viewer after certain amount of crashes.
But I guess I can keep this as my backup memory limit code and focus on dockerizing the viewer.

PR #1374, opened 2020-10-06
Update start_balena.sh

ealmonte32:
turning off swap to reduce wear on SD card, and deleting 25mb of temp files from each container as well..

if anyone knows a reason why we shouldnt do this pls let me know, thanks.


ealmonte32:
will close for now cause did not test thoroughly but will reopen with way to remove up to 25MB of tmp files from container image

PR #1377, opened 2020-10-07
Update utils.py

ealmonte32:
removing the letter `i` so that it is never confused with `l` (lowercase L) from far away, although lowercase L is not included in the string of lowercase letters, users dont normally know this.


vpetersson:
Good catch!

PR #1378, opened 2020-10-07
Update .dockerignore

ealmonte32:
need to ignore .git directory , during some builds, this can grow multiple megabytes in size..

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,9 @@
 *.py[co]
 *.git
+*.github
+docs/*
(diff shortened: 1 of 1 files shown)


vpetersson:
We can't include the tests, as I think we need this to run the unit tests on CI.

ealmonte32:
> We can't include the tests, as I think we need this to run the unit tests on CI.

Done, fixed.

True.. I was tunnel-visioned on balena and dockerfile building that forgot about this.

PR #1239, opened 2019-10-14
Edits: experimental browser as default

rusko124:
**WARNING**: At the moment, this PR is not ready to merge.
Necessary points:
- Launch on balenaCloud(in progress)
- Prepare docker images for pulls

PR #1385, opened 2020-10-18
Edit of install.sh to have option to use local repo

Kipjr:
Ansible will always try to get the Screenly-OSE `$REPOSITORY` https://github.com/Screenly/screenly-ose when executing the install.sh file. For local development this is perhaps not wanted. Therefore I would like to propose the option to retrieve the local repository.

For example: 
`git clone https://github.com/Screenly/screenly-ose.git `
`mv screenly-ose screenly`
`chmod +x screenly/bin/install.sh`
`./screenly/bin/install.sh`

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -83,7 +83,7 @@ EOF
         echo
         exit 1
       fi
-      LOCAL_INSTALL="true"
+      DOCKER_TAG="latest"
     else
       echo "No /home/pi/screenly/ found. Please rename or create and run this script"
       exit 1
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -58,23 +58,8 @@ EOF
 
   echo  
   read -p "Are you executing this script from a local repository? (/home/pi/screenly/bin/install.sh) (y/N)" -n 1 -r -s LOCAL && echo
-  if [ "$LOCAL" != 'y' ]; then
-    if [ -z "${BRANCH}" ]; then
-      echo && read -p "Would you like to use the experimental branch? It contains the last major changes, such as the new browser and migrating to Docker (y/N)" -n 1 -r -s EXP && echo
-      if [ "$EXP" != 'y'  ]; then
-        echo && read -p "Would you like to use the development (master) branch? You will get the latest features, but things may break. (y/N)" -n 1 -r -s DEV && echo
-        if [ "$DEV" != 'y'  ]; then
-          export DOCKER_TAG="production"
-          BRANCH="production"
-        else
-          export DOCKER_TAG="latest"
-          BRANCH="master"
-        fi
-      else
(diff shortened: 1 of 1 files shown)


Kipjr:
This should work now however, when #1239 is accepted, the following needs to be changed with this documentation:
https://docs.ansible.com/ansible/2.6/modules/docker_container_module.html
https://docs.ansible.com/ansible/2.6/modules/docker_image_module.html

in ansible/screenly/tasks/main.yml:
`- name: Create screenly-viewer container`

to
```
- name: Create screenly-viewer image
  docker_image:
    name: screenly-ose-viewer
    source: build
    build: 
        pull: no
        path: /home/pi/screenly/docker
        dockerfile: Dockerfile.viewer
    state: present
    
- name: Create screenly-viewer container
  docker_container:
    name: screenly-ose-viewer
    image: "screenly-ose-viewer"
    state: started
    recreate: true
    restart_policy: unless-stopped
    network_mode: host
    privileged: yes
    log_driver: journald
    volumes:
      - /home/pi/screenly:/home/pi/screenly
      - /home/pi/.screenly:/home/pi/.screenly
      - /home/pi/screenly_assets:/home/pi/screenly_assets
      - /etc/timezone:/etc/timezone:ro
      - /etc/localtime:/etc/localtime:ro

```

ealmonte32:
@Kipjr 

I think almost all users that are doing any local development and troubleshooting simply edit the install script to point to their github repo, and to their local folder where their custom ansible tasks and files are.
Adding these extra questions will probably confuse some users when running the install script and add lines of code unnecessarily.
Also, when a user needs to run the script after already having installed screenly for whatever reason, that screenly directory already exists, and this is confusing, if the directory exists, then what is your local variable doing differently if it will still call the install.sh script and in there you are calling the official git repo information anyway?

I think the focus you have on the dockerfile task is a better direction of resources since we want to dockerize the different services (screenly-viewer, screenly-web, screenly-websocket_server_layer, screenly-celery) taking it one step at a time.

Kipjr:
I can understand. However, the threshold to contribute this project is rather high due to the complexity of each connection. Therefore I wanted to decrease this by keeping everything local.

vpetersson:
@Kipjr Thank you for the effort here. I recognize the bar is high to contribute, but I'm with @ealmonte32 -- this will likely cause a lot of confusion for non-developers. 

The recommended approach for development is to maintain a fork. That way you simply push the changes to your fork, and then re-run the installer with the `REPOSITORY` variable pointing to your fork. The reason for this is that you are unlikely to work directly on the Raspberry Pi in the first place, but rather use a local IDE.

Kipjr:
Thanks I will keep this in mind and will change my approach

PR #1386, opened 2020-10-18
Update install.sh

ealmonte32:
@vpetersson 

created another PR so that it is clearly showing the changes since the previous PR was very old and changes made were already accumulating on top of commits made to master install.

fixes/changes:
- web upgrade was always installing network manager, even when users would not select this
- remove docker packages if the installation is not experimental
- install `whois` package in order to use the password checker function
- some installations would fail if ansible dir contents were not pulled down and replaced from official repo
- update version of ansible to match the one in requirements.txt
- directory `/usr/share/info` should not be deleted or else some errors are thrown when apt is used to remove/uninstall some packages, therefore deleting the contents inside of `/info/*` should suffice
- alias file for locale should not be deleted or else error is thrown when changing some timezone data, which will be used in future with web gui way of changing the timezone
- added default raspberry password checker and warning if found

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -34,8 +34,12 @@ if [ "$WEB_UPGRADE" = false ]; then
     exit 1
   fi
 
+  # clear screen
+  clear;
+  
   # Set color of logo
-  tput setaf 4
+  tput setaf 6
+  tput bold
 
   cat << EOF
        _____                           __         ____  _____ ______
@@ -97,9 +101,9 @@ elif [ "$WEB_UPGRADE" = true ]; then
   fi
 
   if [ "$MANAGE_NETWORK" = false ]; then
-    NETWORK="y"
-  elif [ "$MANAGE_NETWORK" = true ]; then
     NETWORK="n"
+  elif [ "$MANAGE_NETWORK" = true ]; then
+    NETWORK="y"
   else
     echo -e "Invalid -n parameter."
     exit 1
@@ -149,8 +153,13 @@ fi
 sudo sed -i 's/apt.screenlyapp.com/archive.raspbian.org/g' /etc/apt/sources.list
 sudo apt update -y
 sudo apt-get purge -y python-setuptools python-pip python-pyasn1
-sudo apt-get install -y python-dev git-core libffi-dev libssl-dev
+sudo apt-get install -y python-dev git-core libffi-dev libssl-dev whois
 curl -s https://bootstrap.pypa.io/get-pip.py | sudo python
+
+# users who chose experimental and then reverted back to master or production need docker removed
+if [ "$BRANCH" != "experimental" ]; then
+	sudo apt-get purge -y docker-ce docker-ce-cli containerd.io > /dev/null
+fi
 
 if [ "$NETWORK" == 'y' ]; then
   export MANAGE_NETWORK=true
@@ -159,9 +168,9 @@ else
   export MANAGE_NETWORK=false
 fi
 
-sudo pip install ansible==2.8.2
+sudo pip install ansible==2.8.8
 
-sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH"
+sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH force=yes"
 cd /home/pi/screenly/ansible
 
 sudo -E ansible-playbook site.yml $EXTRA_ARGS
@@ -170,9 +179,9 @@ sudo apt-get autoclean
 sudo apt-get clean
 sudo find /usr/share/doc -depth -type f ! -name copyright -delete
 sudo find /usr/share/doc -empty -delete
-sudo rm -rf /usr/share/man /usr/share/groff /usr/share/info /usr/share/lintian /usr/share/linda /var/cache/man
+sudo rm -rf /usr/share/man /usr/share/groff /usr/share/info/* /usr/share/lintian /usr/share/linda /var/cache/man
 sudo find /usr/share/locale -type f ! -name 'en' ! -name 'de*' ! -name 'es*' ! -name 'ja*' ! -name 'fr*' ! -name 'zh*' -delete
-sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*' ! -name 'es*' ! -name 'ja*' ! -name 'fr*' ! -name 'zh*' -exec rm -r {} \;
+sudo find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en*' ! -name 'de*' ! -name 'es*' ! -name 'ja*' ! -name 'fr*' ! -name 'zh*' ! -name 'locale.alias' -exec rm -r {} \;
 
 cd /home/pi/screenly && git rev-parse HEAD > /home/pi/.screenly/latest_screenly_sha
 sudo chown -R pi:pi /home/pi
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -149,19 +153,24 @@ fi
...
   sudo apt-get install -y network-manager
 else
   export MANAGE_NETWORK=false
 fi
 
-sudo pip install ansible==2.8.2
+sudo pip install ansible==2.8.8
 
-sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH"
+sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH force=yes"
```
vpetersson, line 173: Why do we nee `force=yes` here?

vpetersson:
Thanks @ealmonte32. Looks great. Merging.

ealmonte32:
@vpetersson 
With regards to your why `force=yes` question:

Basically, when ansible finds any local modification in the destination, in this case  `/home/pi/screenly/ansible` , the whole ansible installation fails with the error shown below because it defaults to not overwriting. Most of the users that sometimes need to reinstall and they run the script again need this to be forced because it makes sure that it pulls all latest and official files from the github repo and that way everything that needs to be cloned/copied into `/home/pi/screenly/` is done.

```
+ echo 'INSTALLING WITHOUT GIT FORCE = YES'
INSTALLING WITHOUT GIT FORCE = YES
+ sudo -u pi ansible localhost -m git -a 'repo=https://github.com/ealmonte32/screenly-ose.git dest=/home/pi/screenly version=master'
[WARNING]: Platform linux on host localhost is using the discovered Python interpreter at /usr/bin/python, but future installation of another Python
interpreter could change this. See https://docs.ansible.com/ansible/2.8/reference_appendices/interpreter_discovery.html for more information.

localhost | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python"
    }, 
    "before": "f756d31a92eff19e2b7b21a569e66f0987fc21b9", 
    "changed": false, 
    "msg": "Local modifications exist in repository (force=no)."
}
pi@rpi3blk:~ $ 
```

The above failed simply because I changed one character inside a ansible playbook file.

vpetersson:
Interesting point. I wonder if we perhaps want to show some sort of error instead. There are lots of potential issues with forcing, such as merge-conflicts etc.

ealmonte32:
Hmm.. but the reason we force is so that the installation succeeds because if you are installing screenly-ose via the install script then it needs the official files/folders from the screenly-ose repo, so if there are any merging conflicts they would be between a user locally troubleshooting and his/her own repo and not merge conflict with official screenly-ose repo.
I thought the most valid point for this is specifically because it forces downstream the official files, if you happen to change anything locally for any troubleshooting or local development reasons, you should be downloading your own repo, not the official screenly-ose one.

P.S. could you give me an example of one scenario where forcing the official screenly-ose github repo would be an issue? I tried to think of some but it always gets me back to having the user change to their repo if local modifications is what they need.

vpetersson:
Sure, here's the scenario that I'm thinking of where I'm fairly confident that this would break:

* The user is either using the upstream repo or his/her own fork (doesn't really matter for the sake of the example)
* Some lines are changed *differently* in the upstream repo and the local clone. Doesn't really matter if it is a Python file or Ansible file, as long as there are contradicting changes which would result in a merge conflict. This might sound implausible, but I think it is more likely than it sounds like, as there will be 'hot spots' in the code base.
* If this merge conflict happens, which would lead to the file in question looking like this on disk:

```
<<<<<<< HEAD
this is some content to mess with
content to append
=======
totally different content to merge later
>>>>>>> new_branch_to_merge_later
```

* If we ignore this and just push along with a force, something will break. If it's a merge conflict in Ansible, this will become evident quickly, but if it is say in a file that is only periodically called on by the viewer or server, it will be a lot less obvious. 

That's why I'd opt for a [fail fast](https://en.wikipedia.org/wiki/Fail-fast) approach here rather than quietly suppressing it with Force.

ealmonte32:
Hmm.. so if I understand correctly you're saying a developer or power user that has made changes to any of the files he/she is working on locally and also remotely but forgot to push local changes to remote is the issue?

Yes I understand but I checked ansible documentation and the git module does not currently support any merge feature where any local changes will merge depending on which files have latest timestamp 😞  .
I tried the `update=yes` flag but it still failed because this is not a "merge" type of setting where it does the previously mentioned.

If you want, I can make the script check for the `/home/pi/screenly` directory early on during install first and if it exists I can prompt user to select if they wish to back this directory up because if they have made any modifications the installer will replace all files inside of it..
This way, we take care of the dev/power users who might have forgotten to push their changes and also take care of installations never failing because of local changes. 

Thoughts? I like this backup question route actually.. the cloned size of screenly directory (without ngrok binary) is only `6.4 MB` , but with ngrok binary is `29.6 MB` , so it's not like it will be taking up space on the SD card..

vpetersson:
> Hmm.. so if I understand correctly you're saying a developer or power user that has made changes to any of the files he/she is working on locally and also remotely but forgot to push local changes to remote is the issue?

No, i think it can be more generic than that. Say for instance that you're using the upstream (official) branch, and you've then made local changes. I think that is a more realistic scenario. Imagine for instance that the local version is severely out of date and has some local changes and you're trying to roll forward a year or two worth of changes. There is then a good probability that you'll run into a merge conflict. 

> Yes I understand but I checked ansible documentation and the git module does not currently support any merge feature where any local changes will merge depending on which files have latest timestamp 😞 .
I tried the update=yes flag but it still failed because this is not a "merge" type of setting where it does the previously mentioned.

This is not very surprising to be honest. `git merge`s can be complex to do even by hand.

> If you want, I can make the script check for the /home/pi/screenly directory early on during install first and if it exists I can prompt user to select if they wish to back this directory up because if they have made any modifications the installer will replace all files inside of it..
This way, we take care of the dev/power users who might have forgotten to push their changes and also take care of installations never failing because of local changes.

Hmm, it's better to leave it with `force=False` and then add a conditional error message? Perhaps something like 'Local changes were found. For more information see https....'?

I'm not sure that's possible with Ansible though.

ealmonte32:
> Hmm, it's better to leave it with `force=False` and then add a conditional error message? Perhaps something like 'Local changes were found. For more information see https....'?

Yeah I tried to find a way to create a custom fail message, but this step during ansible facts is pre-playbook so only way to do it is modifying the actual source_control python file, which we can then say whatever we want:

Screen Shot 2020-10-23 at 2 40 48 PM

We can insert the "For more information see `https://www.screenly...` and in there explain the whole error msg and what it actualyl means... or maybe I should create a PR in ansible asking them to least send users to an actual documentation as to what exactly this means for those users who are new to ansible and just need a simple explanation.. the warning they give is talking about interpreter_discovery and not about the local modifications, so thats confusing when the error deals with modifications of local files and nothing to do with python.. I cant even find in their own documentation a concise cleanly explained reason for the local modification error, I mean yes most developers know what it means after googling it, because it sends them to github issues, forums, and many PR's where people explain what it means, but the fact that ansible doesnt simply explain in layman terms that this error simply means there were files in the destination that exist and user needs to either discard them or merge them with their repo before proceeding.

I am thinking on working on a custom merge type function where if screenly detects the default installation directory of `/home/pi/screenly`, it copies it to `/home/pi/screenly_local` and after ansible downloads the repo, it uses if conditions and `diff` to merge both together using `patch` by asking user if they want to patch the downloaded repo with the `screenly_local` files or to simply replace all local files in screenly and discarding screenly_local...
It's a lot of work and requires lots of testing but i think the idea can be a feature at least.

ealmonte32:
@vpetersson 
Can you comment here about this -> https://github.com/ansible/ansible/pull/72405
Ansible would accept the change for the git module to be more detailed when the local repo and destination repo conflict, but want the actual name of the destination to be shown to user and not just wording, which makes sense.

I made the change and it is now letting users know the conflicting directory, is there anything else you would change?
```
pi@transparentpi:~ $ sudo -u pi ansible localhost -m git -a "repo=https://github.com/screenly/screenly-ose.git dest=/home/pi/screenly version=$BRANCH force=no"

    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python"
    }, 
    "before": "1079dbdeabf86b2804c66a790f1d80245671eea0", 
    "changed": false, 
    "msg": "Local modifications exist in destination: /home/pi/screenly (force=no)"
}
pi@transparentpi:~ $ 
```

Changes to `git.py` module local_mods function:
```
if local_mods:
            # failure should happen regardless of check mode
            if not force:
                module.fail_json(msg="Local modifications exist in destination: " + dest + " (force=no)", **result)
            # if force and in non-check mode, do a reset
            if not module.check_mode:
                reset(git_path, module, dest)
                result.update(changed=True, msg='Local modifications exist in destination: ' + dest)
```

vpetersson:
Perfect. I've thumbs up'd your PR upstream. Let's wait for it to get merged (even if i know it can take a while). But yes, good idea and i think it makes sense.

PR #1387, opened 2020-10-18
Update server.py

ealmonte32:
fixes:
- if escape was used in `uri`, the characters it was escaping to were preventing certain websites such as google slides from parsing parameters necessary for proper function, thus a manual replace had to be implemented to accomplish both needs (escaping while allowing `&` to be used properly)
- when escaping the name of the asset, if you make asset with `&` character inactive and active multiple times, the name starts having many duplicate instances of the word `amp;` (see screenshot)

ampfix-needed-for-name

### DIFF
--- a/server.py
+++ b/server.py
@@ -416,9 +416,7 @@ def prepare_asset(request, unique_name=False):
     if not all([get('name'), get('uri'), get('mimetype')]):
         raise Exception("Not enough information provided. Please specify 'name', 'uri', and 'mimetype'.")
 
-    ampfix = "&amp;"
-    name = escape(get('name').replace(ampfix, '&'))
-    
+    name = escape(get('name'))
     if unique_name:
         with db.conn(settings['database']) as conn:
             names = assets_helper.get_names_of_assets(conn)
@@ -441,7 +439,7 @@ def prepare_asset(request, unique_name=False):
         'nocache': get('nocache'),
     }
 
-    uri = (get('uri')).replace(ampfix, '&').replace('<', '&lt;').replace('>', '&gt;').replace('\'', '&apos;').replace('\"', '&quot;')
+    uri = escape(get('uri').encode('utf-8'))
 
     if uri.startswith('/'):
         if not path.isfile(uri):
@@ -507,7 +505,8 @@ def prepare_asset_v1_2(request_environ, asset_id=None, unique_name=False):
         raise Exception(
             "Not enough information provided. Please specify 'name', 'uri', 'mimetype', 'is_enabled', 'start_date' and 'end_date'.")
 
-    name = escape(get('name'))
+    ampfix = "&amp;"
+    name = escape(get('name').replace(ampfix, '&'))
     if unique_name:
         with db.conn(settings['database']) as conn:
             names = assets_helper.get_names_of_assets(conn)
@@ -528,7 +527,7 @@ def prepare_asset_v1_2(request_environ, asset_id=None, unique_name=False):
         'nocache': get('nocache')
     }
 
-    uri = escape(get('uri'))
+    uri = (get('uri')).replace(ampfix, '&').replace('<', '&lt;').replace('>', '&gt;').replace('\'', '&apos;').replace('\"', '&quot;')
 
     if uri.startswith('/'):
         if not path.isfile(uri):
--- a/server.py
+++ b/server.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 __author__ = "Screenly, Inc"
-__copyright__ = "Copyright 2012-2019, Screenly, Inc"
+__copyright__ = "Copyright 2012-2020, Screenly, Inc"
 __license__ = "Dual License: GPLv2 and Commercial License"
 
 import json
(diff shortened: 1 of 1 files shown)


vpetersson:
@ealmonte32 What happens if you use characters like ö in the URL?

ealmonte32:
> @ealmonte32 What happens if you use characters like ö in the URL?

we get the unicoderror:

```
Oct 19 15:03:40 rpi3blk python[1077]: UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 15: ordinal not in range(128)
```

I also edited the wrong line in the server.py, it should have been in under `def prepare_asset_v1_2`.
I will work on figuring out the unicode error, i remember the workaround of forcing the system encoding to utf-8 is not recommended so we have to probably decode first and then encode last for non-ascii characters.

ealmonte32:
Just wanted to remind that this PR is specific for `&` and `amp;` fix and that the fix for universal URI / unicode support should be a different PR that specifically focuses on that because users need the `&` symbol for websites that cannot have it escaped and changed some `&amp;` , thus even without this PR the issue with unicode in server.py and asset_helper.py and sqlite and string vs bytes etc would still be present, so at least with this PR we take care of one fix at least.

ealmonte32:
Sorry I should have written this here instead of https://github.com/Screenly/screenly-ose/pull/1394#issuecomment-718495100

@vpetersson 
Is there a reason why we can't just manually set the site customize to utf-8, python3 is hard-coded for utf-8, and we know we want utf-8, so not sure why we can't just go this route:

file: `/usr/lib/python2.7/sitecustomize.py`
```
import sys
sys.setdefaultencoding('utf-8')
```

by running this on install script somewhere after we know python is installed or after ansible tasks finish:
```
echo -e "\n#setting utf-8 as default python2 encoding\nimport sys\nsys.setdefaultencoding('utf-8')\n" | sudo tee -a /usr/lib/python2.7/sitecustomize.py
```
and pretty much solve this uri international issue once and for all.. remember this way we  dont need to run the sys reload hack in every python file, this is like if the python env was running as python3..

I know you usually have your reasons for not doing certain changes, can you give me one of your examples for this situation?
Thanks.

vpetersson:
Good point about UTF8. Yeah, I know python27 is a PITA when it comes to encoding. I wasn't sure Python 3 did this out-of-the box.

That said, if you look at [this StackOverflow post](https://stackoverflow.com/questions/33274428/python-change-character-encoding-to-utf-8), switching to UTF-8 as the default encoding appears to be discouraged. When we move to Python3 (and Django), this issue will mostly go away. 

In any case, let's merge this in and deal with the wider unicode issue later.

ealmonte32:
> That said, if you look at [this StackOverflow post](https://stackoverflow.com/questions/33274428/python-change-character-encoding-to-utf-8), switching to UTF-8 as the default encoding appears to be discouraged. When we move to Python3 (and Django), this issue will mostly go away.
> 
> In any case, let's merge this in and deal with the wider unicode issue later.

But that post and almost all the posts I've read are talking about the reloading of the file with `sys` after it has already been loaded which is discouraged because you are changing on the fly the type of encoding that the system itself is running on which picks up the python env default encoding, which is ASCII for Python2.7 but this is exactly why Python3 defaulted to UTF-8, also remember all our locales codesets are UTF-8 as well.
What we are doing here is simply saying we want Screenly based RPi's to use the universal standard for encoding of UTF-8 on all python2.7 programs running which will be backwards compatible with ASCII as well, and even the SQLite default encoding is UTF-8 as shown [here](https://www.sqlite.org/search?q=encoding&i=1)

Even HTML [standards](https://html.spec.whatwg.org/multipage/semantics.html#charset) says we must default or use utf-8 as the encoding.

```
The UTF-8 encoding is the most appropriate encoding for interchange of Unicode, the universal coded character set.
Therefore for new protocols and formats, as well as existing formats deployed in new contexts, this specification requires (and defines) the UTF-8 encoding.
```

Using the `sitecustomize.py` method because we currently use Python2.7 seems to be the best and easiest way to solve this. ASCII encoding is not encouraged for web applications that take in unknown or international characters because of the limits it has understanding those, which is why we raise the UnicodeError because you are adding a URI that has characters beyond what ASCII understands, but we know we want UTF-8 since it is backwards compatible and extends ASCII.
Let me know if you want the PR to do this.

References:
https://html.spec.whatwg.org/multipage/semantics.html#charset
https://encoding.spec.whatwg.org/#preface
https://medium.com/@apiltamang/unicode-utf-8-and-ascii-encodings-made-easy-5bfbe3a1c45a
https://docs.python.org/3/howto/unicode.html
https://en.wikipedia.org/wiki/UTF-8
https://www.sqlite.org/search?q=encoding&i=1
https://docs.sdl.com/791187/311575/sdl-contenta-5-7/iso-8859-1-v-s--utf-8
https://docs.oracle.com/cd/E23824_01/html/E26033/glmbx.html

ealmonte32:
Just wanted to post the screenshots of how it correctly handles the URLs:

Screen Shot 2020-10-29 at 2 30 14 PM
___
Screen Shot 2020-10-29 at 2 30 25 PM
___
Screen Shot 2020-10-29 at 2 30 35 PM

PR #1388, opened 2020-10-19
Update main.yml

ealmonte32:
just found a package part of plymouth that is of no use and extra

### DIFF
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -175,6 +175,7 @@
       - lightdm-gtk-greeter
       - dphys-swapfile
       - rabbitmq-server
+      - pix-plym-splash
     state: absent
     
 - name: Remove cups/printing related packages


ealmonte32:
Just wanted to add this comment here for reference if future questions come up about this decision.

Basically, after looking for packages that were optional or not used, and as we want to slim down the installation/system footprint, every little bit counts, so researching led me to this comment:
https://github.com/RPi-Distro/raspberrypi-ui-mods/issues/10#issuecomment-273300264

which led me to see which theme was being selected after installation of screenly:
```
pi@rpi3blk:~ $ cat /usr/share/plymouth/plymouthd.default
# Distribution defaults. Changes to this file will get overwritten during
# upgrades.
[Daemon]
Theme=screenly
ShowDelay=0
```

This proves that pix theme is not used therefore we dont need the pix-plym-splash package.

vpetersson:
Yep, good point. We ship our own theme instead.

PR #1390, opened 2020-10-21
Update README.md

ealmonte32:
addition of developer documentation wiki and link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,11 @@ Should you want to upgrade to the latest development version (for instance if yo
 
 To learn more about Screenly, please visit the official website at [Screenly.io](http://www.screenly.io).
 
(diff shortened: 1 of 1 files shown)


ealmonte32:
Should the Dev Docs just go under the Quick Links bullet items instead?

Also, I was wondering if there is a way to lock down the wiki page, if not (dont think there is but just making sure), then I guess we can instead move the markdown file into the repo under `docs` simply as `developer_documentation.md` and that way it stays within the repo and any edits are made as PR requests instead of the "freedom to edit as you choose" wiki type.

vpetersson:
Good catch @ealmonte32! I had no idea it was publicly writable. I've turned off the Wiki and moved everything that was in there to the docs folder and also updated the index. Rejecting this PR because of that.

PR #1394, opened 2020-10-21
minor deprecation fix on Jessie block - experimental branch

ealmonte32:
minor fix for deprecation warnings:

```

TASK [splashscreen : Installs dependencies (Jessie)] ***********************************************************************************************************************
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name:
 "{{ item }}"`, please use `name: ['fbi']` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [localhost] => (item=[]) 



TASK [splashscreen : Installs dependencies (not Jessie)] *******************************************************************************************************************
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name:
 "{{ item }}"`, please use `name: ['plymouth']` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
ok: [localhost] => (item=[u'plymouth'])
```

### DIFF
--- a/ansible/roles/splashscreen/tasks/main.yml
+++ b/ansible/roles/splashscreen/tasks/main.yml
@@ -3,9 +3,7 @@
 # If Jessie
 - name: Installs dependencies (Jessie)
   apt:
-    name: "{{ item }}"
-  with_items:
-    - fbi
+    name: fbi
   when: ansible_distribution_major_version|int <= 7
 
 - name: Copies in splash screen
@@ -44,9 +42,7 @@
 
 - name: Installs dependencies (not Jessie)
   apt:
-    name: "{{ item }}"
-  with_items:
-    - plymouth
+    name: plymouth
   when: ansible_distribution_major_version|int > 7
 
 - name: Copies plymouth theme


vpetersson:
Looks good, but could you clean up this file a bit while at it? I guess it's Pre and post Jessie really.

ealmonte32:
> Looks good, but could you clean up this file a bit while at it? I guess it's Pre and post Jessie really.

So remove tasks that were pertinent to jessie only correct?

vpetersson:
No, i would leave them in just for backward compatibility. They should be automatically skipped anyways. I just mean that that we should probably rename the post-Jessie block

ealmonte32:
> No, i would leave them in just for backward compatibility. They should be automatically skipped anyways. I just mean that that we should probably rename the post-Jessie block

Oh, i usually never even change the topic of any edited PR, I simply edit the contents of the file and just commit, but yes I should get used to changing the topic to better reflect what is happening..

ealmonte32:
Done.

Kipjr:
continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error:
_UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 13: ordinal not in range(128)_

Forcing the Locale on en_US.UTF-8 might solve this. 

```
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
export LANGUAGE=en_US.UTF-8
```

vpetersson:
> continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error:
> _UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 13: ordinal not in range(128)_
> 
> Forcing the Locale on en_US.UTF-8 might solve this.
> 
> ```
> export LC_ALL=en_US.UTF-8
> export LANG=en_US.UTF-8
> export LANGUAGE=en_US.UTF-8
> ```

I'm working on migrating to GitHub Actions instead here: 
https://github.com/Screenly/screenly-ose/pull/1397

ealmonte32:
> continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error:
> _UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 13: ordinal not in range(128)_
> 
> Forcing the Locale on en_US.UTF-8 might solve this.
> 
> ```
> export LC_ALL=en_US.UTF-8
> export LANG=en_US.UTF-8
> export LANGUAGE=en_US.UTF-8
> ```

@Kipjr 
This does not work. These do not affect the python program once it is ran, you can use the reload(sys) workaround but it is highly unrecommended, read about why here:
https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script

I've spent countless hours trying to troubleshoot this, even trying to install international language packs on the system, importing codecs, unicode conversion, etc, and have currently ran into the `./screenly/assets_helper.py` file as current part of the workflow that is giving me an error after using `.encode('utf-8')` on the uri creation line in `server.py` under `prepare_asset_v1_2` function.

As far as I know, I think we need to decode the URL and then encode the URL to utf8 when it is added to assets, but then as well when it goes into the sqlite database on create function, but problem is asset helper gives the following error:
```
Server Error: You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings.
```

It also gives error about expecting bytes or buffer, I've been trying to research this to learn how to properly workaround it or properly fix it, but have yet to figure it out. I know that the reason why we get unicode error is because the program tries to decode the URL entered with `ascii` and since the characters are beyond the `ascii` symbols table, it gives the error.
This is why we have to somehow, decode the URL entered, and encode it back as utf-8, but then we make sure the sqlite database accepts this encoding as well.
If I am completely wrong in how I'm thinking about this, pls let me know.

ealmonte32:
@Kipjr , I wanted to show you what I meant with regards to encoding a file after it is running as opposed to before it runs or during it running... i think we might have to set a systemwide python2 environment encoding for this to work, or need to do something with `u''.join(uri)` in the asset helper file.... I am leaning more towards systemwide utf-8 changing..

Anyway, so look let me show you what I mean... you see that the`./screenly/server.py` file has utf-8 coding on the top, but look at what the python env defaults to when a python2 file runs..

Python2 file contents:
```
pi@transparentpi:~ $ cat ./testencoding.py
#!/usr/bin/python
# -*- coding: utf-8 -*-

import sys
import os

# we assign a URL that contains international characters/symbols:
url = "http://www.пример.com"

print "This is the international URL: " + url

url_type = type(url)
print "Now we check the type: " + str(url_type)

check_encoding = sys.getdefaultencoding()
print "We now check the current encoding this file is running: " + check_encoding
```

This is what that python2 file shows when ran:
```
pi@transparentpi:~ $ python ./testencoding.py 
This is the international URL: http://www.пример.com
Now we check the type: <type 'str'>
We now check the current encoding this file is running: ascii
```

As you can see, the encoding environment is `ascii`

..

Now, we run the same file with Python3 and modified a bit to comfort to py3:
```
pi@transparentpi:~ $ cat ./testencoding_py3.py 
#!/usr/bin/python3.7
# -*- coding: utf-8 -*-

import sys
import os

# we assign a URL that contains international characters/symbols:
url = 'http://www.пример.com'

print ('This is the international URL: ', url)

url_type = type(url)
print ('Now we check the type: ', url_type)

check_encoding = sys.getdefaultencoding()
print ('We now check the current encoding this file is running: ', check_encoding)
```

This is what running that file results in:
```
pi@transparentpi:~ $ python3.7 ./testencoding_py3.py 
This is the international URL:  http://www.пример.com
Now we check the type:  <class 'str'>
We now check the current encoding this file is running:  utf-8
```

As you can see, this one does in fact run with the utf-8 encoding as default environment.

ealmonte32:
@vpetersson 
Is there a reason why we can't just manually set the site customize to utf-8, python3 is hard-coded for utf-8, and we know we want utf-8, so not sure why we can't just go this route:

file: `/usr/lib/python2.7/sitecustomize.py`
```
import sys
sys.setdefaultencoding('utf-8')
```

by running this on install script somewhere after we know python is installed or after ansible tasks finish:
```
echo -e "\n#setting utf-8 as default python2 encoding\nimport sys\nsys.setdefaultencoding('utf-8')\n" | sudo tee -a /usr/lib/python2.7/sitecustomize.py
```
and pretty much solve this uri international issue once and for all.. remember this way we  dont need to run the sys reload hack in every python file, this is like if the python env was running as python3..

I know you usually have your reasons for not doing certain changes, can you give me one of your examples for this situation?
Thanks.

PR #1395, opened 2020-10-21
Update install.sh

ealmonte32:
update installer to clearly show different branch versions and make appropriate selection use case statements..

what do you think?


vpetersson:
Yeah that looks sensible. I want to overhaul the installer in a not too distant future to use `whiptail` anyways.

PR #1401, opened 2020-10-29
update /lib/diagnostics.py to include latest raspberry pi models

ealmonte32:
ran into this on the new rpi4:
```
Raspberry Model | Unable to determine raspberry model. Revision: Unable to determine raspberry revision. Ram: Unable to determine raspberry RAM. Unable to determine raspberry manufacturer.
```

updated to include all latest models and info.


vpetersson:
Nice fix @ealmonte32! Yeah this file was pretty far out of date.

PR #1404, opened 2020-11-03
cleanup and minor changes on password checking block

ealmonte32:
Minor cleanup on password checking block and experimental branch group revert
-made password checking into function
-use local variables instead of global
-reverting back git flag`force=no` until we find better solution for scenarios with local merge conflicts

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -175,7 +175,7 @@ fi
 
 sudo pip install ansible==2.8.8
 
-sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH force=yes"
+sudo -u pi ansible localhost -m git -a "repo=$REPOSITORY dest=/home/pi/screenly version=$BRANCH force=no"
 cd /home/pi/screenly/ansible
 
 sudo -E ansible-playbook site.yml $EXTRA_ARGS
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -160,9 +160,10 @@ sudo apt-get purge -y python-setuptools python-pip python-pyasn1
 sudo apt-get install -y python-dev git-core libffi-dev libssl-dev whois
 curl -s https://bootstrap.pypa.io/get-pip.py | sudo python
 
-# users who chose experimental and then reverted back to master or production need docker removed
+# users who chose experimental then reverted back to master or production need docker pkg removed and user group reverted
 if [ "$BRANCH" != "experimental" ]; then
-	sudo apt-get purge -y docker-ce docker-ce-cli containerd.io > /dev/null
+  sudo apt-get purge -y docker-ce docker-ce-cli containerd.io > /dev/null
+  sudo usermod -g pi pi
 fi
 
 if [ "$NETWORK" == 'y' ]; then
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -160,9 +160,10 @@ sudo apt-get purge -y python-setuptools python-pip python-pyasn1
 sudo apt-get install -y python-dev git-core libffi-dev libssl-dev whois
 curl -s https://bootstrap.pypa.io/get-pip.py | sudo python
 
-# users who chose experimental and then reverted back to master or production need docker removed
+# users who chose experimental then reverted back to master or production need docker pkg removed and user group reverted
 if [ "$BRANCH" != "experimental" ]; then
-	sudo apt-get purge -y docker-ce docker-ce-cli containerd.io > /dev/null
+  sudo apt-get purge -y docker-ce docker-ce-cli containerd.io > /dev/null
+  sudo usermod -g pi pi
```
vpetersson, line 166: @ealmonte32 why do we need this? Shouldn't this always be true? IIRC, we only *add* the group `docker` in the experimental branch. We don't change the main group ownership.

ealmonte32: Hmm.. well I was testing by installing master or production branch, then switching to experimental as a way to test something as we always suggest, then after I finish testing I go and switch back to production or master by running the install script, and when I did that, every file I created in the pi user home directory had the group `docker` as the primary one.. so not sure in which part of the configuration under experimental this is being done but it doesn't seem like it is just adding the group.. do you suggest I re-examine this process and see if it is only adding the primary group to a directory ? because even if that is the case, if the branch is not experimental, the pi user should not have directories where docker is the primary group.

vpetersson: Ahh. Good catch. I found the issue and pushed a fix [here](https://github.com/Screenly/screenly-ose/commit/8d3aa71dced59e4f67f0651b55ee0479c12af310).

ealmonte32: This github search is so annoying sometimes 😠 , the first thing I did when trying to find where this happened was to do the usual thing of searching the repo, but nothing was returned in the results with the mention of docker in anything dealing with group, look at screenshots even taken right now still can't find the word `docker` inside the yml file.. it's as if it only searches certain levels of directories.. that is so annoying that means all this time I should have not relied on that search and instead used `grep` 🤦🏻‍♂️  

I'll remove that part of code.. thx

Screen Shot 2020-11-05 at 4 42 10 PM
Screen Shot 2020-11-05 at 4 42 59 PM

vpetersson: :)

Both me and @rusko124 are actively working on the Experimental branch and getting it ready to merge into master. There were a large number of merge conflicts etc so @rusko124 did a great job cleaning that up. I will ping you once ready as it will require a fair bit of testing.

ealmonte32: sounds good

ealmonte32:
So should I remove the lines that uninstall docker, etc, so that we can merge the other changes to the install script?

vpetersson:
Yeah sure. I've spent a lot of cycles on refactoring this week (https://github.com/Screenly/screenly-ose/pull/1409) and hope to be in good enough state to merge it into the experimental branch later today.

ealmonte32:
Done and ready to merge.

Wow those are a lot of changes you've made for experimental.. 🤞🏻

vpetersson:
> Wow those are a lot of changes you've made for experimental.. 

Yeah @rusko124 did the bulk of the work. I'm just putting the finishing touches on it. :)

PR #1414, opened 2020-11-24
Task/screenly conf file

rusko124:
After migration from master to experimental, server and viewer were failed because they misidentified the paths from `screenly.conf`


ansible/roles/screenly/tasks/main.yml
```diff
@@ -9,6 +9,14 @@
     - .config
     - screenly_assets
 
+- name: Copy Screenly default config
```
vpetersson, line 12: We can remove this block entirely. It's done [here](https://github.com/Screenly/screenly-ose/blob/experimental/bin/start_server.sh#L8-L10).

rusko124:
@vpetersson updated

vpetersson:
You shouldn't need this. I moved this to [here](https://github.com/Screenly/screenly-ose/blob/experimental/bin/start_server.sh).

vpetersson:
@rusko124 is this still relevant?

rusko124:
@vpetersson No, we can close it, but we need to add a similar code to a migrate script.

PR #1420, opened 2020-11-26
Add balena.yml for balenaHub

shawaj:
Further to #1419 for the master branch (only difference is Pi4 support removed)


vpetersson:
@shawaj This PR is against the wrong branch

PR #1422, opened 2020-11-27
Edits: update for viewer-builder with qt5.15

rusko124:
Tested on examples from https://github.com/qt/qtbase/tree/dev/examples


vpetersson:
@rusko124 I'm closing out this one in favor of https://github.com/Screenly/screenly-ose/issues/1426. I've tested the added dependencies in here on my node but it didn't make a difference.

PR #1444, opened 2020-12-16
Update server.py

ealmonte32:
fixes API v1.2 missing `asset_id` parameter like used in API v1 which would allow custom asset_id to be used when using API and POST requests..

@rusko124 if you have any objection to this pls let me know why, I followed your prepare_asset function for `if not asset_id` from version 1, I am trying to create some custom dashboard control and scripts that allows me to enable/disable all assets by running through loop that I customized with asset_id's named like:
`male_ads_randomtext1`
`male_ads_randomtext2`
`female_ads_randomtext1`
`female_ads_randomtext2`

Which will allow a developer that is creating some API calls to use the wildcard or preface regex such as `male_ads_` for example just like you use "default_" for default assets YAML file, and so this allows a customized API call that can enable or disable all tagged assets based on the regex/wildcard that the developer creates themselves..
Again, this is just allowing the asset_id to be customized upon a POST API call/request. When the asset_id is not given on the --data JSON request, it automatically creates it, and if it already exists it gives you error from SQL constraint, so I think all bases are covered.

Thoughts against this?

Thanks.

---
Just a screenshot of what I mean about building a custom dashboard:
Screen Shot 2020-12-16 at 3 22 05 AM

### DIFF
--- a/server.py
+++ b/server.py
@@ -436,7 +436,7 @@ def prepare_asset(request, unique_name=False):
         'asset_id': get('asset_id'),
         'is_enabled': get('is_enabled'),
         'is_processing': get('is_processing'),
-        'nocache': get('nocache'),
+        'nocache': get('nocache')
     }
 
     uri = escape(get('uri').encode('utf-8'))
@@ -523,6 +523,7 @@ def prepare_asset_v1_2(request_environ, asset_id=None, unique_name=False):
     asset = {
         'name': name,
         'mimetype': get('mimetype'),
+        'asset_id': get('asset_id'),
         'is_enabled': get('is_enabled'),
         'nocache': get('nocache')
     }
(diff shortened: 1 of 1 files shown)


vpetersson:
@ealmonte32 interesting. Haven't looked at the code in detail yet, but basically, you need to bump up the version. APIs should be frozen within a given version and never change. This is to not break capability. Hence, we need to bump up the version to by X.1 to add this.

ealmonte32:
> @ealmonte32 interesting. Haven't looked at the code in detail yet, but basically, you need to bump up the version. APIs should be frozen within a given version and never change. This is to not break capability. Hence, we need to bump up the version to by X.1 to add this.

Ahh.. I see, that makes sense.. although, I don't think I made any real changes to the actual API code itself, unless the function to prepare the asset is the API entirely, because the API call still works, the adding of assets through HTTP POST and dashboard/GUI still works, I tested it every way and it works (unless there is something very very specific about the lines I added/modified), meaning the line change is sort of like extending the v1.2 API call to create the asset with a combination of simply allowing for the asset_id to either be created automatically if you leave it blank (when doing POST request, etc) or that if you as a developer include the asset_id on a POST, that it would honor that and set that as the asset_id and it would also not allow duplicate asset_id..
But, if v1 API won't be removed any time soon, then I guess I can just code the custom control center with using the v1 API since it works fine that way because I wouldn't want to create a whole new v1.2.1 function/API just for 2 lines of code 😞

ealmonte32:
closing since it wont be necessary to create new API for these few lines.. will continue to use API v1 instead.

PR #1445, opened 2020-12-16
Fix: successful installing on pure Buster with pi4

rusko124:
Edits: 
- Disabled tasks for wifi-connect in ansible
- Args of Pi verstion for Dockerfile.viewer

Some notes what I've done for success running:
- Rebuilding `screenly/srly-ose-nginx:experimental`
- `$ nmcli d wifi connect <wifi_point> password <password>` will be helpful as well


docker-compose.yml
```diff
@@ -19,6 +19,8 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.viewer
+      args:
+        PI_VERSION: ${DEVICE_TYPE}
```
vpetersson, line 23: No, we can't do this. `docker-compose.yml` is reserved for Balena, and they don't support arguments in their `docker-compose.yml` file.

docker/Dockerfile.viewer
```diff
@@ -112,7 +112,7 @@ RUN pip install --no-cache-dir -r /tmp/requirements.txt
 RUN c_rehash
 
 # QT Base from packages does not support eglfs
-ARG PI_VERSION=pi3
+ARG PI_VERSION
```
vpetersson, line 115: Let's leave this as a default. We have ways to work around this in `docker-compose.override.yml. for Docker on Buster, and [a different way](https://www.balena.io/docs/reference/balena-cli/#-b---buildarg-buildarg) to work around this for balena for non-Pi 3.

docker-compose.yml
ealmonte32: How about a [label](https://docs.docker.com/engine/reference/builder/#label)?

Issue #1460, opened 2020-12-22
Containerd cgroups error - OOM monitoring not supported


ealmonte32:
journalctl shows error from containerd:
```
Dec 21 19:08:09 screenlypi4 containerd[531]: time="2020-12-21T19:08:09.303321597-05:00"
level=warning msg="OOM monitoring failed" error="cgroups: memory cgroup not supported on this system"
```

this comment explicitly talks about the need to enable cgroups or else OOM wont work:
[https://github.com/moby/moby/issues/35587#issuecomment-354308802](https://github.com/moby/moby/issues/35587#issuecomment-354308802)

#### Environment

* Raspberry Pi Hardware Version: Pi4
* Screenly OSE Version: master

---

Will open PR for this..

PR #1462, opened 2020-12-22
Enable cgroup in kernel for containerd

ealmonte32:
Containerd requires cgroup memory enabled

See https://github.com/Screenly/screenly-ose/issues/1460


vpetersson:
Nice find, @ealmonte32! I assumed that was enabled by default.

PR #1465, opened 2020-12-24
remove dummy package libgles2-mesa

ealmonte32:
I know it's just one package for now and the cleanup of all is long process, but one less to do..

```
libgles2-mesa:
transitional dummy package
This is a transitional dummy package, it can be safely removed.


File list of package libgles2-mesa in buster of architecture armhf:
/usr/share/bug/libgles2-mesa/control
/usr/share/bug/libgles2-mesa/script
/usr/share/doc/libgles2-mesa/changelog.Debian.gz
/usr/share/doc/libgles2-mesa/copyright


File list of package libgles2-mesa-dev in buster of architecture armhf:
/usr/include/GLES2/gl2.h
/usr/include/GLES2/gl2ext.h
/usr/include/GLES2/gl2platform.h
/usr/include/GLES3/gl3.h
/usr/include/GLES3/gl31.h
/usr/include/GLES3/gl32.h
/usr/include/GLES3/gl3ext.h
/usr/include/GLES3/gl3platform.h
/usr/lib/arm-linux-gnueabihf/pkgconfig/glesv2.pc
/usr/share/bug/libgles2-mesa-dev/control
/usr/share/bug/libgles2-mesa-dev/script
/usr/share/doc/libgles2-mesa-dev/changelog.Debian.gz
/usr/share/doc/libgles2-mesa-dev/copyright
```

### DIFF
--- a/docker/Dockerfile.viewer
+++ b/docker/Dockerfile.viewer
@@ -24,7 +24,6 @@ RUN apt-get update && \
         libfreetype6-dev \
         libgbm-dev \
         libgcrypt20-dev \
-        libgles2-mesa \
         libgles2-mesa-dev \
         libglib2.0-dev \
         libgst-dev \


vpetersson:
Good find, @ealmonte32. So what I would suggest we start with is to adding in `--no-install-recommends`. After we've done this, we can start removing packages. If not, we run a real risk of losing a package that is an implicit dependency.

ealmonte32:
> Good find, @ealmonte32. So what I would suggest we start with is to adding in `--no-install-recommends`. After we've done this, we can start removing packages. If not, we run a real risk of losing a package that is an implicit dependency.

Since you already included the no install recommends and the viewer still works, what is the best way to sort of go about doing this and removing packages that are not necessary without having to remove each package one by one, reinstall by running docker-compose, and then check logs to see if anything is not working.. I would think a github action or CI should be able to do this right? because remember that sometimes a package or module is not called until a user presses one button or adds an asset or changes a setting, etc.. so it'll be kind of hard to test everything after removing each package if the docker container builds successfully..

PR #1466, opened 2020-12-24
fix webview documentation link and other misc

ealmonte32:
just updating/modifying some webview info and miscellaneous


vpetersson:
👍

PR #1477, opened 2021-01-21
fix for failed vchiq instance on celery container

ealmonte32:


vpetersson:
In theory, we shouldn't need this as no celery task touches the GPU. It must be something doing it indirectly.

That said, this is a pragmatic fix. Accepting.

PR #1488, opened 2021-02-03
Update balena.yml to adhere to latest spec

chrisys:
Just some housekeeping - we're updating the format of balena.yml to be futureproof and in the future the current format will be deprecated. Docs: https://www.balena.io/docs/learn/deploy/deploy-with-balena-button/#balenayml-configuration-file


vpetersson:
Thanks @chrisys! Merging.

Issue #1490, opened 2021-02-07
Backup/Restore Still Not Working with Balena Cloud Install


TheQuank:
Howdy everyone! I've done quite a bit of reading about Backup/Restoring in Screenly when installed with Balena Cloud. There seems to be a number of differing options about it working/not working and whether or not its fixed. I've tried with clean installs over the last couple of days and am not able to get it to work. 

Error Message is: Server Error: [Errno 2] No such file or directory: '/data/screenly/static/screenly-backup-2021-02-07T18-26-10.tar.gz'

I'm running this on balenaOS 2.54.2+rev1 the latest (as of 2/7/21) stable version of Screenly. Any help would be much appreciated!!

TheQuank:
So if I manually create the /data/screenly/static path with 777 permissions inside the srly-ose-server container, it appears to generate the backup file, then errors when trying to download it:

Sorry, this page does not exist!

Logs: 09.02.21 06:58:57 (-0600)  srly-ose-nginx  192.168.5.136 - - [09/Feb/2021:12:58:57 +0000] "GET /favicon.ico HTTP/1.1" 200 32 "http://192.168.5.206/static_with_mime/Reception%20Screenly-2021-02-09T12-56-25.tar.gz?mime=application/x-tgz" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36 Edg/88.0.705.63" "-"

ealmonte32:
@TheQuank 
hmm.. maybe this was forgotten in the QA checklist..
I am able to reproduce this even on native non-balena master branch (this is what we call the version you're referring to)..
I will do some troubleshooting, hoping this is most likely a simple fix on the Dockerfile/docker-compose to create those directories as persistent storage and not manually creating it after the fact..
Will post back if I figure anything out.. if you play around with the Dockerfile and docker-compose files and get it to work first, just post here as well.
Thx

ealmonte32:
Ok I fixed it locally, needed to add the alias in nginx.. making PR for it now..

ealmonte32:
@TheQuank pls run the bash script again to reinstall and it should work.
Dont forget to clone/pull the github repo again since two files were modified and these need to be updated in your local directory when rebuilding the image for balena.

BrendonKoz:
[...]

PR #1495, opened 2021-02-10
Update nginx.conf

ealmonte32:
allow all private subnets to download backup file..

part of fixes for reported issue: https://github.com/Screenly/screenly-ose/issues/1490


vpetersson:
Nice!

PR #1514, opened 2021-03-30
Implement reboot and shutdown for Dockerized Screenly via host agent

nkaminski:
This change implements a mechanism for executing a restricted set of commands on the host from within the screenly Docker containers by using a very simple host agent process, which requires only 2 packaged dependencies (python3 and python3-redis) from the standard Raspbian repositories.

Initially, this functionality is used to implement the previously disabled reboot and shutdown features. However, this could fairly easily be expanded to re-implement the USB asset loading functionality.

Furthermore, this change removes the reinstallation of build-essential, as /usr/lib/gcc/arm-linux-gnueabihf/4.9/cc1plus does not exist after reinstalling such and screenly builds without error.

### DIFF
--- a/host_agent.py
+++ b/host_agent.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+__author__ = "Nash Kaminski"
+__license__ = "Dual License: GPLv2 and Commercial License"
+
+import logging
+import os
+import redis
+import subprocess
+
+# Name of redis channel to listen to
+CHANNEL_NAME = b'hostcmd'
+
+# Explicit command whitelist for security reasons, keys as bytes objects
+CMD_TO_ARGV = {b'reboot': ['/usr/bin/sudo', '-n', '/usr/bin/systemctl', 'reboot'],
+               b'shutdown': ['/usr/bin/sudo', '-n', '/usr/bin/systemctl', 'poweroff']}
+
+def execute_host_command(cmd_name):
+    cmd = CMD_TO_ARGV.get(cmd_name, None)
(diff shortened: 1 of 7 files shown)


host_agent.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+__author__ = "Nash Kaminski"
+__copyright__ = "Copyright 2021, Nash Kaminski"
```
vpetersson, line 5: Please remove this. Feel free to leave the author in, but we can't accept copyrighted code.

host_agent.py
```diff
@@ -0,0 +1,53 @@
...
+__license__ = "MIT License"
```
vpetersson, line 6: Please remove this too. It should inherit the main license.

vpetersson (changes requested):
Great work @nkaminski. Thank you. Just left some minor comments with regards to copyright.

host_agent.py
nkaminski: __copyright__ line / variable removed.

host_agent.py
nkaminski: Changed as requested

nkaminski:
@vpetersson Changes implemented as requested.

Also just a suggestion, it could be worthwhile to add a [https://en.wikipedia.org/wiki/Contributor_License_Agreement](Contributor License Agreement) noting these constraints.

nkaminski:
Corrected minor typo that I noticed in my one of my comments as well

vpetersson:
Thanks @nkaminski. LGTM. Thanks and welcome to the community! :)

PR #1515, opened 2021-03-30
Remove /etc/localtime in viewer container, such that the Docker mount takes precedence

nkaminski:
This remediates an issue where the timezone used by the QTWebView instance within the viewer container is always UTC, regardless of the host timezone.

### DIFF
--- a/docker/Dockerfile.viewer
+++ b/docker/Dockerfile.viewer
@@ -148,6 +148,8 @@ ENV GIT_BRANCH=$GIT_BRANCH
 
 RUN useradd -g video viewer
 
+RUN rm -f /etc/localtime
+
 WORKDIR /usr/src/app
 RUN mkdir -p /usr/src/app
 COPY . /usr/src/app/


vpetersson:
Thanks @nkaminski. This is weird though as this should be overwritten by [this](https://github.com/Screenly/screenly-ose/blob/master/docker-compose.override.yml#L21). Could you verify that this isn't the case?

nkaminski:
@vpetersson The mount that you have referred to does not seem to be taking precedence over the symlink inside the container as long as the symlink is present in the image:

From docker inspect output after reverting this:
```
{
            "Type": "bind",
            "Source": "/etc/localtime",
            "Destination": "/etc/localtime",
            "Mode": "ro",
            "RW": false,
            "Propagation": "rprivate"
}
```

And then inside of the container post-revert:
```
root@98fcce1da6d3:/usr/src/app# ls -al /etc/localtime
lrwxrwxrwx 1 root root 27 Mar 10 08:06 /etc/localtime -> /usr/share/zoneinfo/Etc/UTC
```

vpetersson:
Huh. That's interesting. Sounds almost like a bug in Docker, but yeah good stuff. Thanks!.

PR #1518, opened 2021-04-09
Add timezone configuration through environment variables

tomjschwanke:
When deploying via Balena you can define environment variables. This change applies the `TIMEZONE` variable as a timezone for the viewer so that pages using the internal clock will display the correct time for the timezone.
Taken from the official Balena timezone example: https://github.com/balena-io-examples/balena-timezone/blob/master/start.sh

### DIFF
--- a/bin/start_viewer.sh
+++ b/bin/start_viewer.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Default to UTC if no TIMEZONE env variable is set
+echo "Setting time zone to ${TIMEZONE=UTC}"
+# This only works on Debian-based images
+echo "${TIMEZONE}" > /etc/timezone
+dpkg-reconfigure tzdata
+
 # Fixes permission on /dev/vchiq
 chgrp -f video /dev/vchiq
 chmod -f g+rwX /dev/vchiq


vpetersson:
Thanks! @tomjschwanke. Is `TIMEZONE` exported by default on Balena? If that's the case, we should probably adopt the same system in our environment variables. 

We could do something like this: 
* Set /etc/timezone to UTC in all `Dockerfiles`
* Do a conditional check in the entry-point files in all various containers and check if `TIMEZONE is set. If it is set, apply it.

tomjschwanke:
Hi there, unfortunately the `TIMEZONE` variable is not defined by default. However it makes it really easy to set the time through the Balena Dashboard

vpetersson:
I see. I just wonder how this would interact/interfere with https://github.com/Screenly/screenly-ose/pull/1515.

tomjschwanke:
I have tried it on my Raspberry Pi 3 and it set the timezone correctly and seemed to work. Didn't observe any errors

tomjschwanke:
> We could do something like this: 
> * Set /etc/timezone to UTC in all `Dockerfiles`
> * Do a conditional check in the entry-point files in all various containers and check if `TIMEZONE` is set. If it is set, apply it.

This sounds good, maybe I can look into it, don't have much time right now

PR #1519, opened 2021-04-12
added revision c03114

Samba89:
Came across this revision in a Pi 4 I bought recently, was causing crashes in viewer.py


vpetersson:
Nice! Thanks!

PR #1528, opened 2021-05-13
balena openfleets

shawaj:
add stuff for open fleets

@vpetersson one thing i was thinking - since Deploy with Balena only works with raspbery pi 3...maybe should change the balena.yml to only support that? As balena.yml only really relevant for DwB?


vpetersson:
@shawaj Issue with that is that `balena.yml` is also used by the other deployment method. I'm not sure it's possible to specify a custom one.

PR #1567, opened 2022-01-26
Added autocomplete attributes to form elements

JustinByrne:
Added `autocomplete="off"` to player_name field
Added `autocomplete="new-password"` to both the auth password and password2 fields

These have been added to stop auto filling username and passwords from being entered. This is mainly an issue when you go to make changes to other settings and it renames the player, and tries to change the password if using basic authentication.


ofifoto:
Bump. This bot should really be turned off

Issue #1568, opened 2022-02-07
Installation script is outdated. Some packages have no installation candidate.


NassimBentarka:
#### Overview of the Issue

The main installation script does not install properly on Raspbian Lite. My environment is Raspberry Pi Zero W. It looks like python-pip was renamed to python3-pip and other packages were renamed as well.
The result is that Screenly doesn't install on a fresh Raspbian Lite on Raspberry Pi Zero W.

#### Reproduction Steps

On a fresh Raspbian Lite on Raspberry Pi Zero W, run: 

`bash <(curl -sL https://www.screenly.io/install-ose.sh)`

#### Output

`       _____                           __         ____  _____ ______
      / ___/_____________  ___  ____  / /_  __   / __ \/ ___// ____/
      \__ \/ ___/ ___/ _ \/ _ \/ __ \/ / / / /  / / / /\__ \/ __/
     ___/ / /__/ /  /  __/  __/ / / / / /_/ /  / /_/ /___/ / /___
    /____/\___/_/   \___/\___/_/ /_/_/\__, /   \____//____/_____/
                                     /____/
Screenly OSE requires a dedicated Raspberry Pi / SD card.
You will not be able to use the regular desktop environment once installed.

Do you still want to continue? (y/N)
________________________________________

Which version/branch of Screenly OSE would you like to install:

 Press (1) for the Production branch, which is the latest stable.
 Press (2) for the Development/Master branch, which has the latest features and fixes, but things may break.

You selected: Production

Do you want Screenly OSE to manage your network? This is recommended for most users because this adds features to manage your network. (Y/n)

Would you like to perform a full system upgrade as well? (y/N)
+ REPOSITORY=https://github.com/screenly/screenly-ose.git
+ sudo mkdir -p /etc/ansible
+ sudo tee /etc/ansible/hosts
+ echo -e '[local]\nlocalhost ansible_connection=local'
+ '[' '!' -f /etc/locale.gen ']'
+ sudo sed -i s/apt.screenlyapp.com/archive.raspbian.org/g /etc/apt/sources.list
+ sudo apt update -y
Hit:1 http://archive.raspberrypi.org/debian bullseye InRelease           
Hit:2 http://raspbian.raspberrypi.org/raspbian bullseye InRelease        
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
All packages are up to date.
+ sudo apt-get purge -y python-pyasn1
Reading package lists... Done
Building dependency tree... Done
[...]

Issue #1558, opened 2021-10-14
Raspberry Pi 3 B+ new install with install-ose.sh script fails


ZhangXector:
#### Overview of the Issue

New install to Raspberry Pi 3 B+ (fresh Raspberry Pi OS Lite install) using the install script at https://www.screenly.io/install-ose.sh fails with the following error:

`./install-ose.sh: line 179: /home/pi/screenly/bin/upgrade_containers.sh: No such file or directory`

#### Reproduction Steps

On a fresh Raspberry Pi OS install:
Run: `bash <(curl -sL https://www.screenly.io/install-ose.sh)`
Select: Stable channel
Select: Do not manage network
Select: Perform a system upgrade
Wait for install to reach the point that it fails, exiting the install script.

#### Environment

* Raspberry Pi Hardware Version: 3 B+
* Raspberry Pi Network Setup: Ethernet (Wi-Fi disabled with rfkill)
* Screenly OSE Version: N/A

rebootit:
Bump -- still an issue

ealmonte32:
I will look into some of these issues today, some dont make sense if the repository is pulled down then the bin folder comes down with all files, including that one..

ealmonte32:
I ran into docker-ce not being installed during installation (raspbian buster lite, do not use bullseye).. please see my comment here: https://github.com/Screenly/screenly-ose/issues/1559#issuecomment-1048480683
Worked afterwards.

ZhangXector:
We actually tried both, but I feel like development didn't work either. Stable was just the one I used while writing the reproduction steps. It's been a while since we've tried it, I might be able to try it again tomorrow. Then I can let you know if the development version also has issues.

ZhangXector:
It appears that Bullseye was the culprit. We just ran the installer for the dev version from a fresh Buster (lite) install, and it worked like a charm.

PR #1577, opened 2022-03-26
Update compatible Raspberry Pi OS Lite version

strongly-typed:
As mentioned in https://github.com/Screenly/screenly-ose/issues/1558#issuecomment-1050347737 and https://github.com/Screenly/screenly-ose/issues/1568#issuecomment-1033228026, Debian bullseye is not supported.


vpetersson (approved):

vpetersson:
Closing out as it should now work with the experimental version in Bullseye.

PR #1606, opened 2022-09-16
Upgrades QT

vpetersson:
Attempts to upgrade QT from 5.15.2 to 5.15.6.


vpetersson:
Abandoning this PR as newer versions fail with the following error in the `configure` phase:

```
Running configuration tests...
Project ERROR: Evaluation of condition 'tests.webengine-nodejs && tests.webengine-nodejs.version10' yielded non-boolean value '' in feature 'webengine-nodejs10'.
```

PR #1609, opened 2022-09-19
Fix Qt Cross Compile Related Parts

ondergormez:
* Changes have been made to avoid rewriting the script for new Qt versions.
* Changes were tested on the local machine.


vpetersson (approved):
Nice cleanup, @ondergormez! Thanks!

PR #1612, opened 2022-09-28
Bring back test suite runs on GitHub actions

nicomiguelino:
### General

- Development will be continued on #1616.

### Running Tests Locally (Dockerized)

```bash
docker-compose -f docker-compose.test..yml up -d
docker-compose -f docker-compose.test.yml exec srly-ose-test bash bin/run_tests.sh
```

### Possible Next Steps

- Remove any unnecessary commands in `bin/run_tests.sh`.
- Ensure that all `nose` tests are passing.
- Modify the GitHub Actions workflow file so that tests could be run.
- Speed up the GitHub Actions workflow.
- Modify `README.md` and/or any relevant docs.
- If applicable, restore the command for running PhantomJS tests.
- Make local tests idempotent. Results should be the same after running it multiple times.

### DIFF
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+prepare () {
+    mkdir -p ~/.screenly ~/.config/uzbl/ ~/screenly_assets /tmp/USB/cleanup_folder
+    # cp ansible/roles/screenly/files/uzbl-config ~/.config/uzbl/config-screenly
+    cp ansible/roles/screenly/files/screenly.conf ~/.screenly/
+    cp ansible/roles/screenly/files/screenly.db ~/.screenly/
+    cp bin/install.sh /usr/local/sbin/upgrade_screenly.sh
+    chmod 0700 /usr/local/sbin/upgrade_screenly.sh
+    echo -e "[local]\nlocalhost ansible_connection=local" > ansible/localhost
+    curl https://www.screenly.io/upload/ose-logo.png > /tmp/image.png
+    cp /tmp/image.png /tmp/USB/image.png
+    cp /tmp/image.png /tmp/USB/cleanup_folder/image.png
+    curl https://www.screenly.io/upload/ose-logo.png > ~/screenly_assets/image.tmp
+    curl https://www.screenly.io/upload/big_buck_bunny_720p_10mb.flv > /tmp/video.flv
+    cp tests/assets/asset.mov /tmp/asset.mov
+    export DISPLAY=:99.0
+    # /sbin/start-stop-daemon --start --quiet \
+    #     --pidfile /tmp/custom_xvfb_99.pid --make-pidfile \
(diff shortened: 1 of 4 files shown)


bin/run_tests.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+prepare () {
+    mkdir -p ~/.screenly ~/.config/uzbl/ ~/screenly_assets /tmp/USB/cleanup_folder
+    # cp ansible/roles/screenly/files/uzbl-config ~/.config/uzbl/config-screenly
+    cp ansible/roles/screenly/files/screenly.conf ~/.screenly/
+    cp ansible/roles/screenly/files/screenly.db ~/.screenly/
+    cp bin/install.sh /usr/local/sbin/upgrade_screenly.sh
+    chmod 0700 /usr/local/sbin/upgrade_screenly.sh
+    echo -e "[local]\nlocalhost ansible_connection=local" > ansible/localhost
```
vpetersson, line 10: We should be able to remove this.

bin/run_tests.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+prepare () {
+    mkdir -p ~/.screenly ~/.config/uzbl/ ~/screenly_assets /tmp/USB/cleanup_folder
+    # cp ansible/roles/screenly/files/uzbl-config ~/.config/uzbl/config-screenly
+    cp ansible/roles/screenly/files/screenly.conf ~/.screenly/
+    cp ansible/roles/screenly/files/screenly.db ~/.screenly/
+    cp bin/install.sh /usr/local/sbin/upgrade_screenly.sh
+    chmod 0700 /usr/local/sbin/upgrade_screenly.sh
```
vpetersson, line 9: Permission should already be set from Git.

bin/run_tests.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo -e "[local]\nlocalhost ansible_connection=local" > ansible/localhost
+    curl https://www.screenly.io/upload/ose-logo.png > /tmp/image.png
+    cp /tmp/image.png /tmp/USB/image.png
+    cp /tmp/image.png /tmp/USB/cleanup_folder/image.png
+    curl https://www.screenly.io/upload/ose-logo.png > ~/screenly_assets/image.tmp
+    curl https://www.screenly.io/upload/big_buck_bunny_720p_10mb.flv > /tmp/video.flv
+    cp tests/assets/asset.mov /tmp/asset.mov
+    export DISPLAY=:99.0
```
vpetersson, line 17: This should only be needed from front-end tests.

vpetersson (changes requested):
Looks like a good start, but it is never actually implemented in [docker-build.yaml](https://github.com/Screenly/screenly-ose/blob/master/.github/workflows/docker-build.yaml) as the first job with `needs` set to this job on the existing first step.

nicomiguelino:
@vpetersson, as mentioned in the comments, I will continue development on #1616 instead.

Issue #1580, opened 2022-04-06
Screenly bootscreen shows, then blackscreen Raspberry Pi 4B


sl8tr:
Hey!

I’m really stuck with my Raspberry 4 installation.

So I’ve installed the developers build with no problems, but after a restart the screenly bootscreen is shown and then it switches to a black screen. Only ALT + F1 is working and gets me into the terminal.

I’ve deleted the /home/pi/screenly folder and reran the installation. Still not working.
Started the Installation several times. 

I used the Pi imager and preset for Pi OS Lite (Legacy) with Debian Buster and also downloaded and installed an older one.

I hope you can help me out. I have to get 2 of them done for a company. :/

carlesjavierre:
Same issue here, 4 Pi's in a row. Will try investigate further. I have installed one a month ago and it worked correctly. (Yes, with the rbp4)

carlesjavierre:
I have tried the new installs I made with the "old" pi that I had working with its own install before: THEY WORK. However, the newer hardware I bought doesn't. 

The board that is working has the following:

HW version: BCM2711
Revision: b03114
Model: RBP 4B Rev 1.4

The board not working:
HW version: BCM2711
Revision: b03115
Model: RBP 4B Rev 1.5

However, it seems that the 1.5 revision is correctly added to the list on [here](https://github.com/Screenly/screenly-ose/blob/master/lib/raspberry_pi_helper.py#L144)

I hope this helps debugging.

paFFr:
We've encountered the same issue, Raspberry Pi 4B Rev 1.5, using the Default Buster Lite image and doing a manual installation there won't be any errors during the installation, but on reboot it will show just the Splash Screen and switches to a black screen afterwards. As carlesjavierre wrote the WebUI will be reachable but going to system info will show an internal error.

I also tried to comment out line 49-51 on viewer.py but this sadly didn't change anything in my case.

I have no idea how to get this to work anymore.

carlesjavierre:
@paFFr it requires that you edit the docker-compose so it replaces the file inside the container (that'ts why I said i did some fiddly stuff). I'll edit the comment later on so its full, as it might help others. I just didn't want to post a nasty patch, but if you can get it to work its fine for me!

carlesjavierre:
@paFFr I have updated with full instructions. Please try it out and let me know :smile:

paFFr:
[...]

Issue #1573, opened 2022-03-10
Cannot start viewer on Raspberry Pi Zero 2 W


ClintHaerinck:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

When starting viewer:
```
 srly-ose-viewer  Traceback (most recent call last):
 srly-ose-viewer    File "viewer.py", line 50, in <module>
 srly-ose-viewer      media_player = VLCMediaPlayer() if get_raspberry_model(get_raspberry_code()) == 'Model 4B' else OMXMediaPlayer()
 srly-ose-viewer    File "/usr/src/app/lib/diagnostics.py", line 128, in get_raspberry_model
 srly-ose-viewer      return raspberry_pi_helper.lookup_raspberry_pi_revision(raspberry_pi_revision)['model']
 srly-ose-viewer  TypeError: string indices must be integers, not str
 srly-ose-viewer  ./bin/start_viewer.sh: line 47: kill: (38) - No such process
```

#### Reproduction Steps

Add device type to `raspberrypi0-2w-64` to balena.yml. Push application to device.

#### Environment

* Raspberry Pi Hardware Version: Raspberry Pi Zero 2 W
* Raspberry Pi Network Setup: USB Ethernet Dongle
* Screenly OSE Version: @

semaf:
Seems same issue like here https://github.com/Screenly/screenly-ose/issues/1571

vpetersson:
`OMXMediaPlayer` and the old `lib/diagnostics.py` revision-lookup helper are both gone from the current codebase. Pi Zero 2 W is also outside our supported board matrix. Closing.

Issue #1571, opened 2022-02-22
Assets are not loading, screen keeps black


HOEMichelR:
#### Overview of the Issue
Screenly fresh installed on Raspberry Pi 4 B - Assets are not showing up
Screenly is starting with the Splash screen, I can load Screenly web interface, can add assets, but the screen still keeps black

#### Reproduction Steps
* Installed the latest Raspbian Buster Image 
* Installed developer branch from Screenly: $ bash <(curl -sL https://www.screenly.io/install-ose.sh)
* Considered all hints from #1389 especially @ealmonte32 

#### Environment
* Raspberry Pi Hardware Version: 4 B
* Raspberry Pi Network Setup: LAN

![WhatsApp Image 2022-02-22 at 09 32 22](https://user-images.githubusercontent.com/100192658/155093177-649124ec-472b-405a-9414-3613bc425b5b.jpeg)

I hope, someone can help me?
Thanks in advance

vpetersson:
Is the time set correctly?

HOEMichelR:
Yes:
![Date](https://user-images.githubusercontent.com/100192658/155106436-de0baf15-67df-47a6-b329-89e4381edf86.PNG)

HOEMichelR:
Here is an actual log file created with journalctl:
[LogFile.txt](https://github.com/Screenly/screenly-ose/files/8122388/LogFile.txt)

Does Screenly have an ownl logfile itself?

HOEMichelR:
Update: I've tried the same installation routine with another Raspi 4B, but Rev. 1.1 -> it's working there, but with an older buster image (2021-01-11-raspios-buster-armhf-lite)

My Raspi with the problems is a 4B is Rev. 1.2 - on this one, the old buster image does not load:
![Raspi](https://user-images.githubusercontent.com/100192658/155391142-5d1ac332-8ba4-4f2c-9d03-598f63c42b30.PNG)

So I took the latest one, but with this image I have above-mentioned problems.

ealmonte32:
@HOEMichelR 
Ok, I see..

Well, I tested with the last buster version and end result worked.
During initial installation, the following packages were not installed and are required, which you can manually install even before running the install script:
`python-backports.ssl-match-hostname`
`python-backports.shutil-get-terminal-size`

I also noticed `docker-compose` was missing so installed that too.

[...]

PR #1617, opened 2022-09-29
Tjm/fix helper

tmeiczin:
Currently, when trying to run against a Pi 4B Rev 1.5 with 1GB, the viewer container will continuously restart. The method `lookup_raspberry_pi_revision` returns a str when the lookup fails. This in turn causes `get_raspberry_model` to cause a TypeError exception, since it is assuming the return is a dict. Adding the new type, but also adding some protection code.
 
May fix https://github.com/Screenly/screenly-ose/issues/1580

### DIFF
--- a/lib/diagnostics.py
+++ b/lib/diagnostics.py
@@ -3,11 +3,9 @@
 import os
 import sh
 import sqlite3
-import re
 import utils
 import cec
 from lib import raspberry_pi_helper
-from utils import connect_to_redis
 from pprint import pprint
 from datetime import datetime
 
@@ -84,6 +82,7 @@ def get_git_short_hash():
 def get_git_hash():
     return os.getenv('GIT_HASH')
 
+
 def try_connectivity():
     urls = [
         'http://www.google.com',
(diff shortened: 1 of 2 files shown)


nicomiguelino (approved):

tmeiczin:
@nicomiguelino @vpetersson I believe this PR might also partially fix https://github.com/Screenly/screenly-ose/pull/1617. It should at least prevent the viewer container from crashing, it just won't have all the correct information about the device.

nicomiguelino:
Thanks @tmeiczin! Did you mean #1580? (There might've been a typo.)

tmeiczin:
Sorry, copy/paste error. I meant 1571 and 1573.

nicomiguelino:
Got it, thanks!

nicomiguelino:
Saw the issues. Thanks, @tmeiczin! Those issues (#1571, #1573, #1580) share the same error.
Yes, I agree that it just won't have all the correct device info.

Anyways, I've just realized that the `database` (see [`lib/raspberry_pi_helper.py`](https://github.com/Screenly/screenly-ose/blob/master/lib/raspberry_pi_helper.py#L35)) could be updated based on what's included in the revision codes provided by Raspberry Pi itself. Here are the references for the revision codes:
- https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/raspberry-pi/revision-codes.adoc
- https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#new-style-revision-codes-in-use

PR #1640, opened 2022-10-21
Fix MP4 video playback issue.

### LINKED ISSUE #1590: MP4 not playing & crashing on RaspberryPi 4
#### Overview of the Issue

JPEGS play fine, however when playing an MP4 video, Screenly crashes and reboots.  Did not happen on RPI3.

#### Reproduction Steps

Add MP4 assest in normal way.

#### Environment

* Raspberry Pi Hardware Version: Raspberry Pi 4 2GB
* Raspberry Pi Network Setup: Connected via Ethernet and internal WiFi
* Screenly OSE Version:  Current from GitHub

`Load Average: 0.55`
`Free Space: 24G`
`Memory : Total: 1797 \| Used: 631 \| Free: 647 \| Shared: 63 \| Buff: 40 \| Available: 1078`
`Raspberry Pi Model: Model 4B (Sony UK)`
`Screenly OSE Version: @`

Crash log attached below.

[digitaldisplay_bold-breeze-28.07.22_18_02_34_(+0100).txt](https://github.com/Screenly/screenly-ose/files/9218981/digitaldisplay_bold-breeze-28.07.22_18_02_34_.%2B0100.txt)

Many thanks,

nicomiguelino:
Might fix #1590

#### Checklist

- [x] Modify relevant Ansible files so that /boot/config.txt can be updated as well.
- [ ] Test changes on Balena.

#### References

- https://github.com/RPi-Distro/vlc/issues/3

### DIFF
--- a/.github/workflows/docker-test.yaml
+++ b/.github/workflows/docker-test.yaml
@@ -5,19 +5,22 @@ on:
     branches:
       - master
       - production
-    paths-ignore:
-      - 'README.md'
-      - 'webview/**'
-      - '.github/workflows/build-webview.yaml'
-      - README.md
+    paths:
+      - '!README.md'
+      - '!webview/**'
+      - '!.github/workflows/build-webview.yaml'
+      - '!.github/workflows/build-balena-disk-image.yaml'
+      - '!docs/**'
   pull_request:
     branches:
       - master
       - production
-    paths-ignore:
-      - 'README.md'
-      - 'webview/**'
-      - '.github/workflows/build-webview.yaml'
+    paths:
+      - '!README.md'
+      - '!webview/**'
+      - '!.github/workflows/build-webview.yaml'
+      - '!.github/workflows/build-balena-disk-image.yaml'
+      - '!docs/**'
   workflow_call:
 
 jobs:
--- a/.github/workflows/docker-test.yaml
+++ b/.github/workflows/docker-test.yaml
@@ -16,6 +16,7 @@ on:
       - master
       - production
     paths:
+      - '**/**'
       - '!README.md'
       - '!webview/**'
       - '!.github/workflows/build-webview.yaml'
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -75,7 +75,7 @@
     replace: '#\1'
 
 - name: Add FKMS config for Pi 4.
-  ansible.builtin.lineinfile:
+  lineinfile:
     path: /boot/config.txt
     insertafter: '^\[pi4\]$'
     line: dtoverlay=vc4-fkms-v3d
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -71,8 +71,8 @@
 - name: Change dtoverlay value for DRM VC4 V3D driver
   replace:
     dest: /boot/config.txt
-    regexp: ^(dtoverlay)=vc4-kms-v3d$
-    replace: \1=vc4-fkms-v3d
+    regexp: ^(\[pi4\])$
+    replace: \1\ndtoverlay=vc4-fkms-v3d
 
 - name: Backup kernel boot args
   copy:
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -68,11 +68,35 @@
   tags:
     - touches_boot_partition
 
-- name: Change dtoverlay value for DRM VC4 V3D driver
+- name: Comment out the original dtoverlay config.
   replace:
-    dest: /boot/config.txt
+    dest: ./config.txt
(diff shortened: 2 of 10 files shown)


lib/media_player.py
```diff
@@ -28,7 +28,13 @@ def is_playing(self):
 class VLCMediaPlayer(MediaPlayer):
     def __init__(self):
         MediaPlayer.__init__(self)
-        self.instance = vlc.Instance()
+
+        options = (
+          '--video-x=1920 '
```
vpetersson, line 33: Not sure we want to hardcode this. It might be 4k. Or even a lower resolution. Can we read this dynamically?

lib/media_player.py
```diff
@@ -28,7 +28,13 @@ def is_playing(self):
...
+          '--mmal-display hdmi-2'
```
vpetersson, line 34: Which port is this? The one closest or furthest away from the power plug? Also, can we pass on some mirror mode instead?

nicomiguelino: I'm not sure, 'cause what's printed on the board are HDMI0 and HDMI1.
HDMI0 is the one that's closest to the USB-C power supply port.

For some reason, when I removed the changes from `lib/media_player.py`, video playback is still okay.

nicomiguelino: It seems that video playback will only work if `dtoverlay` is set to `vc4-fkms-v3d` instead of `vc4-kms-v3d` (from `boot/config.txt`).

lib/media_player.py
nicomiguelino: I tried removing the `--video-x` option. As long as `--mmal-display=hdmi-2` is present, it works. Still figuring out if mirror mode is possible.

.github/workflows/docker-test.yaml
```diff
@@ -16,6 +16,7 @@ on:
       - master
       - production
     paths:
+      - '**/**'
```
nicomiguelino, line 19: @vpetersson, it seems like the execution of unit tests are skipped, even though changes were made on Python files, so I've introduced this change.

ansible/roles/system/tasks/main.yml
```diff
@@ -68,6 +68,36 @@
...
+    dest: ./config.txt
+    regexp: ^(dtoverlay=vc4-kms-v3d)$
+    replace: '#\1'
+
+- name: Change dtoverlay value for DRM VC4 V3D driver.
+  replace:
+    dest: ./config.txt
+    regexp: ^(\[pi4\])$
+    replace: \1\ndtoverlay=vc4-fkms-v3d
+
```
vpetersson, line 82: I don't think the above logic is idempotent.  You will keep adding `dtoverlay=vc4-fkms-v3d` to the `[pi4]` block on every run. 

I would probably either try [builtin.ini](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ini_lookup.html) or just [builtin.lineinfile](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html).

ansible/roles/system/tasks/main.yml
```diff
@@ -68,6 +68,36 @@
   tags:
     - touches_boot_partition
 
+- name: Comment out the original dtoverlay config.
+  replace:
+    dest: ./config.txt
```
vpetersson, line 73: This needs to be `/boot/config.txt`.

ansible/roles/system/tasks/main.yml
```diff
@@ -68,6 +68,36 @@
...
+    regexp: ^(dtoverlay=vc4-kms-v3d)$
+    replace: '#\1'
+
+- name: Change dtoverlay value for DRM VC4 V3D driver.
+  replace:
+    dest: ./config.txt
```
vpetersson, line 79: This needs to be `/boot/config.txt`.

ansible/roles/system/tasks/main.yml
```diff
@@ -68,6 +68,36 @@
...
+    regexp: ^(\[pi4\])$
+    replace: \1\ndtoverlay=vc4-fkms-v3d
+
+- name: Specify configs for pi1, pi2, and pi3.
```
vpetersson, line 83: I'm pretty sure you can remove this entire block. This is only needed for pi4.

vpetersson (changes requested):

ansible/roles/system/tasks/main.yml
nicomiguelino: I've used `lineinfile` instead. Thanks!

ansible/roles/system/tasks/main.yml
nicomiguelino: Thanks! I forgot to change it to an absolute path. Was experimenting with Ansible file manipulation on a separate repo.

vpetersson (approved):

PR #1644, opened 2022-10-26
Adds balena disk image build

vpetersson:
Attempts to add automatic build of (preloaded) disk images for Balena as per [this article](https://forums.balena.io/t/balena-preload-on-github-actions/337349/15).

### DIFF
--- a/.github/workflows/build-balena-disk-image.yaml
+++ b/.github/workflows/build-balena-disk-image.yaml
@@ -1,0 +1,60 @@
+name: Build Balena Disk Images
+on: workflow_dispatch
+jobs:
+  balena-build-images:
+    strategy:
+      matrix:
+        board: ['pi1', 'pi2', 'pi3', 'pi4']
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+
+      - name: Setup NodeJS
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-test.yaml
```diff
@@ -5,19 +5,21 @@ on:
...
       - production
-    paths-ignore:
-      - 'README.md'
-      - 'webview/**'
-      - '.github/workflows/build-webview.yaml'
+    paths:
+      - '!README.md'
+      - '!webview/**'
+      - '!.github/workflows/build-webview.yaml'
+      - '!.github/workflows/build-balena-disk-image.yaml'
```
nicomiguelino, line 22: Might as well include `'!docs/**'`?

.github/workflows/build-balena-disk-image.yaml
```diff
@@ -0,0 +1,60 @@
+name: Build Balena Disk Images
+on: workflow_dispatch
```
nicomiguelino, line 2: - Does this mean that this job can be executed on demand?
- Will this job be used by other jobs?

.github/workflows/docker-test.yaml
vpetersson: Good shout. Will add that.

PR #1658, opened 2022-11-10
fix containers not starting after install

ealmonte32:
For some reason, I think the system was pruning the containers before they were starting, because all fresh installs I did would not have containers starting after the installation reboot, `docker ps` would show no containers, and running `./screenly/bin/upgrade_containers.sh` was building and starting them, but that was a manual after the fact process, so I customized the `install.sh` and removed the `sudo docker system prune -f` from the install script and the containers were starting normal after installation.

This PR just moves the bringing up of the containers to the end of the installation script/process just to make sure they are brought up before the reboot.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -188,9 +188,6 @@ cd /home/pi/screenly/ansible
 
 sudo -E ansible-playbook site.yml "${EXTRA_ARGS[@]}"
 
-# Pull down and install containers
-/home/pi/screenly/bin/upgrade_containers.sh
-
 sudo apt-get autoclean
 sudo apt-get clean
 sudo docker system prune -f
@@ -268,6 +265,9 @@ check_defaultpw () {
 }
 
 check_defaultpw;
+
+# Pull down and install containers
+/home/pi/screenly/bin/upgrade_containers.sh
 
 echo -e "Screenly version: $(git rev-parse --abbrev-ref HEAD)@$(git rev-parse --short HEAD)\n$(lsb_release -a)" > ~/version.md
 
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -188,6 +188,9 @@ cd /home/pi/screenly/ansible
 
 sudo -E ansible-playbook site.yml "${EXTRA_ARGS[@]}"
 
+# Pull down and install containers
+/home/pi/screenly/bin/upgrade_containers.sh
+
 sudo apt-get autoclean
 sudo apt-get clean
 sudo docker system prune -f
@@ -265,9 +268,6 @@ check_defaultpw () {
 }
 
 check_defaultpw;
-
-# Pull down and install containers
-/home/pi/screenly/bin/upgrade_containers.sh
 
 echo -e "Screenly version: $(git rev-parse --abbrev-ref HEAD)@$(git rev-parse --short HEAD)\n$(lsb_release -a)" > ~/version.md
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.server
-      network: host
+      network: bridge
     environment:
       - HOME=/data
       - LISTEN=0.0.0.0
@@ -24,7 +24,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.viewer
-      network: host
+      network: bridge
     depends_on:
       - srly-ose-server
     environment:
@@ -42,7 +42,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.websocket
-      network: host
+      network: bridge
     depends_on:
       - srly-ose-server
     environment:
@@ -57,7 +57,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.celery
-      network: host
+      network: bridge
     depends_on:
       - srly-ose-server
       - redis
@@ -76,7 +76,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.redis
-      network: host
+      network: bridge
     ports:
       - 127.0.0.1:6379:6379
     restart: always
@@ -88,7 +88,7 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.nginx
-      network: host
+      network: bridge
     ports:
       - 80:80
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.server
-      network: bridge
     environment:
       - HOME=/data
       - LISTEN=0.0.0.0
@@ -24,7 +23,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.viewer
-      network: bridge
     depends_on:
       - srly-ose-server
     environment:
@@ -42,7 +40,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.websocket
-      network: bridge
     depends_on:
       - srly-ose-server
     environment:
@@ -57,7 +54,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.celery
-      network: bridge
     depends_on:
       - srly-ose-server
       - redis
@@ -76,7 +72,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.redis
-      network: bridge
     ports:
       - 127.0.0.1:6379:6379
     restart: always
@@ -88,7 +83,6 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.nginx
-      network: bridge
     ports:
       - 80:80
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,11 @@
-version: "2"
+version: "2.2"
 services:
   srly-ose-server:
     image: screenly/srly-ose-server
     build:
       context: .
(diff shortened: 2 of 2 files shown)


ealmonte32:
Hold on this.. wondering if that daemon error is the culprit because tried this change and now it didnt work as it did initially..

Screen Shot 2022-11-09 at 11 43 38 PM

nicomiguelino:
Hi @ealmonte32, I've ran your install script, I got a similar error, but on the `screenly-redis-1` container instead.

image

bin/install.sh
```diff
@@ -269,6 +266,9 @@ check_defaultpw () {
 
 check_defaultpw;
 
+# Pull down and install containers
+/home/pi/screenly/bin/upgrade_containers.sh
```
nicomiguelino, line 270: Not sure about how `docker system prune` work. Since `upgrade_containers.sh` executes `docker compose up -d`, the containers should be up and running. Hypothetically, running containers shouldn't be removed after the `prune` command was executed. Disclaimer: Haven't tested this one by myself.

vpetersson: Here are the man pages for`docker system prune`:
https://docs.docker.com/engine/reference/commandline/system_prune/

tl;dr is that it only deletes unused containers and volumes.

ealmonte32: Yeah I was just testing cause I thought maybe the containers were still starting when the command then ran, but after telling Nico I did three fresh installs with that prune commented out one time it worked then another time leaving the building of containers to the end also worked but then a final test didn't work but all I did different was say no to manage network but Nico said he tried it with saying yes to everything so I'll work on this more on weekend.. I was telling Nico we need to leave the PR just cause the issue is still active and we just have to change up the commits when we figure out the issue.

ealmonte32:
@nicomiguelino 
After some more reading for this specific error and without restarting the pi after the initial error that prevents the redis and server containers from being brought up, the only way I got the containers to build was to add the `network_mode: host` flag to the `docker-compose.yml` file... but this is not a solution since it ruins the ip address of the web server/containers..
But this error has been popping up on some other [github repos](https://github.com/portainer/portainer/issues/6110) specific to docker and raspbian/debian/ubuntu..
There is something we're just not seeing that happens upon reboot that fixes it, and also why would the other containers be able to come up while these two always fail.. 😫 


Screen Shot 2022-11-15 at 1 43 40 AM

Screen Shot 2022-11-15 at 1 43 02 AM

ealmonte32:
@nicomiguelino 

Got it working after changing the docker-compose.yml version to `2.2` since the network parameter was introduced in that version, and adding the `network: host` to each container build.

Please try my test branch and confirm if it works for you..
run the bash install with this URL:
https://raw.githubusercontent.com/ealmonte32/ealmonte32.github.io/main/resources/install.sh

ealmonte32:
P.S. the reason why i added the `network: host` parameter is because the docker compose kept failing/having an issue going out to get packages when i was testing by manually running the builds and it would always error on the apt-get part of the build/dockerfile, so this led me to the networking part of it and based on the statement below by someone in the docker forums, it just makes sense to always allow the network through the host when building the containers.

Screen Shot 2022-11-16 at 1 48 17 AM

vpetersson:
Switching to host networking isn't a solution tho. We can't make that change upstream as it has a lot of implications. 

The root cause must be elsewhere.

ealmonte32:
> Switching to host networking isn't a solution tho. We can't make that change upstream as it has a lot of implications.
> 
> The root cause must be elsewhere.

Hmm.. but this option is only allowing the connection to go out through host to get packages during the building of the containers, are the implications dealing with container/host security and network/communication isolation?

I am also testing to see if this is a bullseye vs buster issue or a docker-ce version issue as well..

vpetersson:
> Hmm.. but this option is only allowing the connection to go out through host to get packages during the building of the containers, are the implications dealing with container/host security and network/communication isolation?

No, no. If you set `network: host` in the docker-compose file, that is how the container is run. The build process doesn't care about this setting at all.

ealmonte32:
> No, no. If you set network: host in the docker-compose file, that is how the container is run. The build process doesn't care about this setting at all.

So following that, and based on this:
image

Means by default the containers are using bridge by default? And if so, I used `network: bridge` instead just now and it also worked properly and allowed containers to be built.

vpetersson:
Yeah the default is using a per-namespace bridge network if memory serves me right.

vpetersson:
Yes bumping the compose version is fine as long we don't exceed the version supported by balena (which iirc is 2.x)

ealmonte32:
2.4 according to their reference [doc](https://www.balena.io/docs/reference/supervisor/docker-compose/)

vpetersson:
Strange that that solves the issue but hey 🤷‍♂️

ealmonte32:
> Strange that that solves the issue but hey 🤷‍♂️

Yeah, this is a head-scratcher... wish we could have an SME in docker compose tell us why the veth pairing operation fails saying it isn't supported during the building of container.. oh well 🤷🏻‍♂️

docker-compose.yml
```diff
@@ -1,10 +1,11 @@
-version: "2"
+version: "2.2"
 services:
   srly-ose-server:
     image: screenly/srly-ose-server
     build:
       context: .
       dockerfile: docker/Dockerfile.server
+      network: bridge
```
vpetersson, line 8: Please remove all of these, as that is the default.

vpetersson:
I am still torn if this is a fluke or not that bumping the version to 2.2, but I guess there is no harm either.

ealmonte32:
> I am still torn if this is a fluke or not that bumping the version to 2.2, but I guess there is no harm either.

Well, this doesnt fix the issue though, I am still testing and troubleshooting on fresh installs, they are all coming up with this issue which i still think it might be attributed to docker-ce itself and not bullseye.. but also it is weird because nico tried the changes i made and for him it didnt fix it yet it did for me, so more troubleshooting awaits.. 😭

ealmonte32:
😡  now all of a sudden everything is working fine without any changes, meaning i have installed fresh copy of bullseye and ran screenly github repo install script three different times from scratch, on pi4 and pi3, and it has worked, i even went as far as getting the older .deb versions of docker-ce and docker-compose to see if it was something with that, and that didnt work either, so when i gave up and tried again recently, all worked fine.. i'll just close this after a week without any reported problems.. 😞

vpetersson:
Random idea. Try using 'docker compose' rather than 'docker-compose'. These are actually different. The installer should be using the former. But this could explain it as the former is built into docker and the latter isn't.

ealmonte32:
> Random idea. Try using 'docker compose' rather than 'docker-compose'. These are actually different. The installer should be using the former. But this could explain it as the former is built into docker and the latter isn't.

the `upgrade_containers.sh` script uses docker compose and not docker-compose, i just typed docker-compose because the .deb file was called that 😐 
i think if we dont get multiple reports of issues with black screen anymore which is this error, we can probably just mark it as one of the great docker compose mysteries 🔍

vpetersson:
So you should really not need to install `docker-compose` anymore. It should be packaged into `docker-ce` in recent versions.

ealmonte32:
@vpetersson I was reading something on one of the scripts created by the docker team which they use to fix any sort of corrupt install, and one of their last sentences about [daemon/container](https://docs.docker.com/engine/security/#docker-daemon-attack-surface) attack surfaces got me thinking, you know how we check for the default pi user password, but we simply advise for them to change it, based on reading that and the fact we dont use rootless mode, i think we should give users a stronger warning and maybe even reference that url that says if they dont change the default pi user password, that they accept the risk of the vulnerabilities that can be side-loaded from docker itself.. dunno, just thought about it real quick..


Anyway.. we're back to the issue, I just tried the fresh install again from screenly ose repo and got the error:
image

I will test out removing docker-compose and see, last time i tried it said it didnt recognize the command "compose" so that's why I left it installed..

vpetersson:
> I was reading something on one of the scripts created by the docker team which they use to fix any sort of corrupt install, and one of their last sentences about [daemon/container](https://docs.docker.com/engine/security/#docker-daemon-attack-surface) attack surfaces got me thinking, you know how we check for the default pi user password, but we simply advise for them to change it, based on reading that and the fact we dont use rootless mode, i think we should give users a stronger warning and maybe even reference that url that says if they dont change the default pi user password, that they accept the risk of the vulnerabilities that can be side-loaded from docker itself.. dunno, just thought about it real quick..

I'm conflicted. On the one hand, *securing* the Pi is completely out of scope for the project, but you could also argue that it is good hygiene. 

> Anyway.. we're back to the issue, I just tried the fresh install again from screenly ose repo and got the error:

Can we just rule out that we're not dealing with a broken SD card too? Also, you didn't install docker *first*, right?

ealmonte32:
> I'm conflicted. On the one hand, _securing_ the Pi is completely out of scope for the project, but you could also argue that it is good hygiene.

Yeah, it's just simply putting stronger wording in the installer script, doesn't do any harm and we're not doing extra work.. up to you..
___

> Can we just rule out that we're not dealing with a broken SD card too? 

Yeah I have plenty of SD cards that work perfectly fine, Nico also tested on his end and got the same error which rules out anything with bad SD card.
___

> Also, you didn't install docker _first_, right?

I didn't install anything, these tests are fresh installs of latest release of bullseye lite, and running the screenly ose install script from bash..
At the end of the installation, the only packages installed are:
image

ealmonte32:
giving up on this.. everything i've read on docker forums and raspbian say it might deal with veth kernel and running rpi-update and rebooting fixes it, which it does but i cannot figure out why it works and why other times it doesnt on fresh installs.. since the fix is easier than finding the culprit, closing.

PR #1663, opened 2022-11-23
Update raspberry_pi_helper.py

ealmonte32:
Even if we stop using this file in a near future, it seems the key "model name" is not part of the outcome of proc/cpuinfo in bullseye, rather it just says "Model".

The first run of test file `model.py` using the original code returns cpu_count, hardware, revision, and serial, but not model. Second edited test file `model.py` which replaces the "model name" key with "Model" as it is returned by proc/cpuinfo, properly shows the model:
Screen Shot 2022-11-22 at 11 49 13 PM


vpetersson (approved):

ealmonte32:
@nicomiguelino 
I assume we can get a better method for looking up which pi model we want, but for now if we want this to only work when model is Pi 4+, then we have to use Pi and 4, because just looking for `Model B` would be for older Pi's as well, unless that was the original reason, which I actually forgot what we wanted to look for.. 🤔 

image

ealmonte32:
Unit tests are failing but saying it's because of chrome browser version... 🤨

ealmonte32:
@vpetersson 

I should really be using model here instead of revision right.. 🤔 

image

ealmonte32:
@nicomiguelino 
I think this is good to go.. unless you have something you would want to change..

vpetersson:
@nicomiguelino please review and sign off.

nicomiguelino:
@vpetersson, will do

lib/diagnostics.py
```diff
@@ -112,48 +112,32 @@ def get_debian_version():
...
+@TODO
+Need to find different methods (built-in) that does not rely on manually entered static database info to get these results.
+Example, for mem: cat /proc/meminfo for memory
+Not sure how we should approach this for now, but not really important at the moment.
 
 def get_raspberry_ram(raspberry_pi_revision):
-    """
-    Quick DRY workaround. Needs to be refactored later.
-    """
     return raspberry_pi_helper.lookup_raspberry_pi_revision(raspberry_pi_revision).get('ram', False)
```
nicomiguelino, line 122: Remove this call to `lookup_raspberry_pi_revision`.
You can also remove the function definition for `get_raspberry_ram`.

lib/diagnostics.py
```diff
@@ -112,48 +112,32 @@ def get_debian_version():
...
 
-
 def get_raspberry_manufacturer(raspberry_pi_revision):
-    """
-    Quick DRY workaround. Needs to be refactored later.
-    """
     return raspberry_pi_helper.lookup_raspberry_pi_revision(raspberry_pi_revision).get('manufacturer', False)
```
nicomiguelino, line 125: Remove this call to `lookup_raspberry_pi_revision`.
You can also remove the function definition for `get_raspberry_manufacturer` since this method is not being used anymore (upon `grep`-ping on the source code).

nicomiguelino (changes requested):
Hello, @ealmonte32. Added some minor cleanup comments.

nicomiguelino:
@ealmonte32, after resolving the comments, the changes are good to go.

nicomiguelino (approved):

PR #1665, opened 2022-11-23
Remove ffmpeg installation and setup from script & Dockerfile.

nicomiguelino:
Relevant parts of the code were removed. These will be brought back when `test_add_asset_streaming` is fixed.


nicomiguelino:
@vpetersson, since `test_add_asset_streaming` is not being run at the moment, I removed the installation and setup of `ffmpeg`, as this could cause overhead in CI build times.

vpetersson:
Let's just comment it out. We're going to have to restore it shortly anyways so feels a bit excessive to remove it fully.

nicomiguelino:
Got it @vpetersson. Just pushed a new commit. Requesting another review.

vpetersson (approved):

PR #1668, opened 2022-11-26
Feature/make swap on disc configurable

oxivanisher:
**Attention: This PR has not been tested yet!** I sadly don't have a Pi available. If somebody else wants to test it, please let me know so I can help to debug any problems.

This feature allows to configure aspects of the ansible based setup. The first use case ist to not disable the dphys swapfile removal.

Also: I think it is better to disable and mask the `dphys-swapfile` service. If it only gets removed, its always possible that a upgrade will re-install the package, since it is a Raspbian base OS package. If a service gets disabled by a user (in this case the screenly ose setup), it will remain even if a update is installed.


oxivanisher:
As discussed in #1668.

Issue #1666, opened 2022-11-25
Installation fails using the installation script: Docker fails to start


NassimBentarka:
#### Overview of the Issue

I'm experiencing the following issue on a fresh Raspbian Bullseye installation on Raspberry Pi Zero W. The issue also happens on a fresh Raspbian Buster installation as well.

When the [Installation Script](https://install-ose.srly.io) finishes, Screenly doesn't run after the reboot. When we check on the status of `docker.service` we get the following output:
```
pi@raspberrypi:~ $ sudo systemctl status docker
● docker.service - Docker Application Container Engine
     Loaded: loaded (/lib/systemd/system/docker.service; enabled; vendor preset: enabled)
     Active: inactive (dead)
TriggeredBy: ● docker.socket
       Docs: https://docs.docker.com

Nov 25 13:17:30 raspberrypi systemd[1]: Dependency failed for Docker Application Container Engine.
Nov 25 13:17:30 raspberrypi systemd[1]: docker.service: Job docker.service/start failed with result 'dependency'.
Nov 25 13:24:57 raspberrypi systemd[1]: Dependency failed for Docker Application Container Engine.
Nov 25 13:24:57 raspberrypi systemd[1]: docker.service: Job docker.service/start failed with result 'dependency'.
```
After futher investigating, it looks like `docker.service` fails because it depends on `containerd.service`, whose `ExecStart=/usr/bin/containerd` command itself fails.

When we run `sudo /usr/bin/containerd` we get the following:
```
pi@raspberrypi:~ $ sudo /usr/bin/containerd
Segmentation fault
```

Now I found out that the docker fails to start because `/usr/bin/containerd` fails with a Segmentation Fault.

According to [Issue Comment 615203782](https://github.com/containerd/containerd/issues/4112#issuecomment-615203782) and [Issue Comment 615193596](https://github.com/moby/moby/issues/40733#issuecomment-615193596), the reason `/usr/bin/containerd` fails with a Segmentation Fault is that `docker-ce` using a Debian source instead of a Raspbian source to compile.

#### How to Fix (temporarily as of 11/24/2022 / or at least until my Pull Request is merged)

I have been able to have Containerd and Docker, and Screenly-OSE (of course) running by replacing the the source to be `raspbian` rather than `debian` on `/etc/apt/sources.list.d/docker.list` as you can see below:

From this:
```
pi@raspberrypi:~ $ cat /etc/apt/sources.list.d/docker.list 
[...]

Issue #239, opened 2014-11-16
Display Time Needed


cabsports:
Hi There,

We are using Screenly OSE for our Menu Boards with in our club but have come across an enhancement that would be fantastic if it could be implemented.

We have one menu board for day time (9am to 5pm) and then one evenings (5pm to midnight).  At the moment there is no way to setup the display in advance so it would be great to add a display time with in the Active date & time that is already in the system

Example:
Menu Board 1: 
Active Date & Time: 01/Nov/2014 00:00 to 30/Nov/2014 23:59
Display Time: 09:00 to 17:00

Menu Board 2: 
Active Date & Time: 01/Nov/2014 00:00 to 30/Nov/2014 23:59
Display Time: 17:00 to 23:59

I hope I have posted this in the correct area.

Thanks
Duncan

PeacefulRuler:
+1 on this feature. It would definitely be nice.

Thanks for this great software.
Fred

lenny76:
+1

danbsfc:
+1 for this. Something that would certainly be a welcome addition to OSE if possible?

Thanks

Dan

cabsports:
Hi @vpetersson,

Just wondering if any thought has been given to adding this to Screenly OSE yet?

Thanks
Duncan

Issue #228, opened 2014-10-08
Asset times changes on edit


tengil:
When an asset is edited the start and end times changes automatically. Times decreses usually by two hours but sometimes only by one. You can test this just by clicking on edit and then on Save.

Version: 2014-08-13

over64:
wait for https://github.com/wireload/screenly-ose/pull/318
Fixed and tested on FF, Chrome, IE11. Bug was reproduced on FF and IE11

PR #1669, opened 2022-11-27
Patching Issue #1666

NassimBentarka:
[Line #239](https://github.com/Screenly/screenly-ose/blob/5f07e0abc356db574aa4083b32c96711807b19c4/ansible/roles/system/tasks/main.yml#L239) contradicts with [Line #228](https://github.com/Screenly/screenly-ose/blob/5f07e0abc356db574aa4083b32c96711807b19c4/ansible/roles/system/tasks/main.yml#L228)

This fixes the issue [#1666](https://github.com/Screenly/screenly-ose/issues/1666)

### DIFF
--- a/ansible/roles/system/tasks/main.yml
+++ b/ansible/roles/system/tasks/main.yml
@@ -236,7 +236,7 @@
   lineinfile:
     path: /etc/apt/sources.list.d/docker.list
     create: yes
-    line: "deb [arch=armhf] https://download.docker.com/linux/debian {{ raspbian_name.stdout }} stable"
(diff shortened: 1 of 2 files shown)


nicomiguelino:
@vpetersson, although #1666 focuses on Raspberry Pi Zero W, the changes can be tested on the Pi 3 and 4 as well.

vpetersson:
> Sounds good, let's run some testing on Pi 2, 3, and 4 and see. If not, then maybe it should only apply to Pi Zero W.

Yeah I'm inclined to only apply this to Pi Zero as it seems to be isolated to this board.

nicomiguelino:
@NassimBentarka, closing this one as #1666 is already closed.

PR #1673, opened 2022-12-01
Update install.sh

ealmonte32:
`-f` which is for file was used instead of `-d` for directory.. and doing testing on buster now brought this up.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -133,7 +133,7 @@ if [ -z "${REPOSITORY}" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


vpetersson (approved):

ealmonte32:
@vpetersson I still cant merge myself even after approvals.

ealmonte32:
@vpetersson the tests are breaking because the version of the browser from the logs I saw..

vpetersson:
Got it. So we need to bump up [this line](https://github.com/Screenly/screenly-ose/blob/master/docker/Dockerfile.test#L29).

PR #1675, opened 2022-12-01
Update Dockerfile.test

ealmonte32:
@vpetersson 

Further thinking about this, noticed this part:
image

So changed to specify specific versions of chrome driver and browser.
Agree?

### DIFF
--- a/docker/Dockerfile.test
+++ b/docker/Dockerfile.test
@@ -26,12 +26,11 @@ ADD requirements/requirements.txt /tmp/requirements.txt
 ADD requirements/requirements.dev.txt /tmp/requirements.dev.txt
 ADD requirements/requirements.viewer.txt /tmp/requirements.viewer.txt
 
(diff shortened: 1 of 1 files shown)


vpetersson (approved):
Good point.

vpetersson:
Is Chromium better than Chrome in this case?

ealmonte32:
> Is Chromium better than Chrome in this case?

I cannot currently find the chrome. deb package directly from google because their stable is always the latest so we cannot hard code that, unless we download it now, rename it to version 108, and upload to our own ose repo under tests resources and link the raw url to that..

ealmonte32:
@vpetersson all good now, ready to merge whenever you want.

nicomiguelino:
All goot, @ealmonte32

ealmonte32:
Remember, one of you needs to merge since it doesn't let me for whatever reason.

PR #1676, opened 2022-12-01
Copyright year update

ealmonte32:
Just updating the copyright years.

### DIFF
--- a/server.py
+++ b/server.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 __author__ = "Screenly, Inc"
-__copyright__ = "Copyright 2012-2021, Screenly, Inc"
+__copyright__ = "Copyright 2012-2023, Screenly, Inc"
 __license__ = "Dual License: GPLv2 and Commercial License"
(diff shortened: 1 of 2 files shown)


server.py
```diff
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 __author__ = "Screenly, Inc"
-__copyright__ = "Copyright 2012-2021, Screenly, Inc"
+__copyright__ = "Copyright 2012-2023, Screenly, Inc"
```
vpetersson, line 5: Can't really make it 2023 for another 30 days, can we? :)

vpetersson (changes requested):

server.py
ealmonte32: "future-proofing" 😆

vpetersson: Can't legally do that I'm afraid.

ealmonte32: fine, we'll leave this here for 30 days :)

vpetersson (approved):

PR #1688, opened 2022-12-12
Configure Raspberry Pi 4 to use VLC again.

### LINKED ISSUE #1685: Use a media player other than OMXPlayer for Raspberry Pi 4
#### Overview of the Issue

Raspberry Pi 4 uses VLC for playing videos. However, a blocking issue popped up &ndash; see #1679 for details.
One workaround is to use OMXPlayer &ndash; which is deprecated &ndash; in the meantime. See #1684 for relevant changes.

#### Environment

* Raspberry Pi Hardware Version: Raspberry Pi 4 Model B
* Raspberry Pi Network Setup: Ethernet
* OS: Raspberry Pi OS Lite

nicomiguelino:
Fixes #1685

#### Tasks

- [x] Make VLC work with an audio jack as well.
- [x] Check to see if the audio jack feature works on OMX.


nicomiguelino:
This might come in handy &ndash; https://forums.raspberrypi.com/viewtopic.php?p=2061087.
It seems that they pinned VLC to an older version to fix the issue.

However, I re-built the images on my Pi 4 (using the `---no-cache`  flag.

Here's the console output when I ran `sudo -u viewer vlc --version` inside the `srly-ose-viewer` service/container before re-building...

```
VLC media player 3.0.17.4 Vetinari (revision 3.0.13-8-g41878ff4f2)
VLC version 3.0.17.4 Vetinari (3.0.13-8-g41878ff4f2)
Compiled by buildd on arm-conova-01.debian.org (Dec  2 2022 20:18:03)
Compiler: gcc version 8.3.0 (Debian 8.3.0-6)
This program comes with NO WARRANTY, to the extent permitted by law.
You may redistribute it under the terms of the GNU General Public License;
see the file named COPYING for details.
Written by the VideoLAN team; see the AUTHORS file.
```

Here's the output after re-build...

```
VLC media player 3.0.17.4 Vetinari (revision 3.0.13-8-g41878ff4f2)
VLC version 3.0.17.4 Vetinari (3.0.13-8-g41878ff4f2)
Compiled by serge on arm-build.pitowers.org (Dec  7 2022 23:06:22)
Compiler: gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1)
This program comes with NO WARRANTY, to the extent permitted by law.
You may redistribute it under the terms of the GNU General Public License;
see the file named COPYING for details.
Written by the VideoLAN team; see the AUTHORS file.
```

You'll notice that the latter was compiled last Dec. 7, but the former was compiled last Dec. 2, which prompted me to re-build the images.

@vpetersson, we can merge these changes so that a new Docker image will be built and that they will be deployed to Balena instances. Alternatively, VLC could be pinned to an older version. I'd like to hear your thoughts.

vpetersson:
> This might come in handy – https://forums.raspberrypi.com/viewtopic.php?p=2061087. It seems that they pinned VLC to an older version to fix the issue.
> 
> However, I re-built the images on my Pi 4 (using the `---no-cache` flag.
> 
> Here's the console output when I ran `sudo -u viewer vlc --version` inside the `srly-ose-viewer` service/container before re-building...
> 
> ```
> VLC media player 3.0.17.4 Vetinari (revision 3.0.13-8-g41878ff4f2)
> VLC version 3.0.17.4 Vetinari (3.0.13-8-g41878ff4f2)
> Compiled by buildd on arm-conova-01.debian.org (Dec  2 2022 20:18:03)
> Compiler: gcc version 8.3.0 (Debian 8.3.0-6)
> This program comes with NO WARRANTY, to the extent permitted by law.
> You may redistribute it under the terms of the GNU General Public License;
> see the file named COPYING for details.
> Written by the VideoLAN team; see the AUTHORS file.
> ```
> 
> Here's the output after re-build...
> 
> ```
> VLC media player 3.0.17.4 Vetinari (revision 3.0.13-8-g41878ff4f2)
> VLC version 3.0.17.4 Vetinari (3.0.13-8-g41878ff4f2)
> Compiled by serge on arm-build.pitowers.org (Dec  7 2022 23:06:22)
> Compiler: gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1)
> This program comes with NO WARRANTY, to the extent permitted by law.
> You may redistribute it under the terms of the GNU General Public License;
> see the file named COPYING for details.
> Written by the VideoLAN team; see the AUTHORS file.
> ```
> 
> You'll notice that the latter was compiled last Dec. 7, but the former was compiled last Dec. 2, which prompted me to re-build the images.
> 
> @vpetersson, we can merge these changes so that a new Docker image will be built and that they will be deployed to Balena instances. Alternatively, VLC could be pinned to an older version. I'd like to hear your thoughts.


It's unclear to me why the feature was dropped. The forum thread alludes to the fact that it is security related, but haven't dug further. If it was dropped because of security, we shouldn't reinstate an insecure version. 

It does appear that Raspbian is pretty up to date with upstream VLC so I don't think we'll gain much from compiling ourselves (other than being able to tweak build configs).

nicomiguelino:
Thanks @vpetersson. I agree. We can just stick to using VLC installed from a repository (i.e., using `apt-get`).

lib/media_player.py
```diff
@@ -75,7 +75,7 @@ def __init__(self):
     def set_asset(self, uri, duration):
         settings.load()
 
-        if self._arch in ('armv6l', 'armv7l', 'aarch64'):
```
nicomiguelino, line 78: Don't drop `aarch64`.

vpetersson (changes requested):
Good work. Are we not duplicating the settings here though?

lib/media_player.py
```diff
@@ -50,8 +55,14 @@ def __get_options(self):
         return options
 
     def set_asset(self, uri, duration):
-        # @TODO: HDMI or 3.5mm jack audio output
         self.player.set_mrl(uri)
+        settings.load()
+
+        # @TODO: Refactor this conditional statement.
+        if settings['audio_output'] == 'local':
```
nicomiguelino, line 62: @vpetersson Thanks for the prompt feedback! 🙌🏽 Yeah, at first glance, it seems that I've duplicated the settings check (as seen in the `__init__` and `set_asset` methods). Without the settings check during `__init__`, the first video asset in the playlist will still use HDMI for the audio output, even though the settings say it should be via the audio jack.

I'll keep you posted. :radio:

nicomiguelino: @vpetersson, I've just pushed the refactor recently. I tried using VLC methods like `audio_output_device_set` inside `__init__`, but didn't work. The best I could do for now is to do a refactor.

vpetersson: Let's spend a few more cycles on it as it is an important feature that we can't lose.

vpetersson (approved):

PR #1689, opened 2022-12-13
Updates GitHub reference

vpetersson:
Updates all references to GitHub references from 'screenly-ose' to 'anthias'.


bin/install.sh
```diff
@@ -126,10 +126,10 @@ fi
 if [ -z "${REPOSITORY}" ]; then
   if [ "$WEB_UPGRADE" = false ]; then
     set -x
-    REPOSITORY=${1:-https://github.com/screenly/screenly-ose.git}
+    REPOSITORY=${1:-https://github.com/screenly/anthias.git}
```
nicomiguelino, line 129: Just a minor comment. This could be capitalized &ndash; `Anthias`. (Apply to the rest of the changes.)

nicomiguelino (approved):
Just added a minor comment.

bin/install.sh
vpetersson: Yeah i noticed that, but GitHub isn't case sensitive so i rather make it consistent.

PR #1697, opened 2022-12-15
Resolve Wi-Fi connectivity issues on Raspberry Pi running Raspberry Pi OS Lite.

nicomiguelino:
#### Overview

- Might fix #1628

#### Custom Installation Command

```bash
cd && \
echo "export CUSTOM_BRANCH='1628-wifi-connect'" > .env && \
bash <(curl -sL https://raw.githubusercontent.com/nicomiguelino/Anthias/custom-install/bin/install.sh)
```

If you want to skip the part where Docker images are being pulled, run the following:

```bash
cd && \
echo "export CUSTOM_BRANCH='1628-wifi-connect'" > .env && \
echo "export EXIT_EARLY='1'" > .env && \
bash <(curl -sL https://raw.githubusercontent.com/nicomiguelino/Anthias/custom-install/bin/install.sh)
```

PR #1700, opened 2022-12-16
Add changes for setting up WiFi Connect on a Raspberry Pi running balenaOS and Raspberry Pi OS Lite.

### LINKED ISSUE #1628: Resolve issue with balena-wifi-connect (former resin-wifi-connect)
It appears as we've broken the WiFi configuration somewhere along the upgrade process (either in Bullseye or in the containerization).

Since we're using an old version, the first process would probably be to upgrade it (we're still using resin-wifi-connect rather than balena-wifi-connect). Hopefully that will get us going.

### LINKED ISSUE #1699: Add WiFi Connect support for Raspberry Pi devices running balenaOS.
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

- Make use of Balena's [Wifi Connect](https://github.com/balena-os/wifi-connect) so that ad-hoc Wi-Fi connectivity can be supported as well.
- We can refer to the **_[balenaOS: multicontainer app](https://github.com/balena-os/wifi-connect#balenaos-multicontainer-app)_** section for details on how to set **_WiFi Connect_** up.

#### Environment

* Raspberry Pi Hardware Version: 3 Model B+, 4 Model B
* Raspberry Pi Network Setup: Ethernet (initially)

#### References

- https://www.youtube.com/watch?v=Sn2_cuG4DYE

nicomiguelino:
#### Overview

- Might fix #1699
- Might fix #1628 

#### Custom Installation Command

```bash
cd && \
echo "export CUSTOM_BRANCH='1699-wifi-connect'" > .env && \
```

If you want to skip the part where Docker images are being pulled, run the following:

```bash
cd && \
echo "export CUSTOM_BRANCH='1699-wifi-connect'" > .env && \
echo "export EXIT_EARLY='1'" > .env && \
```


nicomiguelino:
Here's the most recent output from the Balena instance:

```
[Logs]    [12/16/2022, 12:38:04 PM] [wifi-connect] Starting WiFi Connect
[Logs]    [12/16/2022, 12:38:04 PM] [wifi-connect] [network_manager::device:WARN] Undefined device type: 30
[Logs]    [12/16/2022, 12:38:04 PM] [wifi-connect] WiFi device: wlan0
[Logs]    [12/16/2022, 12:38:04 PM] [wifi-connect] Access points: ["Hotspot_D53_2.4GHz", "Ratlayon", "", "BubbaTot", "xfinitywifi", "XFINITY", "Hotspot_D53_5GHz", "ATT7aWc66Y", "ATTu7SbDbA", "NETGEAR74", "BONDS21", "ATTMuwMbA2", "Chime-70dafd", "iAqualink-EPY"]
[Logs]    [12/16/2022, 12:38:04 PM] [wifi-connect] Starting access point...
[Logs]    [12/16/2022, 12:38:06 PM] [wifi-connect] Access point 'WiFi Connect' created
[Logs]    [12/16/2022, 12:38:06 PM] [wifi-connect] Starting HTTP server on 192.168.42.1:80
[Logs]    [12/16/2022, 12:38:06 PM] [wifi-connect] Error: Cannot start HTTP server on '192.168.42.1:80': address in use
```

vpetersson:
Ah yes. There is logic in start_server that should essentially hold fire if Wi-Fi connect is running.

nicomiguelino:
Got it, @vpetersson. I found something that's more customizable than [balena-os/wifi-connect][1] &ndash; [balena-labs-projects/wifi-connect][2]. Although they have different organizations, they're both from Balena (checked the URLs from their respective GitHub home pages). The latter is like a wrapper around the former.


I've modified the port from the default `80` to `8000`. The captive portal that automatically appears does not appear. Instead, I go to a browser on my phone and enter the IP address manually. There could be an alternative solution to this one. Still reading up stuff.


[1]: https://github.com/balena-os/wifi-connect
[2]: https://github.com/balena-labs-projects/wifi-connect

vpetersson:
(Resin is the old name for Balena. They renamed the company a few years ago)

wifi-connect/Dockerfile.template
```diff
@@ -0,0 +1,37 @@
+FROM balenalib/%%BALENA_ARCH%%-python as builder
```
vpetersson, line 1: I'm not sure dockerizing this will work on Raspbian as `wifi-connect` relies on communicating with Network Manager over DBUS. It's possible to get DBUS to work in Docker, but probably requires a fair bit of work.

wifi-connect/Dockerfile.template
```diff
@@ -0,0 +1,37 @@
...
+    | grep -hoP 'browser_download_url": "\K.*%%BALENA_ARCH%%\.tar\.gz' \
+    | xargs -n1 curl -Ls \
+    | tar -xvz -C /usr/src/app/
+
+COPY requirements.txt /tmp/requirements.txt
+RUN pip install --no-cache-dir -r /tmp/requirements.txt
+
+COPY trigger_wifi_setup.py .
+
+RUN pip install git+https://github.com/larsks/dockerize
```
vpetersson, line 25: Avoid using anything that uses GPL3, as it creates a lot of spill-over issues.

wifi-connect/Dockerfile.template
nicomiguelino: Got it. Yes, I agree that it does. The possibility that the changes in this PR might not work on Raspbian is the reason behind creating #1699. Efforts on #1628 will still be on #1697.

wifi-connect/Dockerfile.template
nicomiguelino: Got it. I was planning on using [balena-os/wifi-connect][1], but it uses _Apache-2.0_, [which is not compatible with _GPL-2.0_][2] (the one that _Anthias_ uses).

Looks like a need to rework on some changes.

[1]: https://github.com/balena-os/wifi-connect
[2]: https://www.gnu.org/licenses/license-list.en.html#apache2

nicomiguelino: However, the current changes use [balena-labs-projects/wifi-connect][1], which doesn't explicitly include a license in its GitHub page, but uses [dockerize][2], which uses GPL-3.0 (as you've mentioned earlier). Hmmm. Will look again for alternatives.

[1]: https://github.com/balena-labs-projects/wifi-connect
[2]: https://github.com/larsks/dockerize

viewer.py
```diff
@@ -116,33 +118,68 @@ def send_current_asset_id_to_server():
...
+        template = Template(f.read())
+
+    context = {
+        'network': decoded.get('network', None),
+        'ssid_pswd': decoded.get('ssid_pswd', None),
+        'address': decoded.get('address', None),
+    }
+
+    with open('/tmp/hotspot.html', 'w') as out_file:
+        out_file.write(template.render(context=context))
```
nicomiguelino, line 144: @vpetersson @rusko124, I might need NGINX-related help on this one. I've managed to have `hotspot.html` written in `/tmp`, but what's left to do (apart from the license compatibility stuff) is to fix `nginx.conf`.

In the meantime, I'll read up (and/or do a very quick crash course) on writing NGINX config files. While the syntax might seem straightforward, there could be some catches.

wifi-connect/Dockerfile.template
vpetersson: Hmm We can't use `%%BALENA_ARCH%%` in the `Dockerfile` nor `Dockerfile.template` syntax in general. These are Balena specific, and we will not be able to build these in CI. We have a bit of custom logic to accomplish the same thing in our `[docker-build.yaml](https://github.com/Screenly/Anthias/blob/master/.github/workflows/docker-build.yaml#L56-L77)` file.

nicomiguelino: Oh, got it, thanks! I can do a POC on this one to see if making use of build args on Dockerfiles will work in Anthias. According to [this article](https://www.jeffgeerling.com/blog/2017/use-arg-dockerfile-dynamic-image-specification), it's possible.

[1]: https://www.jeffgeerling.com/blog/2017/use-arg-dockerfile-dynamic-image-specification

vpetersson: Yep, that's a valid alternative solution. Perhaps even cleaner than what I had. Whatever we do, we just need to be consistent across all container builds.

docker-compose.balena.yml
```diff
@@ -1,5 +1,19 @@
 version: "2"
 services:
+  srly-ose-wifi-connect:
+    image: screenly/srly-ose-wifi-connect:BOARD_TAG
+    build: ./wifi-connect # @TODO: Remove this.
```
nicomiguelino, line 5: Remove the `build` option since we'll be fetching the images from Docker Hub instead.

docker-compose.balena.yml
```diff
@@ -1,5 +1,19 @@
...
+    environment:
+      PORTAL_LISTENING_PORT: 8000
+      CHECK_CONN_FREQ: 10
+      PORTAL_SSID: 'Anthias WiFi Connect'
+    restart: always
+    network_mode: host
+    privileged: true
+    labels:
+      io.balena.features.dbus: "1"
+      io.balena.features.firmware: "1"
```
nicomiguelino, line 15: We might not be using these labels. Remove them if we ain't going to use them.

nicomiguelino:
I got the following error upon running `./bin/upgrade_containers.sh`:

```
➜  screenly git:(1699-wifi-connect) ./bin/upgrade_containers.sh
[+] Running 6/7
 ⠿ srly-ose-wifi-connect Warning                                                                                                                         1.7s
 ⠿ srly-ose-server Pulled                                                                                                                                1.7s
 ⠿ redis Pulled                                                                                                                                          1.7s
 ⠿ srly-ose-celery Pulled                                                                                                                                1.6s
 ⠿ srly-ose-viewer Pulled                                                                                                                                1.8s
 ⠿ srly-ose-nginx Pulled                                                                                                                                 1.6s
 ⠿ srly-ose-websocket Pulled                                                                                                                             1.7s
WARNING: Some service image(s) must be built from source by running:
    docker compose build srly-ose-wifi-connect
1 error occurred:
        * Error response from daemon: pull access denied for screenly/srly-ose-wifi-connect, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Just gonna leave this here for reference...

docker-compose.balena.yml
nicomiguelino: Remove the `:BOARD_TAG` part as well.

vpetersson:
I've created `srly-ose-wifi-connect`. CI Should be able to push to it.

docker-compose.yml
```diff
@@ -30,6 +46,8 @@ services:
       - PORT=80
       - NOREFRESH=1
       - LISTEN=srly-ose-nginx
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
vpetersson, line 50: Why do we need this?

docker/Dockerfile.wifi-connect
```diff
@@ -0,0 +1,43 @@
+# @TODO: Make use of build args or `sed` to customize the
+#     base image to be used.
+FROM balenalib/aarch64-python as builder
```
vpetersson, line 3: This is not a multi-stage build. Since it's not, please don't mix base images as that will add a lot of additional storage on the host.

docker-compose.yml
nicomiguelino: Ahh yes... It's necessary for the viewer service to access the ZeroMQ PUB socket at port 10001.
Since `srly-ose-wifi-connect` is in host networking mode, the viewer will not be able to refer to the `srly-ose-wifi-connect` service by its service name.

Alternatively, we can just use the default host gateway IP. I can't exactly remember, but it's something like `172.x.x.x`. That way, we can omit the changes.

nicomiguelino:
Efforts to make containerized `wifi-connect` service work with Raspbian are still in the works. Here's an output for reference:

```
screenly-srly-ose-wifi-connect-1  | Checking internet connectivity ...                                                                                       
screenly-srly-ose-wifi-connect-1  | Your device is not connected to the internet.                                                                            
screenly-srly-ose-wifi-connect-1  | Starting up Wifi-Connect.                                                                                                
screenly-srly-ose-wifi-connect-1  | Connect to the Access Point and configure the SSID and Passphrase for the network to connect to.                         
screenly-srly-ose-wifi-connect-1  | Cannot get the NetworkManager service state                                                                              
screenly-srly-ose-wifi-connect-1  | thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: D-Bus error: Failed to connect to socket /host/run/dbus/system
bus_socket: No such file or directory (org.freedesktop.DBus.Error.FileNotFound)', /checkout/src/libcore/result.rs:906:4                         
screenly-srly-ose-wifi-connect-1  | note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

docker-compose.balena.yml
nicomiguelino: Oops. Wrong file. This comment should be om `docker-compose.yml`.

docker/Dockerfile.wifi-connect
```diff
@@ -0,0 +1,32 @@
+FROM balenalib/BALENA_ARCH-python
```
nicomiguelino, line 1: If possible, make use of just `-python-run` instead of `-python`. More details could be found [here](https://hub.docker.com/r/balenalib/aarch64-python) for instance, specifically on **_Image Variants_**.

nicomiguelino: If possible, pin this to a specific tag like `3.7-buster`.

docker/wifi-connect/requirements.txt
```diff
@@ -0,0 +1 @@
+pyzmq==24.0.1
```
nicomiguelino, line 1: Change the version if necessary.

nicomiguelino: I've changed my base image to use Python 3.7 instead of 3.11, and using this version of `pyzmq` still works. Will leave the version as it is.

docker/Dockerfile.wifi-connect
nicomiguelino: The `-python-run` and `-python` images are of the same size, at least for the `3.7-buster` tag.

nicomiguelino:
@vpetersson, most of the changes are ready for review. I still have to test the changes thoroughly on Pi 3 and Pi 4.
I'm currently trying to make the splash page (with the device's IP address) appear upon reconnecting (via Wi-Fi). The feature isn't stable yet, I still have to fix the instability if necessary.

nicomiguelino:
@vpetersson, I've already tested this on a Pi 4, and it's working. Requesting for a review...

.github/workflows/docker-build.yaml
```diff
@@ -69,10 +71,14 @@ jobs:
               docker/Dockerfile.base
             sed -i 's/balenalib\/raspberrypi3/balenalib\/raspberry-pi2/' \
               docker/Dockerfile.viewer
+            sed -i 's/\bBALENA_ARCH\b/rpi/' \
```
vpetersson, line 74: Can we make this consistent with the structure in the other `Dockerfiles` instead?

.github/workflows/docker-build.yaml
```diff
@@ -69,10 +71,14 @@ jobs:
...
+              docker/Dockerfile.wifi-connect
             echo 'TARGET_PLATFORM=linux/arm/v6' >> $GITHUB_ENV
 
           # We use the same target for Pi 3 and Pi 4 for now.
           else
+            sed -i 's/\bBALENA_ARCH\b/armv7hf/' \
```
vpetersson, line 80: Same.

docker/Dockerfile.wifi-connect
```diff
@@ -0,0 +1,32 @@
+FROM balenalib/BALENA_ARCH-python:3.7-buster
```
vpetersson, line 1: This should probably be refactored into a multi-stage build.

.github/workflows/docker-build.yaml
nicomiguelino: I'm not sure if I understood this very well.
Did you mean like making `armv7fh` the default on the `Dockerfile.wifi-connect`, then just replacing the arch with `rpi` (using `sed`) for Raspberry Pi 1 and 2 devices?

image

image

Or, is it necessary to go an extra step by using a different related image?

vpetersson: Rework this to use `Dockerfile.base` as the base image instead. Otherwise we need to pull down a whole different image (`balenalib/armvhf-python:3.7-buster`) that will require a lot of more download and disk space.

After you've done this, rework it to use the same structure that we use for all the other images in [docker-build.yaml](https://github.com/Screenly/Anthias/blob/master/.github/workflows/docker-build.yaml). There's no point in reinventing the wheel here.

docker/Dockerfile.wifi-connect
nicomiguelino: Got it. For the second build stage, I can just use the base image that `screenly/srly-ose-base` is using, `balenalib/raspberypi3:buster`.

nicomiguelino:
I got the following output when testing the changes on a Raspberry Pi 3 Model B+ running a 64-bit Raspberry Pi OS Lite (Bullseye).

```
screenly-srly-ose-wifi-connect-1  | Checking internet connectivity ...
screenly-srly-ose-wifi-connect-1  | Your device is not connected to the internet.
screenly-srly-ose-wifi-connect-1  | Starting up Wifi-Connect.
screenly-srly-ose-wifi-connect-1  | Connect to the Access Point and configure the SSID and Passphrase for the network to connect to.
screenly-srly-ose-wifi-connect-1  | [network_manager::device:WARN] Undefined device type: 30
screenly-srly-ose-wifi-connect-1  | WiFi device: wlan0
screenly-srly-ose-wifi-connect-1  | [network_manager::dbus_api:ERROR] Get org.freedesktop.NetworkManager.AccessPoint::RsnFlags property failed on /org/freedesktop/NetworkManager/AccessPoint/562: wrong property type
screenly-srly-ose-wifi-connect-1  | Error: Getting access points failed
screenly-srly-ose-wifi-connect-1  |   caused by: D-Bus failure: Get org.freedesktop.NetworkManager.AccessPoint::RsnFlags property failed on /org/freedesktop/NetworkManager/AccessPoint/562: wrong property type
```

bin/start_wifi_connect.sh
```diff
@@ -0,0 +1,48 @@
...
+    then
+        freq=$CHECK_CONN_FREQ
+    else
+        freq=120
+fi
+
+
+sleep 5
+
+while [[ true ]]; do
```
vpetersson, line 15: This is an infinite loop. You want to make it check for if `IS_CONNECTED`'s condition and `exit 0` once it is connected.

.github/workflows/docker-build.yaml
```diff
@@ -69,10 +71,14 @@ jobs:
               docker/Dockerfile.base
             sed -i 's/balenalib\/raspberrypi3/balenalib\/raspberry-pi2/' \
               docker/Dockerfile.viewer
+            sed -i 's/\barmv7hf\b/rpi/' \
```
vpetersson, line 74: This shouldn't be necessary since you're using the base image (which we patch above).

.github/workflows/docker-build.yaml
```diff
@@ -69,10 +71,14 @@ jobs:
...
+              docker/Dockerfile.wifi-connect
             echo 'TARGET_PLATFORM=linux/arm/v6' >> $GITHUB_ENV
 
           # We use the same target for Pi 3 and Pi 4 for now.
           else
+            sed -i 's/\bBALENA_ARCH\b/armv7hf/' \
```
vpetersson, line 80: This shouldn't be necessary if we're using the regular base image.

.github/workflows/docker-build.yaml
```diff
@@ -61,6 +61,8 @@ jobs:
               docker/Dockerfile.base
             sed -i 's/balenalib\/raspberrypi3/balenalib\/raspberry-pi/' \
               docker/Dockerfile.viewer
+            sed -i 's/\barmv7hf\b/rpi/' \
```
vpetersson, line 64: This shouldn't be necessary since you're using the base image (which we patch above).

docker/Dockerfile.wifi-connect
```diff
@@ -0,0 +1,38 @@
+FROM screenly/srly-ose-base as builder
+
+WORKDIR /usr/src/app
+
+RUN curl https://api.github.com/repos/balena-io/wifi-connect/releases/latest -sL \
+    | grep -hoP 'browser_download_url": "\K.*armv7hf\.tar\.gz' \
+    | xargs -n1 curl -Ls \
+    | tar -xvz -C /usr/src/app/
+
+FROM balenalib/raspberrypi3:buster
```
vpetersson, line 10: Shouldn't this be `FROM screenly/srly-ose-base as builder` too?

vpetersson (changes requested):
Good start. Added some feedback.

.github/workflows/docker-build.yaml
nicomiguelino: Got it. Will remove this one.

.github/workflows/docker-build.yaml
nicomiguelino: Haven't tried the changes yet on Raspberry Pi 1 and 2 devices, but this line might be necessary.

Inside `Dockerfile.wifi-connect`, we can see that the latest `wifi-connect` archive will be downloaded.

https://github.com/nicomiguelino/Anthias/blob/bd4499e8c6f33aba0572055f2f406b7ac7a5c2a0/docker/Dockerfile.wifi-connect#L5-L8

```bash
RUN curl https://api.github.com/repos/balena-io/wifi-connect/releases/latest -sL \
    | grep -hoP 'browser_download_url": "\K.*armv7hf\.tar\.gz' \
    | xargs -n1 curl -Ls \
    | tar -xvz -C /usr/src/app/
```

Archive files that have `armv7hf` in their file name are for Raspberry Pi 3 and 4 devices, while those that have `rpi` are for Raspberry Pi 1 and 2 devices.

docker/Dockerfile.wifi-connect
nicomiguelino: Will use `screenly/srly-ose-base` as the base image. Thanks.

bin/start_wifi_connect.sh
nicomiguelino: Doing exiting (plus using `on-failure` for the restart policy) doesn't work as expected. The container doesn't start when the Pi boots up (or at least, when the Docker daemon starts). Using `always` or `unless-stopped` might not cut it.

ATM, I can do the following:
- `break` out of the `while` loop once connected
- Execute `/usr/bin/balena-idle` at the very end of the script. (`balena-idle` is just like a script running `sleep infinity`).
  - With this, it's idle, although it's still running.

nicomiguelino: Here's an output of using idling:

```
screenly-srly-ose-wifi-connect-1  | Checking internet connectivity ...
screenly-srly-ose-wifi-connect-1  | Your device is already connected to the internet.
screenly-srly-ose-wifi-connect-1  | Skipping setting up Wifi-Connect Access Point. Will check again in 10 seconds.
screenly-srly-ose-wifi-connect-1  | Idling...
```

.github/workflows/docker-build.yaml
nicomiguelino: Yeah, it's hardcoded ATM...

.github/workflows/docker-build.yaml
nicomiguelino: Same comment as with the other thread.

bin/start_wifi_connect.sh
```diff
@@ -0,0 +1,50 @@
...
+
+        IS_CONNECTED='false'
+
+        /usr/src/app/wifi-connect -u /usr/src/app/ui
+    fi
+
+    sleep $freq
+done
+
+/usr/bin/balena-idle
```
nicomiguelino, line 50: Can rm this one...

bin/install_wifi_connect.sh
```diff
@@ -0,0 +1,15 @@
...
+    architecture='armv7hf'
+fi
+
+wc_download_url='https://api.github.com/repos/balena-os/wifi-connect/releases/45509064'
+jq_filter=".assets[] | select (.name|test(\"linux-$architecture\")) | .browser_download_url"
+archive_url=$(curl -sL $wc_download_url | jq -r "$jq_filter")
+archive_file=$(basename $archive_url)
+
+wget $archive_url && \
+    tar -xvz -C /usr/src/app -f $archive_file
```
vpetersson, line 15: nitpicking but there's no reason for `&&` here as it in a a layer we throw away. It will only make debugging harder.

vpetersson (approved):

PR #1708, opened 2023-01-10
Update nginx.conf

ealmonte32:
Subnet for class A network 10.0.0.0 should be /8 (255.0.0.0) and not (255.255.0.0) /16 in order to support all internal private IP addresses for that class.

This was brought to my attention after a user on the forums reported [403 forbidden error](https://forums.screenly.io/t/403-forbidden-nginx-error-when-generating-backup/942/16).

image

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -77,7 +77,7 @@ server {
     }
     
     location /static_with_mime {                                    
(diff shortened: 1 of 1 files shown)


vpetersson (approved):

vpetersson:
Good catch @ealmonte32! Not sure how we ended up messing that up.

Issue #1703, opened 2023-01-03
Finish rebranding


vpetersson:
* Update all references from Screenly OSE to Anthias in the UI (and installer)
* Update the design to use the new design from the web interface from @L4cko

PR #1715, opened 2023-01-20
Rename service/container names so that it will use `anthias-` instead of `srly-ose`.

nicomiguelino:
#### Overview

- Related issues &ndash; #1703 

##### Installing changes on `master` on a clean instance

```bash
cd && \
echo "export CUSTOM_BRANCH='master'" > .env && \
```

##### Installing changes on an existing installation

```bash
cd && \
echo "export CUSTOM_BRANCH='1703-rename-services'" > .env && \
```

### DIFF
--- a/docker-compose.balena.yml
+++ b/docker-compose.balena.yml
@@ -1,6 +1,6 @@
 version: "2"
 services:
-  srly-ose-wifi-connect:
+  anthias-wifi-connect:
     image: screenly/srly-ose-wifi-connect:BOARD_TAG
     environment:
       PORTAL_LISTENING_PORT: 8000
@@ -14,7 +14,7 @@ services:
       io.balena.features.dbus: "1"
       io.balena.features.firmware: "1"
 
-  srly-ose-server:
+  anthias-server:
     image: screenly/srly-ose-server:BOARD_TAG
     environment:
       - HOME=/data
@@ -29,15 +29,15 @@ services:
     labels:
       io.balena.features.supervisor-api: '1'
 
-  srly-ose-viewer:
+  anthias-viewer:
     image: screenly/srly-ose-viewer:BOARD_TAG
     depends_on:
-      - srly-ose-server
+      - anthias-server
     environment:
       - HOME=/data
       - PORT=80
       - NOREFRESH=1
-      - LISTEN=srly-ose-nginx
+      - LISTEN=anthias-nginx
     extra_hosts:
       - "host.docker.internal:host-gateway"
     privileged: true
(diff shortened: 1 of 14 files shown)


docker-compose.balena.yml
```diff
@@ -29,39 +29,42 @@ services:
...
     restart: always
     volumes:
       - resin-data:/data
     labels:
       io.balena.features.supervisor-api: '1'
 
-  srly-ose-websocket:
-    image: screenly/srly-ose-websocket:BOARD_TAG
+  anthias-websocket:
+    image: screenly/anthias-ose-websocket:BOARD_TAG
```
vpetersson, line 54: `anthias-ose` -> `anthias`.

docker-compose.balena.yml
```diff
@@ -29,39 +29,42 @@ services:
...
       - HOME=/data
       - LISTEN=0.0.0.0
     restart: always
     volumes:
       - resin-data:/data
 
-  srly-ose-celery:
-    image: screenly/srly-ose-celery:BOARD_TAG
+  anthias-celery:
+    image: screenly/anthias-ose-celery:BOARD_TAG
```
vpetersson, line 65: `anthias-ose` -> `anthias`.

docker-compose.yml
```diff
@@ -114,12 +117,21 @@ services:
...
-      - srly-ose-server
-      - srly-ose-websocket
+      - anthias-server
+      - anthias-websocket
+    networks:
+      default:
+      anthias-extra-1:
+        aliases:
+          - 'anthias-nginx'
+          - 'srly-ose-nginx'
```
vpetersson, line 127: Why do we need this line? The whole file should be transactional so there should be no `srly-ose-nginx` available after this update.

docker-compose.yml
```diff
@@ -38,48 +38,51 @@ services:
...
       - HOME=/data
       - PORT=80
       - NOREFRESH=1
-      - LISTEN=srly-ose-nginx
+      - LISTEN=anthias-nginx
     extra_hosts:
       - "host.docker.internal:host-gateway"
+    networks:
+      default:
+      anthias-extra-1:
```
vpetersson, line 57: I'm confused about this part. Why do we need this?

docker-compose.yml
nicomiguelino: these are only temporary, and will be removed when there's a new build of the webview (given that there's a change in view.cpp)

docker-compose.yml
nicomiguelino: this is (temporarily) needed so that the viewer service can recognize the NGINX service as `anthias-nginx` or `srly-ose-nginx`

vpetersson (approved):
Approved pending that it has been tested on an existing installation.

PR #1716, opened 2023-01-23
Build x86 docker images

vpetersson:
* Removes old `Dockerfile.X.dev` as they are no longer needed
* Introduces x86 docker image builds on CI


nicomiguelino (commented):
The following Dockerfiles are still required by the unit test job:

- `Dockerfile.base.dev`
- `Dockerfile.test`
- `Dockerfile.celery.dev`

There are three options (so far) to fix the unit test job:

- Restore the Dockerfiles mentioned above.
- Inside `docker-compose.test.yml`, remove the build options from each of the defined services (except for the one that uses Dockerfile.base.dev).
  - One drawback of this approach is the developer experience. If the developer modifies the source code locally, instant feedback would not be possible. We still have for the next image build (during a PR merge).
  - Make use of the non-dev base images (both locally and CI-wise).
    - This would be the least intrusive approach among the three.
    - We can just do `sed` on the relevant Dockerfiles before running the tests.

vpetersson:
There's actually a more elegant solution outlined [here](https://github.com/containers/buildah/issues/2404). Basically, we can modify our `Dockerfile`'s to use an argument instead, like this:

```
ARG FROM=balenalib/raspberrypi3:buster
FROM $FROM
```

Then, we can simply append `--build-arg` to override this. That has the following benefits:

* We can reduce all the `sed` logic in the GitHub Actions flow
* We can have the same `Dockerfile`s for x86 and Pis

PR #1733, opened 2023-02-10
Update upgrade_containers.sh and install.sh

ealmonte32:
Tested for days now and after increasing `mem_limit` for the viewer container to `.80` from `.70`, this change prevented the Pi from freezing when resource-heavy websites were added.
I dont know the magic number we should be using here, but we can't go lower because it literally created more problems since it didn't give enough memory to the viewer, and making it higher solved an issue I was seeing, so this is a percentage I am comfortable with.


vpetersson:
Not loving this fix but I'll accept it until we have a better fix in place.

vpetersson (approved):

ealmonte32:
> Not loving this fix but I'll accept it until we have a better fix in place.

I hear you... but the 80% seems to work better, and I am stressing it even more by adding more heavy websites that load news, and with the 70% it would be frozen by now.

ealmonte32:
@vpetersson 
I still cant merge due to no write access.

PR #1742, opened 2023-02-17
Add command-line support for skipping services when building images.

nicomiguelino:
#### Overview

- Add support for handling command-line arguments.


vpetersson:
Reviewing this code, I'm thinking we're over complicating things. I know that I said we should convert the service list to an array, but maybe we're over complicating it. What if we instead do something like this:

```bash

if [ -n ${SKIP_VIEWER}" ]; then
   continue
fi
```

We'd need to generalize the code to:
* Convert the container name to upper case (something like `CONTAINER_UPPER_CASE="$(echo $container | tr '[:lower:]' '[:upper:]')"` would do 
* Build a variable on the fly for `SKIP_$CONTAINER_UPPER_CASE` in the loop

The latter part is a bit tricky in Bash, but pretty sure it's doable.

vpetersson (approved):
Great cleanup, but let's also skip the unrelated containers for the test phase while we are doing this.

nicomiguelino:
> Great cleanup, but let's also skip the unrelated containers for the test phase while we are doing this.

Good catch @vpetersson, thanks! I'll edit the workflow file accordingly.

PR #1757, opened 2023-02-27
Add support for generating only Dockerfiles when running `bin/build_containers.sh`.

nicomiguelino:
#### Overview

Here's the sample usage of the script:

```bash
DOCKERFILES_ONLY=1 ./bin/build_containers.sh
```

The variable `DOCKERFILES_ONLY` should be set to a non-zero number. Otherwise, the script will still build the containers.


vpetersson (approved):
Lgtm

PR #1778, opened 2023-03-14
Fix `wifi-connect` issues.

nicomiguelino:
#### Overview

- Fixes #1746
- Fixes #1747

#### Tasks

- [x] Fix relevant blocking issues.
- [x] The `anthias-wifi-connect` service must exit early if no Wi-Fi devices are found, esp. for Pi 1 and Pi 2 devices.
- [x] Create a `systemd` unit file on RPi OS so that the `anthias-wifi-connect` service will start during startup.
- [x] Add documentation on how to configure Wi-Fi on the device.
- [x] Add a disclaimer in the docs that it's not yet possible to customize the access point settings &ndash; e.g., SSID, password, and gateway (the IP address and port, which defaults to 80).

#### Custom Installation Command (Testing)

```bash
cd && \
echo "export CUSTOM_BRANCH='wifi-fix'" > .env && \
bash <(curl -sL https://raw.githubusercontent.com/nicomiguelino/Anthias/custom-install/bin/install.sh)

PR #1786, opened 2023-03-29
Fix issues with the image build CI action.

nicomiguelino:
#### Overview

- Follow-up fixes to #1778

##### Installation

```bash
cd && \
echo "export CUSTOM_BRANCH='hf-wifi'" > .env && \
```

##### Errors

```
#16 [stage-0  7/12] RUN curl -sL -o /tmp/wifi_connect.zip "" &&     unzip -o /tmp/wifi_connect.zip -d /usr/src/app &&     rm /tmp/wifi_connect.zip
#16 ERROR: process "/bin/sh -c curl -sL -o /tmp/wifi_connect.zip \"\" &&     unzip -o /tmp/wifi_connect.zip -d /usr/src/app &&     rm /tmp/wifi_connect.zip" did not complete successfully: exit code: 3
------
 > [stage-0  7/12] RUN curl -sL -o /tmp/wifi_connect.zip "" &&     unzip -o /tmp/wifi_connect.zip -d /usr/src/app &&     rm /tmp/wifi_connect.zip:
------
Dockerfile.wifi-connect:60
--------------------
  59 |     
  60 | >>> RUN curl -sL -o /tmp/wifi_connect.zip "" && \
  61 | >>>     unzip -o /tmp/wifi_connect.zip -d /usr/src/app && \
  62 | >>>     rm /tmp/wifi_connect.zip
  63 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c curl -sL -o /tmp/wifi_connect.zip \"\" &&     unzip -o /tmp/wifi_connect.zip -d /usr/src/app &&     rm /tmp/wifi_connect.zip" did not complete successfully: exit code: 3
Error: Process completed with exit code 1.
```

##### When changes were applied

When running `docker compose logs -f anthias-wifi-connect` on a Raspberry Pi 2 running 32-bit RPi OS Lite (Bullseye)...

```
screenly-anthias-wifi-connect-1  | No Wi-Fi adapters were detected. Exiting...
```

#### Failed jobs (for reference)

- https://github.com/Screenly/Anthias/actions/runs/4539699386/jobs/8016045429
- https://github.com/Screenly/Anthias/actions/runs/4539699386/jobs/8015497704

#### Relevant Forums

- https://forums.balena.io/t/no-archive-file-for-rpi-devices-for-wifi-connect-4-11-1/367302

### DIFF
--- a/bin/build_containers.sh
+++ b/bin/build_containers.sh
@@ -46,7 +46,7 @@ elif grep -qF "Raspberry Pi 3" /proc/device-tree/model || [ "${BUILD_TARGET}" ==
 elif grep -qF "Raspberry Pi 2" /proc/device-tree/model || [ "${BUILD_TARGET}" == 'pi2' ]; then
     export BOARD="pi2"
     export BASE_IMAGE=balenalib/raspberry-pi2
-    export TARGET_PLATFORM=linux/arm/v7
+    export TARGET_PLATFORM=linux/arm/v6
 elif grep -qF "Raspberry Pi 1" /proc/device-tree/model || [ "${BUILD_TARGET}" == 'pi1' ]; then
     export BOARD="pi1"
     export BASE_IMAGE=balenalib/raspberry-pi
--- a/bin/build_containers.sh
+++ b/bin/build_containers.sh
@@ -46,7 +46,7 @@ elif grep -qF "Raspberry Pi 3" /proc/device-tree/model || [ "${BUILD_TARGET}" ==
 elif grep -qF "Raspberry Pi 2" /proc/device-tree/model || [ "${BUILD_TARGET}" == 'pi2' ]; then
     export BOARD="pi2"
     export BASE_IMAGE=balenalib/raspberry-pi2
-    export TARGET_PLATFORM=linux/arm/v6
+    export TARGET_PLATFORM=linux/arm/v7
 elif grep -qF "Raspberry Pi 1" /proc/device-tree/model || [ "${BUILD_TARGET}" == 'pi1' ]; then
     export BOARD="pi1"
     export BASE_IMAGE=balenalib/raspberry-pi
--- a/bin/start_wifi_connect.sh
+++ b/bin/start_wifi_connect.sh
@@ -2,7 +2,12 @@
 
 IS_CONNECTED=''
 
-if [[ $DEVICE_TYPE =~ ^(pi1|pi2)$ ]]; then
+if [[ $DEVICE_TYPE = "pi1" ]]; then
+    echo "The latest version of WiFi Connect does not support Raspberry Pi 1. Exiting..."
+    exit 0
+fi
+
+if [[ $DEVICE_TYPE == "pi2" ]]; then
     if [[ -z $(iw dev | grep Interface) ]]; then
         echo "No Wi-Fi adapters were detected. Exiting..."
         exit 0
--- a/bin/start_wifi_connect.sh
+++ b/bin/start_wifi_connect.sh
@@ -1,17 +1,15 @@
 #!/bin/bash
 
+# vim: tabstop=4 shiftwidth=4 softtabstop=4
+# -*- sh-basic-offset: 4 -*-
+
+set -euox pipefail
+
 IS_CONNECTED=''
 
-if [[ $DEVICE_TYPE = "pi1" ]]; then
-    echo "The latest version of WiFi Connect does not support Raspberry Pi 1. Exiting..."
+if [[ -z $(nmcli device wifi list) ]]; then
+    echo "No Wi-Fi adapters were detected. Exiting..."
     exit 0
-fi
-
-if [[ $DEVICE_TYPE == "pi2" ]]; then
-    if [[ -z $(iw dev | grep Interface) ]]; then
-        echo "No Wi-Fi adapters were detected. Exiting..."
-        exit 0
-    fi
 fi
 
 if [[ ! -z $CHECK_CONN_FREQ ]]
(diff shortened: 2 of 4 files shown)


docker/Dockerfile.wifi-connect.tmpl
```diff
@@ -10,9 +10,10 @@ RUN --mount=type=cache,target=/var/cache/apt \
 
 WORKDIR /usr/src/app
 
-RUN curl -sL -o /tmp/wifi_connect.zip "$ARCHIVE_URL" && \
+RUN if [[ ! -z "$ARCHIVE_URL" ]]; then \
+    curl -sL -o /tmp/wifi_connect.zip "$ARCHIVE_URL" && \
     unzip -o /tmp/wifi_connect.zip -d /usr/src/app && \
-    rm /tmp/wifi_connect.zip
+    rm /tmp/wifi_connect.zip; fi
```
rusko124, line 16: I follow that will fix the build, but does wifi-connect work if we miss this part?

nicomiguelino: That `RUN` statement is a one-line if-statement. I need to add a check if the `wifi-connect` archive URL is blank (for Pi 1 and Pi 2 devices), otherwise the build process will never proceed past this point.

nicomiguelino: To add, the latest version of WiFi Connect (`4.11.1` at the time) doesn't have archive downloads for ARM v6 devices (i.e., the Raspberry Pi 1).

rusko124: What I mean is that we should not execute this Dockerfile for pi1/pi2 at all(Can we do it?). Because in this case, we have just extra building steps that don't make sense for pi1/pi2, right?

nicomiguelino: At the moment, yes. It makes sense not to execute the Dockerfile for Pi 1 & Pi 2.
However, it might be possible to plug like USB Wi-Fi adapters into those devices.

I'll see what I can do, thanks!

bin/build_containers.sh
```diff
@@ -46,7 +46,7 @@ elif grep -qF "Raspberry Pi 3" /proc/device-tree/model || [ "${BUILD_TARGET}" ==
 elif grep -qF "Raspberry Pi 2" /proc/device-tree/model || [ "${BUILD_TARGET}" == 'pi2' ]; then
     export BOARD="pi2"
     export BASE_IMAGE=balenalib/raspberry-pi2
-    export TARGET_PLATFORM=linux/arm/v6
+    export TARGET_PLATFORM=linux/arm/v7
```
vpetersson, line 49: I'm fairly sure this is wrong.

vpetersson (changes requested):

bin/build_containers.sh
nicomiguelino: Got it, I'm curious that I get `armv7l` when I run `uname -m` on a Raspberry Pi 2 B running 32-bit RPi OS.

nicomiguelino: I already reverted the changes on this one since ARMv7 is backward compatible.

nicomiguelino: Just read your recent comment. Interesting, indeed. It shows `armv7l` on both RPi OS and balenaOS instances (for Pi 2).
For Pi 1, it shows something like `armv6`. Don't have a personal Pi 1 yet, so I'm not able to check it out for the time being.

Let me take a closer look again on those balena instances.

nicomiguelino: @vpetersson, @rusko124
Here's what I got from a Pi 1 running balenaOS:

image

Here's what I got for Pi 2:

image

nicomiguelino: Here's some balenaOS logs (`anthias-wifi-connect` only):

image

bin/start_wifi_connect.sh
```diff
@@ -2,7 +2,12 @@
 
 IS_CONNECTED=''
 
-if [[ $DEVICE_TYPE =~ ^(pi1|pi2)$ ]]; then
+if [[ $DEVICE_TYPE = "pi1" ]]; then
+    echo "The latest version of WiFi Connect does not support Raspberry Pi 1. Exiting..."
```
vpetersson, line 6: WiFi connect doesn't care what Pi it is running on. It cares if it has a WiFi adapter. You can easily add this using USB. Maybe it's a fine temporary workaround to use this logic, but the *proper* fix is to check if the device has a WiFi interface. 

You can use `nmcli` to detect this:

Device without WiFi adapter:
```bash
$ nmcli device wifi list
$ 
```

Device with WiFi adapter:
```bash
$ nmcli device wifi list
IN-USE  BSSID              SSID                  MODE   CHAN  RATE        SIGNAL 
[...]
```

Since we already rely on Network Manager on both Raspbian and Balena, that seems like a reasonable dendency.

nicomiguelino: I agree, thanks. That seems pretty neat.

nicomiguelino:
Hello, @vpetersson. Requesting another review...

image

docker/Dockerfile.wifi-connect.tmpl
vpetersson: This block needs to be reverted now with the `nmcli` change above.

vpetersson (changes requested):
Need to undo the if statement in `Dockerfille.wifi-connect.tmpl`.

vpetersson (approved):

PR #1792, opened 2023-04-09
Fix byte-encoding issue that occurs when attempting to download an asset.

### LINKED ISSUE #1790: Unable to download assets on Raspberry Pi 4 running `balenaOS`.

nicomiguelino:
#### Overview

- Fixes #1790

### DIFF
--- a/server.py
+++ b/server.py
@@ -1537,7 +1537,7 @@ class AssetContent(Resource):
             result = {
                 'type': 'file',
                 'filename': filename,
-                'content': b64encode(content),
+                'content': b64encode(content).decode(),
                 'mimetype': mimetype
             }
         else:


server.py
```diff
@@ -1537,7 +1537,7 @@ def get(self, asset_id):
             result = {
                 'type': 'file',
                 'filename': filename,
-                'content': b64encode(content),
+                'content': b64encode(content).decode(),
```
rusko124, line 1540: looks aggressive to me. We definitely need a unit test for this.

vpetersson (approved):

server.py
nicomiguelino: Got it. I'll definitely create an issue and a PR for this one. Thanks!

PR #1797, opened 2023-04-13
Set restart policy of the `anthias-wifi-connect` service to `on-failure`.

### LINKED ISSUE #1796: Set the restart policy of the `anthias-wifi-connect` service to `on-failure` for devices running balenaOS.
#### Overview

- See the discussion about the unexpected restart loop on Screenly/Anthias#1746 for details.

nicomiguelino:
#### Overview

- Fixes #1796

### DIFF
--- a/docker-compose.balena.yml.tmpl
+++ b/docker-compose.balena.yml.tmpl
@@ -13,6 +13,7 @@ services:
       DBUS_SYSTEM_BUS_ADDRESS: 'unix:path=/host/run/dbus/system_bus_socket'
     network_mode: host
     privileged: true
+    restart: on-failure
     labels:
       io.balena.features.dbus: "1"
       io.balena.features.firmware: "1"


nicomiguelino:
@vpetersson, here are the statuses of the services after deploying the changes to my own fleet.

image

I checked the `anthias-wifi-connect ` logs, and it seems that the service is not restarting every boot.
The only time that it was started is when the changes were deployed (for the first time).

vpetersson:
> Would it be better to start the service every device boot?

Yes, if you're able to prevent a restart loop.

nicomiguelino:
Okay, thanks. One drawback of using `on-failure` is that the service will not start again if it exits with a zero (successful).

It is possible to start `anthias-wifi-connect` service from the viewer service via the Supervisor API.

vpetersson:
> Okay, thanks. One drawback of using `on-failure` is that the service will not start again if it exits with a zero (successful).

That's the desired behavior.
 
> It is possible to start `anthias-wifi-connect` service from the viewer service via the Supervisor API.

That's a good point. However, I'm curious what the behavior is on Raspbian OS. Does it start once per boot, or only once in total?

nicomiguelino:
The behavior in RPi OS is different from that in balenaOS.

In RPi OS, if the restart policy is set to default (no), it will never restart again, unless we'll explicitly start it. I've created a systemd service to start the anthias-wifi-connect service during startup.

vpetersson:
wifi-connect has been sunset.

PR #1808, opened 2023-04-28
Modify the build script so that it will work seamlessly on dev mode.

nicomiguelino:
#### Overview

- Running `./bin/build_containers.sh` on an x86 machine (more specifically, on development mode) wouldn't work smoothly, so I've created a PR.

#### Upcoming changes

- Docs update for developers


nicomiguelino:
We already have a set of scripts for making the Docker image build script in development mode:

https://github.com/Screenly/Anthias/blob/ec67066873290eb1b62ce5c36003ac5f47ba5532/bin/start_development_server.sh#L1-L12

https://github.com/Screenly/Anthias/blob/ec67066873290eb1b62ce5c36003ac5f47ba5532/bin/generate_dev_mode_dockerfiles.sh#L1-L27

PR #1819, opened 2023-06-06
Edit docs so that it will have info whether to use 32-bit or 64-bit OS.

nicomiguelino:
#### Overview

- For Raspberry Pi versions 3 and 4, it's recommended to install 64-bit RPi OS Lite.
- For Raspberry Pi Zero, 1, and, 2, use 32-bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,12 @@ If you'd like more control over your digital signage instance, try installing it
 
 ## Installing on Raspberry Pi OS Lite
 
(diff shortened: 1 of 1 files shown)


nicomiguelino:
There's already a Compatibility section in [README.md](https://github.com/Screenly/Anthias/blob/ec67066873290eb1b62ce5c36003ac5f47ba5532/README.md#white_check_mark-compatibility), which include devices and setup that are compatible with Anthias.

PR #1835, opened 2023-10-23
Fixes the Anthias installer so that it'll work on Debian Bookworm as well

nicomiguelino:
#### Overview

- Fixes Screenly/Anthias#1834


nicomiguelino:
I've set the target branch to a wrong one before confirming the pull request. I'll just create a new one.

PR #1861, opened 2024-03-14
Modify release workflow to have tag/release version as required input

nicomiguelino:


vpetersson (changes requested):
No, tag should be the git tag for a release.

nicomiguelino:
@vpetersson,

Got it. The release step doesn't run successfully if no tags were specified.
For instance, in this [CI run](https://github.com/Screenly/Anthias/actions/runs/8279539933/job/22654166862):
- "Error: Error undefined: No tag found in ref or input!"
- The release action works in such a way that if no tags were found, a new one will be created.

It can also be seen in [this change](https://github.com/Screenly/Anthias/commit/b846243ece9d7d1a04b48c50055fb236e2d0ccac) that the tags property was removed. (See screenshot below for details).


image

vpetersson (approved):

PR #1888, opened 2024-04-23
Migrates from Qt 5 to Qt 6

nicomiguelino:
#### References

- https://wiki.qt.io/Cross-Compile_Qt_6_for_Raspberry_Pi
- https://github.com/ninja-build/ninja/releases
- https://doc.qt.io/qt-6/linux-building.html


nicomiguelino:
Raspberry Pi 5 and x86 devices already uses Qt 6. Efforts for using Qt 6 for the rest of the supported devices (if possible) will be done in separate pull requests.

PR #1890, opened 2024-04-25
Fixes viewer to support RTSP playback

nicomiguelino:
#### Related issues

* Screenly/Anthias#1638

#### Limitations and issues

* When an RTSP asset is added and `Skip asset check` is unchecked, it will be added successfully, but the playback stops after 5 minutes (the default duration for RTSP video assets).
* It will take a while for the device to display the stream again (~10s).


nicomiguelino:
Significant changes has been merged into `master`. This PR only includes workarounds at the moment. A new pull request solving the RTSP issue needs to be created.

PR #1896, opened 2024-05-20
Fixes: Anthias is not starting asset (gets stuck at the logo screen)

### LINKED ISSUE #1882: Anthias is not starting asset (gets stuck at the logo screen)
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

#### Overview of the Issue

- https://forums.screenly.io/t/anthias-not-starting-asset/1245
- https://forums.screenly.io/t/anthias-not-starting-asset/1166

#### Reproduction Steps

Steps to reproduce this issue, eg:

* Press Add Asset
* Enter the Asset URL ...
* Press "On"
* Wait for the asset to show up on the screen

#### Environment

* Raspberry Pi Hardware Version:
* Raspberry Pi Network Setup: (ex. built-in WiFi, Ethernet, USB WiFi Adapter)
* Anthias Version:

nicomiguelino:
#### Description

* Fixes #1882
* A callback was used in replacement of the single-shot timer used for rendering HTML in the webview.


docker/nginx/nginx.conf
```diff
@@ -71,7 +71,9 @@ server {
     }
 
     location /screenly_assets {
```
vpetersson, line 73: This doesn't look right. IIRC this endpoint should only be accessible from the local device.

vpetersson:
I'm a bit confused about this PR. If the purpose is to downscale an image, I don't think this is the right approach. We don't want all assets to be accessible over the network. If the objective is to downscale the images, then an async celery task or similar is a better choice.

nicomiguelino:
@vpetersson The reason I transferred the HTML logic of showing the image is because the first `loadImage` call (from the viewer to the webview) takes a while to finish, and that it doesn't show the image.

docker/nginx/nginx.conf
nicomiguelino: Given that it's a constraint that we should consider, then I should focus on editing the C++ code instead and not exposing the asset files in NGINX.

nicomiguelino:
> Sure, but that's still a synchronous process. Image scaling should really happen when the asset is uploaded (or rather asynchronously in the background after).

Agree, asset optimization (in this case, image scaling) can be done  during the asset upload.

vpetersson:
> > Sure, but that's still a synchronous process. Image scaling should really happen when the asset is uploaded (or rather asynchronously in the background after).
> 
> Agree, asset optimization (in this case, image scaling) can be done during the asset upload.

I also don't think we should ever upscale images -- just down scale.

docker/nginx/nginx.conf
vpetersson: I don't think any C++ code edits are needed. You just need to introspect and analyze how the traffic is flowing and perhaps provide a slightly wider policy to cover all possible Docker IPs.

nicomiguelino: Thanks. After investigating thoroughly, it appears that the D-Bus messages are going through.

nicomiguelino:
@vpetersson or @rusko124, please review.

nicomiguelino:
As of 5d009ed, I was not able to replicate #1882 on my Pi 2, Pi 3, and PI 4 devices.
Back in dc1f963, the issue persists most of the time.

I'll be reopening this PR is the issue persists again.

nicomiguelino:
Based on [the most recent comment in this forum topic](https://forums.screenly.io/t/after-booting-the-pi-if-the-asset-set-is-a-single-image-it-doesnt-display-the-screen-remains-dark-with-only-the-small-purple-screenly-logo-in-the-center/1166/25?u=nicomiguelino), the issue still occurs. I'll be re-opening this PR.

vpetersson (approved):
Approved without review

PR #1931, opened 2024-06-14
Replaces OMXPlayer with Qt (C++)

nicomiguelino:
#### Description

* Qt (via [QMediaPlayer](https://doc.qt.io/qt-5/qmediaplayer.html)) will replace OMXPlayer for playing videos.
* This might fix the black gap between active video assets.


nicomiguelino:
Not all devices support video playback via Qt properly. We could create another pull request that implements similar changes but only targets certain platforms that supports successful playback via Qt.

PR #1940, opened 2024-06-21
Enables HTTPS in Anthias

### LINKED ISSUE #402: Switch to HTTPS as the default port binding
With recent Raspberry Pis being powerful, it's no longer very expensive to run Nginx as a reverse proxy. This is also a more standardized way of serving web apps, so let's do this.

Here's what we need to do:

## Ansible

 * Remove `enable-ssl` from the SKIP_TAGS
 * Remove `/bin/enable_ssh.sh`

## App

 * Change the Gunicorn binding in server.py to bind on a UNIX socket (`/var/run/screenly.socket`)
 * Remove the listen line in `screenly.conf`

## Nginx

 * Re-configure Nginx to listen on listen on 8080 as well (80, 443, 8080)
 * Reconfigure Nginx the's 'ose' upstream to use the UNIX socket
 * Add a forwarding rule to redirect :8080 -> :443

nicomiguelino:
#### Useful References

* https://forums.screenly.io/t/anyone-have-issues-with-latest-builds-and-ssl/951/4

#### Description

- Might fix #402


vpetersson:
I don't think this is a good idea. Shipping self signed certs goes against most modern best practices. Instead we should probably make an integration with acme.sh and/or Tailscale for getting a real certificate.

nicomiguelino:
@vpetersson, I can update this branch and then add a feature that lets users upload their own certificates.
I'm still figuring out how to do the same for devices running BalenaOS.

vpetersson:
I kinda want something more intelligent. Ideally something that can easily be integrated with Let's Encrypt and/or Tailscale.

vpetersson:
Closing as superseded.

HTTPS support landed via a different design: `bin/enable_ssl.sh` installs an **anthias-caddy** sidecar as a compose override, with three modes — Caddy local CA by default, automatic Let's Encrypt with `--domain`, or BYO cert via `--cert`/`--key`. That sidesteps the `# @TODO: Find a way to generate the cert and key files ad-hoc` problem flagged in this diff.

The target files are also gone — the nginx + websocket containers were collapsed into uvicorn (#2757), so `docker/Dockerfile.nginx.tmpl` and `docker/nginx/nginx.conf` no longer exist. Linked issue #402 was closed today.

PR #1947, opened 2024-06-28
Cleans up upgrade (via web UI) code

nicomiguelino:
#### Description

* Upgrading Anthias via the web UI isn't operational, thus having the code removed.


vpetersson:
Please rename all files named screenly to anthias.

vpetersson (approved):
Approved pending renames of `screenly.css` and `screenly.css.map`.

PR #1950, opened 2024-07-01
Fixes IP address on splash screen not updating

### LINKED ISSUE #1766: IP address previewed on the splash screen doesn't update
Recently i have installed several raspberry pi's with Anthias. During installation i noticed that the IP address set during installation and setting up Anthias is stored in the splash screen, however the displayed IP address never updates.  When changing the device configuration to a fixed IP address or switching to WiFi doesn't effect the displayed IP address during startup. Is there a way to fix this?

#### Environment

* Tested on Raspberry Pi Hardware Version: 4, 3B+ and 3B
* Raspberry pi OS: Bullseye Lite 32 bit
* Raspberry Pi Network Setup: Both wifi and wired 
* Anthias Version: master@df00592

nicomiguelino:
#### Description

* Fixes #1766

#### Screenshot

image


host_agent.py
```diff
@@ -5,17 +5,33 @@
...
 # Name of redis channel to listen to
 CHANNEL_NAME = b'hostcmd'
 
+
+def set_ip_addresses():
+    rdb = redis.Redis(**REDIS_ARGS)
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    s.connect(("8.8.8.8", 80))
+    ip_address = s.getsockname()[0]
+    rdb.set('ip_addresses', json.dumps([ip_address]))
```
vpetersson, line 26: What happens if the device is dual homed (IPv6 + IPv4)?

vpetersson (changes requested):

vpetersson (approved):

vpetersson:
Nice. Looks good.

PR #1960, opened 2024-07-05
Renames the host virtual environment directory from `installer_venv` to `anthias_venv`

nicomiguelino:
#### Description

* The rename signifies that the virtual environment will primarily contain the dependencies needed by the host, especially the helper scripts.


vpetersson (changes requested):
Maybe it makes sense to switch to Poetry instead if we are going to touch this?

nicomiguelino:
Closing this pull request in favor of #1969

PR #1965, opened 2024-07-08
Fix: Making Anthias work without Internet

### LINKED ISSUE #1964: Request: Make Anthias work without internet
#### Overview of the Issue

Related to Issue #1494, Issue #1773 and Issue #1949.

Some time ago, I noticed that Anthias won't run without internet access. After closer inspection, it became clear that TCP 443 access to api.github.com is needed. Without this access, the web interface of Anthias will become unstable and sometimes respond with nginx http error 504.
My workaround to this was to enable TCP 443 access to this server in our firewall.

After trying to update to the latest anthias-version, I noticed that this behaviour came back. Everything OK with internet-access allowed on TCP 443, but without, the web interface responds with error, and also Anthias doesn't show assets right anymore. I have assets switch every 10 seconds, but without internet access, the asset will only switch after about 3 to 5 minutes, not every 10 seconds.

I tried to log DNS requests and see what Anthias wants to connect to. At first sight, it looks like access to hub.docker.com is now also needed for stable operation. I enabled TCP 443 traffic to hub.docker.com in our firewall and so far, it looks like its working stable now. Of course I need to watch it for some days to see if it is really stable.

Can you confirm that right now, TCP 443 access to api.github.com and hub.docker.com is needed, and otherwise Anthias will have problems running stable?
Are there plans to remove this need for internet access?
Or maybe you can write in the installation documentation that access to certain external hosts is needed for stable operation and provide a up-to-date list of all external hosts that need to be accessible?

In Issue #1949, @nicomiguelino wrote that it should be possible to use Anthias on a network without internet access, and that web assets that are accessible internally, should be displayed without problems. Unfortunately, that's not working for me.

#### Reproduction Steps

Running latest Anthias without internet access.

nicomiguelino:
#### Description

* Fixes #1964 
* Modifies `is_up_to_date` so that it will fail gracefully.

### DIFF
--- a/lib/github.py
+++ b/lib/github.py
@@ -43,9 +43,10 @@ def handle_github_error(exc, action):
 def is_reachable(domain_name):
     try:
         host = socket.gethostbyname(domain_name)
-        s = socket.create_connection((host, 443), 2)
+        s = socket.create_connection((host, 443), timeout=0.5)
         s.close()
         logging.info('Could reach domain: %s', domain_name)
+        return True
     except (socket.gaierror, OSError):
         logging.error('Could not reach domain: %s', domain_name)
         return False
@@ -168,8 +169,8 @@ def is_up_to_date():
     Returns True if the player is up to date.
     """
 
-    if not is_reachable('github.com') or not is_reachable('hub.docker.com'):
-        logging.warning('GitHub and Docker Hub are not reachable')
+    if not is_reachable('1.1.1.1'):
+        logging.warning('Unable to retrieve updates')
         return True  # We don't want to show the Update Available menu if Internet is not available.
 
     latest_sha, retrieved_update = fetch_remote_hash()
--- a/lib/github.py
+++ b/lib/github.py
@@ -3,6 +3,7 @@ from builtins import str
 from builtins import range
 import os
 import logging
+import socket
 import string
 import random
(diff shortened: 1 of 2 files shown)


lib/github.py
```diff
@@ -155,6 +169,10 @@ def is_up_to_date():
     Returns True if the player is up to date.
     """
 
+    if not is_reachable('1.1.1.1'):
```
vpetersson, line 172: I don't know about this. It just feels like an overly complicated workaround (which is also less reliable). You can't assume the device can't talk to GitHub or docker hub by checking if it can talk to cloudflare dns server. 

Just use a timeout argument on the request call.

vpetersson (changes requested):

lib/github.py
nicomiguelino: Even with adding a timeout argument in a request.get call, there's still some delay/latency. I'll probably try to disable retries.

vpetersson: Try a `request.head()` as a check with a 1s timeout. A HEAD request shouldn't take more than a second or so.

nicomiguelino: Tried that as well, but page loads are still slow.

vpetersson (approved):

PR #1971, opened 2024-07-11
Fixes host agent on Bullseye

nicomiguelino:
#### Description

* Changes to `anthias-host-agent.service` in #1950 introduced this issue to devices running Bullseye.
* This pull request added conditional statements inside the service file template.

### DIFF
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -87,8 +87,6 @@
     path: /etc/udev/rules.d/50-autoplay.rules
 
 - name: Copy screenly systemd units
-  vars:
-    debian_version: "{{ ansible_distribution_major_version }}"
   ansible.builtin.template:
     src: "{{ item }}"
     dest: "/etc/systemd/system/{{ item }}"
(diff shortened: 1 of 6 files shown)


vpetersson:
I'm not sure that I follow. Why is the environment used in one but not the other? Isn't it better to do some sort of migration to make them consistent?

nicomiguelino:
@vpetersson, Debian versions before Bookworm (version 12) don't require the use of a virtual environment when installing Python packages via `pip`. With Debian 12, virtual environment is required, unless `--break-system-packages` is specified.

I can migrate legacy versions to use a virtual environment.

vpetersson:
> @vpetersson, Debian versions before Bookworm (version 12) don't require the use of a virtual environment when installing Python packages via `pip`. With Debian 12, virtual environment is required, unless `--break-system-packages` is specified.
> 
> I can migrate legacy versions to use a virtual environment.

I think moving everything to virtual env/poetry will make things easier to manage.

nicomiguelino:
@vpetersson, I agree with this.

vpetersson (approved):

PR #1975, opened 2024-07-15
Fixes broken API docs page

nicomiguelino:
#### Description

* When #1950 was merged, the `/api/docs` URL broke.
*The URL specified in the code (for the API docs) is dependent `get_node_ip`, which returns multiple IPv4 and IPv6 addresses (as a string, separated by spaces).
* This pull request removes that dependency, making the `API_URL` just `/api/docs` for flexibility.

##### Before fixes

image

##### After fixes

image

image

### DIFF
--- a/server.py
+++ b/server.py
@@ -1318,14 +1318,7 @@ except Exception:
     pass
 else:
     SWAGGER_URL = '/api/docs'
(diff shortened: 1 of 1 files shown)


vpetersson (approved):
Much cleaner. Nice fix.

PR #1980, opened 2024-07-17
Upgrade containers to use Bookworm

### LINKED ISSUE #1753: Upgrade OS of containers from Buster to Bullseye or Bookworm
#### Overview

- We need to upgrade the OS that the services are using from Debian Buster (10.x) to Bullseye (11.x) or Bookworm (12.x).
- Upgrading the OS means that newer Python version will be installed for the containers (i.e., 3.9, 3.11).

#### Prerequisites

- #1091

nicomiguelino:
#### Description

* Resolves #1753
* Depends on #1982

#### Checklist

- [x] Re-build the webview but use Bookworm instead of Buster.
  - The webview can't start properly because it's searching for `*.so` files that are missing.
- [x] Update viewer Dockerfile
- Consider using Poetry &mdash; will be addressed in #1985 

#### Notes

* Moving away from OMXPlayer to VLC requires an upgrade from Buster.
* VLC playback in Buster is not possible, as per testing things quickly.
* One of the benefits of upgrading Python from

### DIFF
--- a/.github/workflows/docker-test.yaml
+++ b/.github/workflows/docker-test.yaml
@@ -64,7 +64,7 @@ jobs:
         shell: 'script -q -e -c "bash {0}"'
         run: |
           docker compose -f docker-compose.test.yml exec anthias-test bash ./bin/prepare_test_environment.sh -s
-          docker compose -f docker-compose.test.yml exec anthias-test nosetests -v -a '!fixme'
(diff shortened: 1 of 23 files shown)


vpetersson (approved):
Should we maybe switch to poetry there too instead?

nicomiguelino:
@vpetersson, do you mean using Poetry inside the containers?

vpetersson:
Yes.

PR #1998, opened 2024-07-28
Updates WebView Git hash and tag to point to the latest version.

nicomiguelino:
#### Description

* Updates the WebView Git hash and tag
* Removed `build_containers.sh` from the files to be ignored in the build workflow.


vpetersson (approved):
Approve assuming you've tested it properly manually.

PR #2007, opened 2024-08-02
Tweaks buildx settings to cache images properly

nicomiguelino:
#### Description

* This could reduce build times.


nicomiguelino:
This is already implemented in `master` and is implemented in Python:

https://github.com/Screenly/Anthias/blob/ec67066873290eb1b62ce5c36003ac5f47ba5532/tools/image_builder/__main__.py#L135-L138

However, the current change doesn't reduce the build time.

PR #2023, opened 2024-08-14
Creates a script for installing (trusted) self-signed certificates

### LINKED ISSUE #1824: Create tool for installing (trust) self-signed certificates
Replaces #1516

We should have a helper script for installing certificates inside the containers. It would be fairly straight forward and work something like this:

`./bin/add_certificate.sh /path/to/my.crt`

This would in turn install do something like this (run both against the viewer and server container):

* `docker exec -ti container-name mkdir /usr/local/share/ca-certificates/custom`
* `docker cp /path/to/my.crt container-name:/usr/local/share/ca-certificates/custom/`
* `docker exec -ti container-name update-ca-certificates`

Note that the above is completely untested, but should in theory work.

nicomiguelino:
#### Description

* Fixes #1824

### DIFF
--- a/bin/add_certificate.sh
+++ b/bin/add_certificate.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -euo pipefail
+
+function print_usage() {
+  echo "Usage: $0 /path/to/certificate.crt"
+}
+
+if [ "$#" -ne 1 ]; then
+  print_usage
+  exit 1
+fi
+
+CERTIFICATE_PATH=$1
+
+function main() {
+    if [ ! -f $CERTIFICATE_PATH ]; then
+        echo "Certificate file not found: $CERTIFICATE_PATH"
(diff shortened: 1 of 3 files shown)


bin/add_certificate.sh
```diff
@@ -0,0 +1,40 @@
...
+CERTIFICATE_PATH=$1
+
+function main() {
+    if [ ! -f $CERTIFICATE_PATH ]; then
+        echo "Certificate file not found: $CERTIFICATE_PATH"
+        exit 1
+    fi
+
+    CERTIFICATE_DIR='/usr/local/share/ca-certificates/custom'
+    CONTAINERS=(antthias-server anthias-viewer)
```
vpetersson, line 23: Typo: `antthias-server`

vpetersson (approved):
Approved assuming the typo change is fixed.

PR #2026, opened 2024-08-15
Modify install script to support install of a specific version

nicomiguelino:
#### Overview

install-anthias


bin/upgrade_containers.sh
```diff
@@ -11,10 +11,13 @@ export SHM_SIZE_KB="$(echo "$TOTAL_MEMORY_KB" \* 0.3 | bc | cut -d'.' -f1)"
 
 export GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
-if [ "$GIT_BRANCH" = "experimental" ]; then
-    export DOCKER_TAG="experimental"
-else
-    export DOCKER_TAG="latest"
+# Check if DOCKER_TAG is set
```
nicomiguelino, line 14: Remove comment.

vpetersson (commented):
Let's fix the wording here. We should say that latest and experimental are rolling releases. However should you want a specific release, you can used a pinned version. Or something like that.

vpetersson (approved):

PR #2028, opened 2024-08-16
Overhaul and refactor the installer script

nicomiguelino:
#### Description

* Downgrades `cryptography` from `38.0.2` (now yanked) to `38.0.1`.
* Refactored and cleaned up the installer script
* Used [Charm's gum](https://github.com/charmbracelet/gum) for:
  * Adding cosmetic changes to the script
  * Making the code use less boilerplate

#### Testing

```bash
bash <(curl -sL https://raw.githubusercontent.com/nicomiguelino/Anthias/modify-installer-script-2/bin/install.sh)
```

#### Installation Process in a Nutshell

Most of the steps were removed for simplicity.

install-anthias

As tested on a Raspberry Pi 4.

##### Headers Only

image

##### Installation

image

image

image

image

image

image


bin/install.sh
```diff
@@ -1,286 +1,301 @@
-#!/bin/bash -e
+#!/bin/bash
```
vpetersson, line 1: We want to use strict mode here.

bin/install.sh
```diff
@@ -1,286 +1,301 @@
...
-
-EOF
-
-  # Reset color
-  tput sgr 0
-
-  echo -e "Anthias requires a dedicated Raspberry Pi / SD card.\nYou will not be able to use the regular desktop environment once installed.\n"
-  read -p "Do you still want to continue? (y/N)" -n 1 -r -s INSTALL
-  if [ "$INSTALL" != 'y' ]; then
+set -euo pipefail
```
nicomiguelino, line 6: @vpetersson, I already set strict mode in this line instead.

bin/install.sh
nicomiguelino: Please see my comment below.

bin/install.sh
vpetersson: Fair.

PR #2031, opened 2024-08-19
Migrate the database to Django

### LINKED ISSUE #1913: Migrate database to use Django ORM.
#### Overview

- Depends on #558.
- Database manipulation is currently handled using the `sqlite3` Python library.
- The database should be migrated to use the Django ORM.

nicomiguelino:
#### Description

* Fixes #1913


api/serializers.py
```diff
@@ -20,7 +31,257 @@ class AssetRequestSerializer(Serializer):
...
+                        break
+
+        asset = {
+            'name': name,
+            'mimetype': data.get('mimetype'),
+            'is_enabled': data.get('is_enabled'),
+            'nocache': data.get('nocache'),
+        }
+
+        uri = (
```
vpetersson, line 188: Is there a better way to do this through some library instead? Just feels like fragile code.

vpetersson (approved):
LGTM. Didn't review super closely, but tests should cover most and then solid QA is needed.  One thing is is that we rely a lot on `sleep()` in tests in particular. Would be better to rely on some kind of timeout instead.

PR #2036, opened 2024-08-27
Install `python-is-python3` so that `python` will point to `python3`

nicomiguelino:
#### Description

* Replaced references to `python3` to `python`.


docker/Dockerfile.viewer.tmpl
```diff
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.4
+#xsyntax=docker/dockerfile:1.4
```
vpetersson, line 1: Typo

vpetersson (approved):

PR #2045, opened 2024-08-30
Replace `nose` attributes with `unittest`'s `skip`

nicomiguelino:
#### Description

* Using the `@skip` decorator will make running tests more simpler.
* With this, we can just run `nose2 -v` instead of `nose2 -v -A '!fixme'`

#### Other Changes

* The `docker-build.yaml` workflow was modified so that the CI won't build the Docker images if tests were added or modified.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -17,6 +17,7 @@ on:
       - '!docs/**'
       - '!bin/install.sh'
       - '!bin/upgrade_containers.sh'
+      - '!tests/**'
 
 jobs:
   run-tests:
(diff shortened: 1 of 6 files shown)


vpetersson (approved):
Approved but let's actually fix the tests rather than skipping them.

nicomiguelino:
@vpetersson, I'll be pushing fixes on separate PR's, as some skipped unit tests require a fair bit of work, while some are already irrelevant (like those in `updates_test.py`, as the `latest_anthias_sha`, formerly `latest_screenly_sha`, [was already decommissioned years ago](https://github.com/Screenly/Anthias/commit/2d81dc404ea5204d9f5e8c1109c5eb8ca975ad0f#diff-0bb30e60c9060eae4f3b622c188c4ed97bc18c04242404cd8a9a4098961082e4)). For the updates test, I'll be writing new tests to validate the current behavior.

PR #2048, opened 2024-09-02
Add support for x86 devices

nicomiguelino:
#### Description

* This pull request adds support for x86 devices.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -123,15 +123,16 @@ function initialize_locales() {
 function install_packages() {
     display_section "Install Packages via APT"
 
-    RASPBERRY_PI_OS_VERSION=$(lsb_release -rs)
-    APT_INSTALL_ARGS=(
+    local DISTRO_VERSION=$(lsb_release -rs)
+    local APT_INSTALL_ARGS=(
         "git"
         "libffi-dev"
         "libssl-dev"
         "whois"
     )
+    local ARCHITECTURE=$(uname -m)
 
-    if [ "$RASPBERRY_PI_OS_VERSION" -ge 12 ]; then
+    if [ "$DISTRO_VERSION" -ge 12 ]; then
         APT_INSTALL_ARGS+=("python3-full")
     else
         APT_INSTALL_ARGS+=(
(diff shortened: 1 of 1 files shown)


ansible/roles/system/tasks/main.yml
```diff
@@ -206,15 +209,28 @@
     update_cache: true
   when: not cdefs_exist
 
-- name: Install Anthias dependencies
+- name: Get architecture
+  ansible.builtin.command: dpkg --print-architecture
```
vpetersson, line 213: Use a more native way for this. You can get it from ansible directly. 

The long term plan is to be less Debian specific so the less hard coded we can be the better.

bin/install.sh
```diff
@@ -146,8 +147,11 @@ function install_packages() {
         APT_INSTALL_ARGS+=("network-manager")
     fi
 
-    sudo sed -i 's/apt.screenlyapp.com/archive.raspbian.org/g' \
-        /etc/apt/sources.list
+    if [ "$ARCHITECTURE" != "x86_64" ]; then
```
vpetersson, line 150: What about x86_32? Better to just look for x86 in the string.

vpetersson (approved):
Needs to comment back a few things and swap out your branch but good progress. Also keep in mind that x86 can run on 32 bit too.

bin/upgrade_containers.sh
```diff
@@ -49,7 +51,7 @@ cat /home/${USER}/screenly/docker-compose.yml.tmpl \
 
 sudo -E docker compose \
     -f /home/${USER}/screenly/docker-compose.yml \
-    pull
+    build
```
nicomiguelino, line 54: Revert this change.

bin/install.sh
nicomiguelino: Focus will be put on 64-bit x86 devices as most of them already uses 64-bit.

ansible/roles/system/tasks/main.yml
vpetersson: @nicomiguelino this is what I mean

https://stackoverflow.com/questions/44713880/how-do-i-make-decision-based-on-arch-in-ansible-playbooks

Meaning, use "ansible_facts.ansible_architecture" instead of shelling out.

vpetersson (commented):
This PR should probably still be in draft given that it had a lot of commented out code with notes. That said, looks promising

ansible/roles/system/tasks/main.yml
nicomiguelino: Got it, thanks! Please see newer commits.

vpetersson (approved):
LGTM

PR #2056, opened 2024-09-06
Create a pipeline for building an x86-compatible WebView

nicomiguelino:
#### Description

* The WebView binary that will be built will use Qt 6 instead of Qt 5.

#### Checklist

- [x] Add docs on how to build an x86-compatible WebView.
- [x] Resolve check-level and other related issues.


vpetersson (approved):
Nicely done

webview/Dockerfile.x86
```diff
@@ -0,0 +1,98 @@
...
+        rsync \
+        sed \
+        texinfo \
+        wget \
+        zlib1g-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*; \
+} 2>&1 | tee -a /build.log
+
+RUN pip install --upgrade pip --break-system-packages
```
vpetersson, line 90: Should we not start moving to poetry?

vpetersson (approved):

PR #2057, opened 2024-09-06
Enable X11 when building Qt 5

nicomiguelino:
#### Description

* Setting `QT_QPA_PLATFORM` in the client (Python) code won't work if X11 is enabled during compilation.


nicomiguelino:
Closing this since this pull request has become stale.

PR #2060, opened 2024-09-09
chore(workflow): port the Docker image builder script to Python

nicomiguelino:
#### Description

* We'll be leveraging Poetry for Python dependency management.
* Customizing Dockerfiles across platforms can be hack-ish when done via Bash scripts. Some parts of the script doesn't look intuitive.
* Porting the workflow to using Python (more specifically, Jinja for templating the Dockerfiles) enables users and contributors to make changes in a more elegant way.

#### Use cases where we can benefit from these changes

* In a development environment, we'll be installing tools like Node.js and CSS preprocessors that shouldn't be included in production Dockerfiles.
* Qt 5/6 installation can vary across all the supported platforms. If we'll be gradually migrating from Qt 5 to 6, this can b e beneficial. We don't need to write `sed` statements in the build script in order to filter out unnecessary deps.


vpetersson (approved):
I think this is actaully more complicated than the `.tmpl` structure, but fine.

PR #2069, opened 2024-09-12
Install `python3-dev` during installation to prevent errors

nicomiguelino:
#### Description

* Installing `python3-dev` prevents the following error when trying to install Python dependencies on a Raspberry Pi 2 running Bookworm.


nicomiguelino:
@vpetersson, sounds good.

PR #2078, opened 2024-10-01
Remove docs for playing around with basic authentication via the command line

nicomiguelino:
* Changing authentication settings can be done via the UI, so those docs could be removed.


vpetersson (commented):
How would you even set this on Balena?

nicomiguelino:
@vpetersson, the documentation is for user that uses Anthias that runs on top of Raspberry Pi OS Lite.
Let me add an info for that.

vpetersson:
> @vpetersson, the documentation is for user that uses Anthias that runs on top of Raspberry Pi OS Lite. Let me add an info for that.

Right - but we probably need to overhaul the general thinking for all features to make sure they can be done in the UI.

nicomiguelino:
@vpetersson, fair point. Thanks. I'll probably remove the documentation as setting up basic authentication in Anthias can be done in the UI, both for RPi OS and BalenaOS instances.

vpetersson (approved):

PR #2080, opened 2024-10-03
Use the complete commit hash in WebView filenames for consistency

nicomiguelino:
* The existing WebView releases have long hashes in their file names, so I updated the docs and the CI workflow file for consistency.


vpetersson:
Short should surely be enough, no?

nicomiguelino:
@vpetersson, fair point. We can just use short hashes when creating new WebView releases in the future.

PR #2081, opened 2024-10-04
Create a pipeline for building WebView via Qt 6

nicomiguelino:
* The compilation works successfully, but it doesn't work in the viewer.


nicomiguelino:
Raspberry Pi 5 and x86 devices already uses Qt 6. Efforts for using Qt 6 for the rest of the supported devices (if possible) will be done in separate pull requests.

PR #2087, opened 2024-10-10
Make the database migration process atomic

nicomiguelino:
#### Description

* Make the database migration process atomic, which means that Anthias will roll back to the previous DB state if the migration fails.
* This [Django doc](https://docs.djangoproject.com/en/3.2/topics/db/transactions/#topics-db-transactions-savepoints) states that savepoints are available if the database backend is SQLite, which Anthias is currently using at the moment.
   * The changes in this PR are for good measures &mdash; for edge cases like when the device powered off mid-migration.


bin/start_server.sh
```diff
@@ -24,8 +24,12 @@ fi
 echo "Running migration..."
 
 ./manage.py initialize_assets
```
renatgalimov, line 26: What about initializing assets? Is it atomic?

renatgalimov: I'd like to see the whole process atomic. 

It's like

```
mkdir data.tmp 
.... do all the stuff
mv data.tmp data
```

So, the whole process is atomic, including other steps, like asset initialization. 

The whole file looks strange. It's called `start_sever.sh`, so I expect it will run all the time, but wouldn't it migrate every time it starts?

nicomiguelino: @renatgalimov,
* `initialize_assets` just makes sure if the required directories are present.
  * Maybe it's safe to remove it, I'll check. See screenshot below (from `start_server.sh`) for details.

image

* Yes, even a long time ago, the migration runs every time the server starts. Is that an acceptable behavior?

bin/start_server.sh
```diff
@@ -12,9 +12,11 @@ cp -n /usr/src/app/ansible/roles/screenly/files/default_assets.yml /data/.screen
...
 echo "Running migration..."
 
-./manage.py initialize_assets
-./manage.py makemigrations
-./manage.py migrate --fake-initial
+# The following block ensures that the migration is transactional and that the
+# database is not left in an inconsistent state if the migration fails.
+cp /data/.screenly/screenly.db /data/.screenly/backup.db
+./manage.py migrate --fake-initial --database=backup
+cp /data/.screenly/backup.db /data/.screenly/screenly.db
```
renatgalimov, line 19: Hmm... this way, we lose the old database. 

```suggestion
cp /data/.screenly/screenly.db /data/.screenly/backup.db
cp /data/.screenly/screenly.db /data/.screenly/screenly.db.bak
./manage.py migrate --fake-initial --database=backup
mv /data/.screenly/backup.db /data/.screenly/screenly.db
```

bin/start_server.sh
```diff
@@ -12,9 +12,11 @@ cp -n /usr/src/app/ansible/roles/screenly/files/default_assets.yml /data/.screen
 
 echo "Running migration..."
 
-./manage.py initialize_assets
-./manage.py makemigrations
-./manage.py migrate --fake-initial
+# The following block ensures that the migration is transactional and that the
+# database is not left in an inconsistent state if the migration fails.
+cp /data/.screenly/screenly.db /data/.screenly/backup.db
```
renatgalimov, line 17: Above, you do `mkdir -p ... /data/.screenly ...`, which will make an empty directory.
But here, you expect the existence of a screenly db.

Could you rewrite the script so it will make sense?

nicomiguelino: @renatgalimov, I accidentally deleted the line that copies `screenly.db` from the `./ansible` folder.
I'll bring it back. I hope it makes more sense later.

renatgalimov: Now, it makes even less sense to me🤔 Is the DB expected to be modified? Would I lose my modifications every time I restart this script?

bin/start_server.sh
```diff
@@ -12,9 +12,17 @@ cp -n /usr/src/app/ansible/roles/screenly/files/default_assets.yml /data/.screen
...
+# The following block ensures that the migration is transactional and that the
+# database is not left in an inconsistent state if the migration fails.
+
+if [ -f /data/.screenly/screenly.db ]; then
+    cp /data/.screenly/screenly.db /data/.screenly/backup.db
+    cp /data/.screenly/screenly.db /data/.screenly/screenly.db.bak
+    ./manage.py migrate --fake-initial --database=backup
+    mv /data/.screenly/backup.db /data/.screenly/screenly.db
+else
+    ./manage.py migrate
```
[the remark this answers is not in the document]
nicomiguelino: Yes, just to make things more explicit/intuitive/straightforward here. Thanks.

PR #2089, opened 2024-10-11
fix: set `architecture` to `dpkg --print-architecture`

nicomiguelino:


vpetersson:
Why are we reverting this?

nicomiguelino:
@vpetersson, so that arm64/aarch64 Docker will be installed for 64-bit OS's.
* I'm only reverting part of the chamges made before.
* In theory, it shouldn't break anything, but I still have to test this thoroughly for all supported platforms.

PR #2040, opened 2024-08-28
Merge `experimental` features to `master`

nicomiguelino:
#### Description

This pull request contains the following Django-related PRs merged to `experimental`: 

* Screenly/Anthias#1821
* Screenly/Anthias#2031
* Screenly/Anthias#2087

PR #2095, opened 2024-10-15
Update asset database schema

nicomiguelino:
#### Description

* As per the review comments in #2040, the **_Asset_** model needs to be updated, schema-wise.


vpetersson (approved):
LGTM

PR #2099, opened 2024-10-18
`is_active` should be boolean in API docs

nicomiguelino:
#### Before

image

#### After

image

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -33,6 +33,8 @@ class AssetRequestSerializer(Serializer):
 
 
 class AssetSerializer(ModelSerializer):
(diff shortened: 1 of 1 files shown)


vpetersson:
Technically speaking you can't change an API without bumping the version. You'd break every integration this way.

nicomiguelino:
@vpetersson, good point. Thanks. Because we've changed the `0` or `1` fields to boolean fields, I'll update v1.2 endpoints so that the boolean fields will be `0` or `1` again. I'll create a new API version, `v2.0`, as those changes will potentially break clients currently using the API.

PR #2103, opened 2024-10-30
Revert field-related changes in v1.2 endpoints

nicomiguelino:
#### Description

* In #2040, the endpoints were modified so that
  * `0` or `1` integer fields will be boolean fields AND
  * the duration field will be integer fields (from char fields)
* Those are breaking changes, as there are already clients using those endpoints.
* A separate pull request will be created where a new set of endpoints will be created, labeled as `v2.0`.
* Changes to the Asset model will not be reverted.

### DIFF
--- a/docs/developer-documentation.md
+++ b/docs/developer-documentation.md
@@ -171,7 +171,7 @@ After installing Poetry, run the following commands:
 ```bash
 # Install the dependencies
 $ poetry install --only=dev-host
-$ poetry run flake8 $(git ls-files '**/*.py')
+$ poetry run flake8 $(git ls-files '*.py')
 ```
 
 To run the linter on a specific file, run the following command:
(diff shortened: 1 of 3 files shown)


docs/developer-documentation.md
```diff
@@ -171,7 +171,7 @@ After installing Poetry, run the following commands:
 ```bash
 # Install the dependencies
 $ poetry install --only=dev-host
-$ poetry run flake8 $(git ls-files '**/*.py')
+$ poetry run flake8 $(git ls-files '*.py')
```
vpetersson, line 174: Don't  change this -- those two things are different.

vpetersson (approved):
Approved pending minor change.

docs/developer-documentation.md
nicomiguelino: Yes, those two are different. In that case, `**/*.py` has less coverage than `*.py`. I tested it before and just now.

image

`*.py` is what's being used in GitHub actions at the moment:

https://github.com/Screenly/Anthias/blob/70afe9ba49d42f8a050cce2faaf2a963cd4a8bb7/.github/workflows/python-lint.yaml#L47

PR #2104, opened 2024-10-31
chore(tools): write a script for starting the development server

nicomiguelino:
#### Description

* Running a few commands can be quite cumbersome, especially those that requires non-default options.

### DIFF
--- a/bin/start_development_server.sh
+++ b/bin/start_development_server.sh
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+COMPOSE_ARGS=(
+    '-f' 'docker-compose.dev.yml'
+)
 
 function generate_dockerfiles() {
     ENVIRONMENT=development \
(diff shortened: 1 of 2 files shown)


bin/start_development_server.sh
```diff
@@ -0,0 +1,23 @@
...
+    local COMPOSE_ARGS=(
+        '-f' 'docker-compose.dev.yml'
+    )
+
+    generate_dockerfiles
+    docker compose "${COMPOSE_ARGS[@]}" down
+    docker compose "${COMPOSE_ARGS[@]}" up -d --build
+}
+
+main
```
vpetersson, line 23: This is the most pythonic bash script I've ever seen 😂

docs/developer-documentation.md
```diff
@@ -25,12 +25,10 @@ Do note that Anthias is using Docker's [buildx](https://docs.docker.com/engine/r
...
 ```bash
-$ ENVIRONMENT=development \
-  DOCKERFILES_ONLY=1 \
-  DISABLE_CACHE_MOUNTS=1 \
-  ./bin/build_containers.sh
-$ docker compose \
-    -f docker-compose.dev.yml up -d --build
+$ ./bin/start_development_server.sh
+
+# This should start the development server and you should be able to
```
vpetersson, line 30: Shouldn't this be an echo instead?

vpetersson (approved):
Approved with minor comments

PR #2111, opened 2024-11-04
fix: add support for Raspberry Pi Zero 2 W

### LINKED ISSUE #2108: Raspberry Pi Zero 2 W Support?
#### Overview of the Issue

Are there any plans to officially support the Raspberry Pi Zero 2 W?

#### Reproduction Steps / Why I Ask

I ask about official support because I've been experimenting with running Anthias on the Pi Zero 2 W and I keep running into horrific disk io contention issues. An overview of my progress is below.

I was able to sort the installer problems I found and was able to get the install to run to completion within 45-60 minutes on average. However, during container start up the disk io contention makes the Pi Zero 2 W unusable. I noticed something similar on my pi4 deployment: whenever the containers start, they use up all the disk io, cause the load average to spike over 15-20 and the system becomes unusable. On the pi4 the heavy disk io eventually settles and the load average drops within a 10 minutes usually but these two things can be very problematic during startup on the pi4.

On the Pi Zero 2... I've yet to find a way to constrain the intense disk io the containers cause during startup to allow Anthias to actually run. I'm seeing load averages as high as 60 when starting the containers without tuning and with some significant tuning I'm seeing load averages over 30. In either scenario the system becomes unusable very quickly. Additionally, the system will eventually throw a kernel error saying a task is hung. Once this happens, the Pi Zero 2 W is effectively crashed and unusable. The server and viewer containers seem to cause the biggest problems for disk io contention in my experiments.

From what I can figure out, the containers would need to be started sequentially and not in parallel and/or tuned to be less aggressive on disk io during startup. Without the disk io contention addressed, I don't see how the Pi Zero 2 W could be supported.

I can provide my notes if desired but they are dense and very long with a pair of patches that make some significant changes to the setup process and docker-compose template. Please let me know if you'd like this additional detail.

#### Environment

* Raspberry Pi Hardware Version: Raspberry Pi Zero 2 W
* Raspberry Pi Network Setup: Built in WiFi
* Anthias Version: GitHub sources, revision 8acb12d683250c7a9b9fa21590d62ae4e98056fb

nicomiguelino:
#### Description

* Resolves #2108 
  * Thanks to @mcrosson for the suggested fixes/workarounds.
  * See this [suggested patch](https://github.com/Screenly/Anthias/issues/2108#issuecomment-2454801984) for details.


juanyacovino77:
is Anthias working for Raspberry pi zero 2w?

nicomiguelino:
@juanyacovino77, Anthias doesn't support Raspberry Pi Zero devices (out of the box) at the moment.

vpetersson:
Closing - we're dropping support for pi 1 / zero.

PR #2115, opened 2024-11-06
chore: install Node.js in Docker containers (for development mode)

nicomiguelino:
#### Description

* One-up developer experience by installing Node.js in Docker containers, so that they won't have to install Node.js on the host OS.'

> [!NOTE]
> #2060 provides a more elegant solution for conditionally generating Dockerfiles.
> If it gets merged, this PR is still up for more changes.

#### Screenshot(s)

image

### DIFF
--- a/bin/build_containers.sh
+++ b/bin/build_containers.sh
@@ -113,25 +113,23 @@ for container in ${SERVICES[@]}; do
         cat "docker/Dockerfile.$container.tmpl" | envsubst > "docker/Dockerfile.$container"
     fi
 
+    PACKAGES_TO_REMOVE=()
+
+    if [[ $OSTYPE == 'darwin'* ]]; then
+        SED_ARGS=(-i "")
+    else
+        SED_ARGS=(-i)
+    fi
+
     # If we're running on x86, remove all Pi specific packages
     if [ "$BOARD" == 'x86' ]; then
-        if [[ $OSTYPE == 'darwin'* ]]; then
-            SED_ARGS=(-i "")
-        else
-            SED_ARGS=(-i)
-        fi
-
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.base.tmpl
```diff
@@ -23,6 +23,8 @@ RUN --mount=type=cache,target=/var/cache/apt \
         lsb-release \
         mplayer \
         net-tools \
+        nodejs \
```
vpetersson, line 26: This should be a multi stage build. No reason to keep this around after build.

vpetersson (changes requested):

docker/Dockerfile.base.tmpl
nicomiguelino: @vpetersson, in the future, #2060 can help us make changes (even the more sophisticated ones) &mdash; like putting multi-stage builds in templates &mdash; more easily.

bin/build_containers.sh
```diff
@@ -144,6 +142,17 @@ for container in ${SERVICES[@]}; do
...
+        PACKAGES_TO_REMOVE+=(
+            "nodejs"
+            "npm"
+        )
+    fi
+
+    for package in "${PACKAGES_TO_REMOVE[@]}"; do
+        sed "${SED_ARGS[@]}" -e "/$package/d" $(find docker/ -maxdepth 1 -not -name "*.tmpl" -type f)
+    done
+
```
nicomiguelino, line 155: @vpetersson, here's my code for removing development dependencies.

nicomiguelino:
@vpetersson, @salmanfarisvp
I'll be closing this pull request since I'll move the changes to #2117.

PR #2117, opened 2024-11-07
chore: use multi-stage builds for server images in both development and production environments

nicomiguelino:


docker/Dockerfile.server.tmpl
```diff
@@ -5,6 +5,12 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     pip3 install -r /tmp/requirements.txt --break-system-packages
 
 RUN mkdir -p /usr/src/app
+COPY --from=node-builder \
+    /app/static/css/ \
+    /usr/src/app/static/css/
+COPY --from=node-builder \
+    /app/static/js/*.js* \
+    /usr/src/app/static/js/
```
nicomiguelino, line 13: @vpetersson, probably one of the motivations for creating #2060 is that a server Dockerfile can have different setups for development and production environments.

docker/Dockerfile.server.j2
```diff
@@ -1,3 +1,24 @@
+{% if environment == 'production' %}
+FROM debian:bookworm as node-builder
+
+RUN --mount=type=cache,target=/var/cache/apt \
```
nicomiguelino, line 4: This line should be wrapped in a conditional statement.

vpetersson (approved):

PR #2118, opened 2024-11-07
fix: resolve issue when building the viewer Docker image for x86

nicomiguelino:
#### Description

* Removes `libraspberrypi0` from `Dockerfile.viewer` for x86
* A follow-up pull request to #2060

#### Failed CI Runs

* https://github.com/Screenly/Anthias/actions/runs/11724576804/job/32659200061#step:10:1605

### DIFF
--- a/tools/image_builder/__main__.py
+++ b/tools/image_builder/__main__.py
@@ -123,7 +123,7 @@ def build_image(
         'sqlite3',
     ]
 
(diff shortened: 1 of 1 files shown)


vpetersson (approved):
Probably better to call this 'raspberry_pi' rather than not x86, but fine.

PR #2124, opened 2024-11-12
Add support for asoundrc

mcrosson:
Some x86_64 mini pc's need an `~/.asoundrc` file within the `viewer` container for sound output to work. This PR enables support for setups that need an `~/.asoundrd`.

This has been tested with a [MeLE PCG02 Mini PC 'Stick'](https://mele.cn/product/127-en.html). The `~/.asoundrc` this device needs is:

```
pcm.!default {
    type hw
    card 0
    device 1
}

ctl.!default {
    type hw
    card 0
    device 1
}
```

### DIFF
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -12,13 +12,13 @@
     - screenly_assets
 
 - name: Create empy .asoundrc
-  file:
+  ansible.builtin.file:
     path: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
     state: touch
     force: false
     group: "{{ lookup('env', 'USER') }}"
     owner: "{{ lookup('env', 'USER') }}"
-    mode: 0644
+    mode: "0644"
 
 - name: Copy Screenly default assets file
   ansible.builtin.copy:
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -12,8 +12,9 @@
     - screenly_assets
 
 - name: Create empy .asoundrc
-    content: ""
-    dest: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
+  file:
+    path: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
+    state: touch
     force: false
     group: "{{ lookup('env', 'USER') }}"
     owner: "{{ lookup('env', 'USER') }}"
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -11,6 +11,14 @@
     - .config
     - screenly_assets
 
+- name: Create empy .asoundrc
+    content: ""
+    dest: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
+    force: false
+    group: "{{ lookup('env', 'USER') }}"
+    owner: "{{ lookup('env', 'USER') }}"
+    mode: 0644
+
 - name: Copy Screenly default assets file
   ansible.builtin.copy:
     owner: "{{ lookup('env', 'USER') }}"
--- a/ansible/roles/screenly/tasks/main.yml
+++ b/ansible/roles/screenly/tasks/main.yml
@@ -11,7 +11,7 @@
     - .config
     - screenly_assets
 
-- name: Create empy .asoundrc
+- name: Create empty .asoundrc
   ansible.builtin.file:
     path: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
     state: touch
--- a/docker-compose.yml.tmpl
+++ b/docker-compose.yml.tmpl
@@ -37,7 +37,6 @@ services:
     restart: always
     volumes:
       - resin-data:/data
-      - /home/${USER}/.asoundrc:/data/.asoundrc
       - /home/${USER}/.screenly:/data/.screenly
       - /home/${USER}/screenly_assets:/data/screenly_assets
       - /home/${USER}/screenly/staticfiles:/data/screenly/staticfiles
@@ -66,6 +65,7 @@ services:
     shm_size: ${SHM_SIZE_KB}kb
     volumes:
       - resin-data:/data
+      - /home/${USER}/.asoundrc:/data/.asoundrc
       - /home/${USER}/.screenly:/data/.screenly
       - /home/${USER}/screenly_assets:/data/screenly_assets
       - /etc/timezone:/etc/timezone:ro
--- a/docker-compose.yml.tmpl
+++ b/docker-compose.yml.tmpl
@@ -37,6 +37,7 @@ services:
     restart: always
     volumes:
       - resin-data:/data
+      - /home/${USER}/.asoundrc:/data/.asoundrc
       - /home/${USER}/.screenly:/data/.screenly
       - /home/${USER}/screenly_assets:/data/screenly_assets
       - /home/${USER}/screenly/staticfiles:/data/screenly/staticfiles


nicomiguelino:
@mcrosson, thanks for creating a pull request!
I'll go test it on my x86 NUC-like device and on my Pi devices as well (to ensure nothing breaks).

docker-compose.yml.tmpl
```diff
@@ -37,6 +37,7 @@ services:
     restart: always
     volumes:
       - resin-data:/data
+      - /home/${USER}/.asoundrc:/data/.asoundrc
```
nicomiguelino, line 40: Looks like you've accidentally placed this in the `anthias-server` section. Please move this to `anthias-viewer`.

nicomiguelino (commented):
* Card and device numbers will vary across devices, so `.asoundrc` needs to be dynamic too.
* Have you checked if there's audio on your mini PC's 3.5mm jack? If there's no `.asoundrc` config (or the config is blank), if could be that it defaults to the audio jack.

mcrosson:
I was able to carve out some time to do a quick test of this device with no `.asoundrc` and an empty `.asoundrc`. Both forms resulted in the following error. I have a feeling this device just has an 'odd' form of sound setup that'll require tuning the audio setup via `.asoundrc` or equivalent.

````
SDL_OpenAudio (2 channels, 44100 Hz): ALSA: Couldn't open audio device: No such file or directory
SDL_OpenAudio (1 channels, 44100 Hz): ALSA: Couldn't open audio device: No such file or directory
```

docker-compose.yml.tmpl
nicomiguelino: That would be fine. We can just squash and merge this pull request later if this gets approved

nicomiguelino:
@mcrosson, I;m still thinking abouit options. We can either let users do one of the following:

* Manually change the `.asoundrc` file
  * This is fine for devices running Raspberry Pi OS or Debian.
  * If we'll add support for BalenaOS as well, this won't be an option.
* Let the user edit the `.asoundrc` file via the web UI (for instance, using `textarea`), on the settings page.
  * This is okay if we'd like users to have more control, but might be overwhelming to some users.
* Have the back-end fetch the card and devices info, but depending on the device, it's hard to know which device/cards are for HDMI, 3.5mm, etc.
  * If we are to go down this path, we might do a fair bit of Python programming.

mcrosson:
Understood re evaluating options.

For the web ui text field: could you set it up to have a url parm to enable editing of the file and/or have it setup behind an 'advanced sound options (you probably dont want this)' toggle or similar? Such an approach could be a good compromise for the time being.

I think the python approach may be 'best' long-term but I have a feeling that'll involve a ton of programming...

nicomiguelino (changes requested):
@mcrosson, would you be able to make changes so that:
* .asoundrc will be created from the viewer docker container instead.
  * Make sure to store it inside `/data`.
  * In the viewer container, `$HOME` is set to `/data`. That path is mounted as a volume, so it will be persisted.
 
If the viewer container will take care of creating an `.asoundrc` file (if it doesn't exist yet), changes to the Ansible files can be removed.

mcrosson:
@nicomiguelino

> * .asoundrc will be created from the viewer docker container instead?

The `.asoundrc` file cannot be instantiated by the container with the way the `/data` directory is setup currently. My understanding and experience with Docker is that a file must exist prior to the container launching otherwise Docker will auto-create a directory instead of a file at the source path. If there is a way to force Docker to create a file instead of a folder, I can make that update but I've never seen a way to do this.

The other option would be to bind mount the whole of `/data` but I don't think that's a desired behavior.

> * Make sure to store it inside `/data`.
>  * In the viewer container, `$HOME` is set to `/data`. That path is mounted as a volume, so it will be persisted.

This is already addressed by the `docker-compose.yml.tmpl` changes in the PR.

nicomiguelino:
@mcrosson

> The other option would be to bind mount the whole of /data but I don't think that's a desired behavior.

Although there's no bind mount for `/data`, there's a volume mount for that (line 67):

https://github.com/Screenly/Anthias/blob/47947f42108920cfc36b35607978ee9699732259/docker-compose.yml.tmpl#L48-L73

Here's the voluime declaration:

https://github.com/Screenly/Anthias/blob/47947f42108920cfc36b35607978ee9699732259/docker-compose.yml.tmpl#L149-L152

One caveat when creating an `.asoundrc` from the Dockerfile is that you have to copy `.asoundrc` from the container to host, edit it, then copy it back. (Unless there's a way to modify it from the settings page.)


I'll keep you posted.

ansible/roles/screenly/tasks/main.yml
```diff
@@ -11,6 +11,15 @@
     - .config
     - screenly_assets
 
+- name: Create empy .asoundrc
+  file:
```
nicomiguelino, line 15:
```suggestion
  ansible.builtin.file:
```

See this [documentation](https://ansible.readthedocs.io/projects/lint/rules/fqcn/#problematic-code) about FQCN for reference. By making things more specific/explicit, there's less chance of unexpected behaviors.

ansible/roles/screenly/tasks/main.yml
```diff
@@ -11,6 +11,15 @@
...
+    path: "/home/{{ lookup('env', 'USER') }}/.asoundrc"
+    state: touch
+    force: false
+    group: "{{ lookup('env', 'USER') }}"
+    owner: "{{ lookup('env', 'USER') }}"
+    mode: 0644
```
nicomiguelino, line 21:
```suggestion
    mode: "0644"
```

Enclose octal values in double quotes, as per https://ansible.readthedocs.io/projects/lint/rules/risky-octal/#correct-code.

nicomiguelino (changes requested):
@mcrosson, also make sure to test this with existing installation.

Here's a suggestion on how to test it:
* Do a fresh installation (`master`).
* Then  try to install your changes.
  * You might want to temporarily modify the install script and use that modified script instead.

Let me know if you need details on the how-tos.

mcrosson:
I've tested the ansible updates and they are working as expected and will not override an existing `.asoundrc` file and will create an empty `.asoundrc` file if it's not present on the system. The update have been pushed and should be good to go.

Also: given the web ui code for hdmi/3.5mm audio toggle should work with rpi devices: would it be good to apply the `.asoundrc` only when the system is not a raspberry pi?

I'd prefer leave the bind mount for `.asoundrc` for the moment, it's far less fiddly than trying to pull & push a file to a docker volume. Is this ok to leave as-is?

bin/install.sh
```diff
@@ -200,10 +197,6 @@ function run_ansible_playbook() {
 function upgrade_docker_containers() {
     display_section "Initialize/Upgrade Docker Containers"
 
-    wget -q \
```
nicomiguelino, line 203: Why was this removed?
* Have you encountered any issues if this wasn't removed?

bin/install.sh
```diff
@@ -184,9 +184,6 @@ function install_ansible() {
 function run_ansible_playbook() {
     display_section "Run the Anthias Ansible Playbook"
 
-    sudo -u ${USER} ${SUDO_ARGS[@]} ansible localhost \
-        -m git \
-        -a "repo=$REPOSITORY dest=${ANTHIAS_REPO_DIR} version=${BRANCH} force=yes"
```
nicomiguelino, line 189: Why was this removed?
(What are the issues that you've encountered before removing this?)

bin/install.sh
```diff
@@ -200,10 +197,6 @@ function run_ansible_playbook() {
 function upgrade_docker_containers() {
     display_section "Initialize/Upgrade Docker Containers"
 
-    wget -q \
-        "$GITHUB_RAW_URL/master/bin/upgrade_containers.sh" \
-        -O "$UPGRADE_SCRIPT_PATH"
```
nicomiguelino, line 205: Why was this removed?
(What are the issues that you've encountered before removing this?)

bin/install.sh
mcrosson: This was a mistake on my part. I've removed this change from my code and it should no longer be a problem.

bin/install.sh
mcrosson: This was a mistake on my part. I've removed this change from my code and it should no longer be a problem.

mcrosson:
The changes to `install.sh` were not supposed to be included in my pushed code. I've removed these changes from my repo and the `install.sh` file is no longer touched by this PR.

nicomiguelino (changes requested):
@mcrosson, here are some comments:

* Resolve the typo.
* Add the following in your `bin/upgrade_containers.sh`:

```bash
    sed -i '/\.asoundrc/d' \
        /home/${USER}/screenly/docker-compose.yml
```

Make sure to add it just below the block below:

```bash
if [ "$DEVICE_TYPE" = "x86" ]; then
    sed -i '/devices:/ {N; /\n.*\/dev\/vchiq:\/dev\/vchiq/d}' \
        /home/${USER}/screenly/docker-compose.yml
```

After that, your `upgrade_containers.sh` would look like the following:

```bash
if [ "$DEVICE_TYPE" = "x86" ]; then
    sed -i '/devices:/ {N; /\n.*\/dev\/vchiq:\/dev\/vchiq/d}' \
        /home/${USER}/screenly/docker-compose.yml
    sed -i '/\.asoundrc/d' \
        /home/${USER}/screenly/docker-compose.yml
fi
```

That change is needed so that your changes will only take effect on x86 devices.

mcrosson:
The typo has been fixed.

nicomiguelino:
@mcrosson, should you have some more changes to make, feel free to create another pull request since I just merged this one.

PR #2131, opened 2024-11-18
chore: deprecate the use of CoffeeScript

nicomiguelino:


nicomiguelino:
If we are to migrate to React for instance, this might not be necessary.

PR #2216, opened 2025-02-21
feat: write scripts for building a Pi 4 (64-bit) WebView

nicomiguelino:
### Description

* Writes a script that builds WebView (via a pre-built version of Qt 6) for Raspberry Pi 4 devices running 64-bit Raspberry Pi OS

### Checklist


webview/docker-compose.yml
nicomiguelino, line 1: Refactor common attributes.

webview/docker/Dockerfile.pi4-64
nicomiguelino, line 1: We can maybe create a script that generates custom Dockerfiles based on the given platform (e.g., pi1, pi2, pi3, pi4, pi4-64, pi5, etc.).

webview/src/mainwindow.cpp
```diff
@@ -4,13 +4,19 @@
...
 MainWindow::MainWindow() : QMainWindow()
 {
     view = new View(this);
     view -> settings() -> setAttribute(QWebEngineSettings::LocalStorageEnabled, false);
     view -> settings() -> setAttribute(QWebEngineSettings::ShowScrollBars, false);
     setCentralWidget(view);
+
+    QRect screenGeometry = QGuiApplication::primaryScreen()->geometry();
+    setGeometry(screenGeometry);
+    showFullScreen();
```
nicomiguelino, line 19: Make use of macros, which gets resolved before compilation.

PR #2270, opened 2025-04-18
fix: cache from the Docker Hub registry when building images

nicomiguelino:
### Description

- Attempted to reduce Docker image build times by caching from Screenly's Docker Hub registry

### Checklist


nicomiguelino:
It's still taking almost an hour for the Pi builds (pi1, pi2, pi3, pi4, and pi4-64) to finish. Let's take a step back and identify the real root cause behind image builds not caching.

PR #2396, opened 2025-07-10
fix: add WebView support for preloading web pages

### LINKED ISSUE #2381: [FEATURE] Preload webpages
### Description

One of the things I use Anthias for is to display current weather information. Three of the webpages are hosted locally on my network, but they pull information from NOAAs API. The information in the API updates a few times per hour, but every time the page is loaded it pulls the data from the external API. Each time a page loads, it takes a few seconds for the information to populate on the screen. I would like the information to be visible as soon as the page loads.

I also display one completely external webpage, the live radar from weather.gov. This page takes about 10 seconds to fully load. I would like this to be loaded and ready to go when Anthias displays this page.

I have 2 instances of each page in the rotation. The URLs are identical, but they display twice in the signage rotation.

Copies of the locally hosted pages:
[Current Conditions](https://shelbylight.house/wx/current.html)
[Extended Forecast](https://shelbylight.house/wx/forecast.html)
[Weather Alerts](https://shelbylight.house/wx/alerts.html)

External page:
[Live Radar](https://radar.weather.gov/?settings=v1_eyJhZ2VuZGEiOnsiaWQiOiJ3ZWF0aGVyIiwiY2VudGVyIjpbLTg2LjQwNCwzOS42NzldLCJsb2NhdGlvbiI6Wy04NS43NzIsMzkuNTEzXSwiem9vbSI6OC41Mjc2NzQ2MTQxMDgwNjEsImxheWVyIjoiYnJlZl9xY2QifSwiYW5pbWF0aW5nIjp0cnVlLCJiYXNlIjoiZGFya2NhbnZhcyIsImFydGNjIjpmYWxzZSwiY291bnR5Ijp0cnVlLCJjd2EiOnRydWUsInJmYyI6ZmFsc2UsInN0YXRlIjp0cnVlLCJtZW51Ijp0cnVlLCJzaG9ydEZ1c2VkT25seSI6ZmFsc2UsIm9wYWNpdHkiOnsiYWxlcnRzIjowLjgsImxvY2FsIjowLjYsImxvY2FsU3RhdGlvbnMiOjAuOCwibmF0aW9uYWwiOjAuNn19)

### Solution

I would like for the pages to either preload in the background for a user-defined time (eg 15 seconds) before it is displayed. Or since my pages automatically refresh the data, I would like the pages to load, and stay loaded in the background, so they are ready to go every time they are displayed.

Since there are multiple instances of a page in the rotation, it should only preload each page once, but allow both instances to pull from the same preload, as not to creat unneeded redundancy in RAM.

I am using Anthias on an HP Mini x86 PC - Intel i5-9500T, 16GB RAM, network connected via Ethernet to 200MB/s synchronous internet, These specs should be more than enough power to handle the pages.

nicomiguelino:
### Issues Fixed

- Fixes #2381

### Description

- Updates the WebView so that the next web asset to be displayed will be loaded before it gets shown on the screen

### Checklist

### DIFF
--- a/webview/src/view.h
+++ b/webview/src/view.h
@@ -29,12 +29,16 @@ protected:
 private slots:
     void handleAuthRequest(QNetworkReply*, QAuthenticator*);
     void updateMovieFrame();
+    void onWebPageLoadFinished(bool ok);
+    void onWebPageLoadProgress(int progress);
 
 private:
     bool tryLoadAsAnimatedGif(const QByteArray& data);
     void loadAsStaticImage(const QByteArray& data);
     void scheduleNextFrame();
     void setupAnimation();
+    void switchToNextWebView();
+    void resetWebViewStates();
 
     QWebEnginePage* pre_loader;
     QEventLoop pre_loader_loop;
@@ -44,4 +48,13 @@ private:
     QMovie* movie;
     QTimer* animationTimer;
     bool isAnimatedImage;
+
+    // Dual web view system
+    QWebEngineView* webView1;
+    QWebEngineView* webView2;
+    QWebEngineView* currentWebView;
+    QWebEngineView* nextWebView;
+    bool isWebView1Active;
+    bool nextWebViewReady;
+    int nextWebViewProgress;
 };
--- a/webview/src/view.h
+++ b/webview/src/view.h
@@ -54,7 +54,5 @@ private:
     QWebEngineView* webView2;
     QWebEngineView* currentWebView;
     QWebEngineView* nextWebView;
-    bool isWebView1Active;
     bool nextWebViewReady;
-    int nextWebViewProgress;
 };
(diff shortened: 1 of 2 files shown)


webview/src/view.h
```diff
@@ -44,4 +48,13 @@ private slots:
...
     bool isAnimatedImage;
+
+    // Dual web view system
+    QWebEngineView* webView1;
+    QWebEngineView* webView2;
+    QWebEngineView* currentWebView;
+    QWebEngineView* nextWebView;
+    bool isWebView1Active;
+    bool nextWebViewReady;
+    int nextWebViewProgress;
```
nicomiguelino, line 59: Remove this since it's not being used.

webview/src/view.h
```diff
@@ -44,4 +48,13 @@ private slots:
     QMovie* movie;
     QTimer* animationTimer;
     bool isAnimatedImage;
+
+    // Dual web view system
+    QWebEngineView* webView1;
+    QWebEngineView* webView2;
+    QWebEngineView* currentWebView;
+    QWebEngineView* nextWebView;
+    bool isWebView1Active;
```
nicomiguelino, line 57: Remove this since it's not being used.

PR #2404, opened 2025-07-16
fix: add CORS handling in Django

nicomiguelino:
### Issues Fixed

- https://forums.screenly.io/t/error-with-create-asset/6476

### Description

- Installs `django-cors-headers`
- Configures CORS settings

### Checklist

### DIFF
--- a/anthias_django/settings.py
+++ b/anthias_django/settings.py
@@ -68,11 +68,13 @@ INSTALLED_APPS = [
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'dbbackup',
+    'corsheaders',
 ]
 
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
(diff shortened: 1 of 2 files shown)


vpetersson:
Closing as stale.

The target files no longer exist on master:
- `anthias_django/settings.py` moved to `src/anthias_server/django_project/settings.py` (#2817, src/ layout refactor)
- `requirements/requirements.txt` replaced by `pyproject.toml` + `uv.lock`

Architecture has also shifted since this was opened: the React frontend was replaced with server-rendered Django templates + HTMX/Alpine (#2818), so browser → API calls are now same-origin and don't trigger CORS. The proposed `CORS_ALLOWED_ORIGINS = ['http://anthias']` only whitelisted the in-Docker hostname, which wouldn't have helped a browser hitting the device by LAN IP or `.local` name anyway.

If the original forum report (Error with create asset) is still reproducible, the fix is likely elsewhere (CSRF, auth, MIME, or upload size) rather than CORS. Happy to reopen if there's a concrete cross-origin use case to support.

PR #2409, opened 2025-07-28
feat: Add support for x86 devices running balenaOS

nicomiguelino:
### Issues Fixed

- The connected display doesn't show the current active asset.
- The display is stuck at displaying the balenaOS logo.

### Description

- This pull request depends on #2411
- Switched from using framebuffer to X11 for x86 devices

### Checklist (Detailed)

- [x] Playback of image assets works.
- [x] Playback of web assets works.
- [x] Playback of video assets works.

### Checklist (General)

- [x] I have performed a self-review of my own code.
- [x] New and existing unit tests pass locally and on CI with my changes.
- [x] I have done an end-to-end test for Raspberry Pi devices.
- [x] I have tested my changes for x86 devices.
- [x] I added a documentation for the changes I have made (when necessary).

PR #2411, opened 2025-07-28
feat: force full screen for x86

nicomiguelino:
### Issues Fixed

- Using frame buffer for display wouldn't work for x86 devices running balenaOS.
- With #2409, X11 will be used for display.
- Even though it works, the assets are not being displayed properly. They are scaled down to about 5%.

### Description

- This pull request forces full screen in the web view.

### Checklist

### DIFF
--- a/webview/src/mainwindow.cpp
+++ b/webview/src/mainwindow.cpp
@@ -1,5 +1,7 @@
 #include <QStandardPaths>
 #include <QWebEngineSettings>
+#include <QGuiApplication>
+#include <QScreen>
 
 #include "mainwindow.h"
 #include "view.h"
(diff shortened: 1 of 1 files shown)


webview/src/mainwindow.cpp
```diff
@@ -11,6 +13,10 @@ MainWindow::MainWindow() : QMainWindow()
     view->webView->settings()->setAttribute(QWebEngineSettings::LocalStorageEnabled, true);
     view->webView->settings()->setAttribute(QWebEngineSettings::ShowScrollBars, false);
     setCentralWidget(view);
+
+    QRect screenGeometry = QGuiApplication::primaryScreen()->geometry();
+    setGeometry(screenGeometry);
+    showFullScreen();
```
nicomiguelino, line 19: Check if this works for Pi devices as well. If not, wrap this inside macros.

webview/src/mainwindow.cpp
```diff
@@ -1,5 +1,7 @@
 #include <QStandardPaths>
 #include <QWebEngineSettings>
+#include <QGuiApplication>
+#include <QScreen>
```
nicomiguelino, line 4: Check if this works for Pi devices as well. If not, wrap this inside macros.

nicomiguelino (commented):
Check if this works for Pi devices as well. If not, wrap this inside macros.

PR #2568, opened 2025-11-12
feat: add support for rotating content (videos, images, web pages)

nicomiguelino:
### Issues Fixed

- https://forums.screenly.io/t/how-do-i-rotate-the-screen-to-play-portrait-video-1080x1920-raspberry-pi-3-b-raspberry-pi-os-lite-64bit-anthias-v0-20-3/6557/15

### Description

- Add support for rotating content (videos, images, web pages)
- I still need to add a setting in the web interface that lets them control the rotation angle.

### Checklist


quinlan96:
Thank you for looking into this! I run Anthias on a few vertically mounted displays and have been wanting this feature for a while. The displays are all powered by Raspberry Pi 4B's (4GB RAM), so I'll share what I've learnt from experimenting with solutions for this hardware.

## WebView
I found [this patch](https://borkedlabs.com/blog/2015/06-01-qt5-linuxfb-rotation-for-lcds/) for QT5 which adds a `rotation` parameter to the linuxfb platform. Upon compiling the patched QT5 and rebuilding the anthias-viewer container, orientation can be controlled with the QT platform environment variable:
```
QT_QPA_PLATFORM=linuxfb:rotation={0,90,180,270}
```

[This gist](https://gist.github.com/quinlan96/76f67376b331dcef9e6b4e4a664779bf) contains a recreation of the patch I made for QT 5.15.14. I've been applying it from the `build_qt5.sh` script like so:
```
pushd "/src/qt$QT_MAJOR/qtbase"
patch -p1 < /webview/qt5_linuxfb_rotation.patch
popd
```
So far this patch has been working flawlessly with web page and images.

## Videos
I hadn't looked into rotating videos prior to now, but this feature request got me playing around with VLC options. Using your code I couldn't get the rotate filter to work, however I found a suggestion that some filters had better support when hardware decoding was disabled with the `--avcodec-hw=none` option. Again this didn't work for me within the viewer, but when applied to the standalone `vlc` command it finally bore fruit. The video was rotated, but the rotate filter preserved the original aspect ratio so the vertical image was squished into a 16:9 frame. I found the desired effect could be achieved with the transform filter instead:
```
vlc --avcodec-hw=none --video-filter=transform{type=90} file.mp4
```
Success!

As for why it won't work with `python-vlc`, apparently the underlying library `libvlc` [ignores this option when constructing a new instance](https://stackoverflow.com/questions/77661537/why-does-python-vlc-libvlc-drop-frames-in-fullscreen-on-a-raspberry-pi-4-b). This means it can be bypassed by applying the option directly to a media object.

In `media_player.py`:
```python
    def set_asset(self, uri, duration):
        media = self.instance.media_new(uri)
        media.add_option(':avcodec-hw=none')
        self.player.set_media(media)
        ...
```
The video-filter must still be passed with the Instance options:
```python
    def __get_options(self):
        ...
        return [
            f'--alsa-audio-device={self.get_alsa_audio_device()}',
            f'--video-filter=transform{{type={angle}}}',
        ]
```

I have also tested with `ffplay` and can confirm the `-vf` options work, although playblack on the Pi 4B 4GB is extremely choppy when compared to VLC, which can handle 1080p smoothly in any orientation.

tezza1971:
re: videos. Apprently exiftool can get the job done since Anthias player seems to respect the rotation flag. I haven't tested it on images yet. I'm gonna see if I can update the settings page to "Auto detect and rotate 9:16 content" [disabled|90|-90]. I want to add another settings option which rejects aspect ratios that are not 16:9 or 9:16 as well (for the OCD among us). Also one which rejects videos that are larger than full HD since noticed that UHD stresses the mini PC I am running it on. IDK. We will see.

vpetersson:
superseded by https://github.com/Screenly/Anthias/issues/2856

PR #2814, opened 2026-05-02
fix(upgrade): pin docker pull platform off DEVICE_TYPE

vpetersson:
### Issues Fixed

Reported on a Pi 4 running 64-bit Raspberry Pi OS Trixie with Docker Engine 29.4.2:

```
$ docker compose pull
Error response from daemon: no matching manifest for linux/arm/v8 in the manifest list entries: no match for platform in manifest: not found
```

`uname -m` is `aarch64`, `dpkg --print-architecture` is `arm64`, `docker info` reports `Architecture: aarch64`, and the GHCR manifest for `latest-pi4-64` exposes a clean `arm64/linux` entry — yet the daemon requests `linux/arm/v8` (32-bit ARMv8) and fails to match. `docker pull --platform linux/arm64 …` succeeds, confirming this is host-platform auto-detect on Docker 29.4.x mis-classifying the host.

### Description

`bin/upgrade_containers.sh` already derives `DEVICE_TYPE` from `/proc/device-tree/model`, so we know the correct platform without trusting Docker's host detection. Set `DOCKER_DEFAULT_PLATFORM` explicitly:

- `pi4-64`, `pi5` → `linux/arm64`
- `pi2`, `pi3` → `linux/arm/v7`
- `x86` → `linux/amd64`

The existing `sudo -E docker compose pull` propagates the export. No build-pipeline / image-builder change needed — the published manifests are correct.

### Checklist

### DIFF
--- a/bin/upgrade_containers.sh
+++ b/bin/upgrade_containers.sh
@@ -53,6 +53,19 @@ else
     exit 1
 fi
 
+# Pin the pull platform off DEVICE_TYPE rather than letting Docker
+# auto-detect from the host. Docker 29.4.x on Trixie has been observed
+# advertising `linux/arm/v8` (32-bit ARMv8) on aarch64 Pi 4 hosts, which
+# doesn't match the single-platform `linux/arm64/v8` manifest we publish
+# for `pi4-64` and breaks `docker compose pull`. We already know the
+# correct platform from DEVICE_TYPE, so just set it explicitly. `-E`
+# below propagates this into the sudo'd `docker compose` invocations.
+case "$DEVICE_TYPE" in
+    pi5|pi4-64) export DOCKER_DEFAULT_PLATFORM="linux/arm64" ;;
+    pi3|pi2)    export DOCKER_DEFAULT_PLATFORM="linux/arm/v7" ;;
+    x86)        export DOCKER_DEFAULT_PLATFORM="linux/amd64" ;;
+esac
+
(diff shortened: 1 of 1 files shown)


vpetersson:
Superseded by the ansible-side fix on \`fix/ansible-docker-arch-reconciliation\` (PR coming).

Diagnosis got further than this PR: the \`linux/arm/v8\` platform error wasn't a Docker host-detection bug, it was a downstream symptom of \`docker-ce:armhf\` getting installed on aarch64 Pi 4 hosts. Root cause is a regression introduced in #2779 — the Trixie upgrade renamed \`DEVICE_TYPE\` from \`pi4\` to \`pi4-64\` in \`bin/install.sh\` / \`bin/upgrade_containers.sh\` / \`ansible/site.yml\` validator, but missed the conditional in \`ansible/roles/system/tasks/docker.yml\`, which still checks \`device_type in ['pi4', 'pi5']\`. Every Pi 4 host that ran \`bin/run_upgrade.sh\` after #2779 merged on 2026-04-30 had its docker-ce silently swapped from arm64 to armhf — confirmed via apt history on a live device (arm64 install on 2026-04-28, armhf re-install on 2026-05-02).

This PR's \`DOCKER_DEFAULT_PLATFORM\` pin would only have worked around the manifest match step; the seccomp filter on an armhf daemon would still SIGSYS containers at runtime regardless. Ansible-side fix is the correct layer.

PR #2822, opened 2026-05-04
fix(youtube): move download to a Celery task and detect URLs in frontend

vpetersson:
## Summary

YouTube assets had two compounding bugs. The frontend Add‑modal silently classified YouTube URLs as `webpage` (the file‑extension heuristic in `assets_create` had no YouTube branch), so the viewer loaded them in an iframe that YouTube refused to embed. The API path that *did* know about `youtube_asset` ran the download in a daemon `Thread` inside uvicorn — lost on restart, blocked the request on two upfront `yt-dlp` shellouts, and silently swallowed yt-dlp's exit code, so a failed download advertised an empty mp4 as ready.

This PR:

- Adds **`src/anthias_common/youtube.py`** as the single source for `is_youtube_url`, the local mp4 destination path, and the Celery dispatch shim.
- Replaces `YoutubeDownloadThread` with **`download_youtube_asset` Celery task**, using `yt_dlp` as a Python library — one `extract_info` call instead of three shellouts plus tab‑separated stdout parsing that YouTube titles can break. `_DownloadYoutubeTask.on_failure` clears `is_processing` so a permanent yt‑dlp failure no longer strands the row at "Processing" forever.
- Routes the **frontend `assets_create`** through `is_youtube_url` before the extension heuristic, so YouTube URLs land as a video asset with `is_processing=True` and the task dispatched.
- Switches both API serializers to **exact‑match** `mimetype == 'youtube_asset'` (substring match also fired on `not_youtube_asset` and crashed on a missing mimetype). The source URI is stashed on `_pending_youtube_uri` for the view to dispatch after `Asset.objects.create`. `url_fails` is skipped for in‑flight YouTube rows whose local mp4 doesn't exist yet.
- Removes `download_video_from_youtube` and `YoutubeDownloadThread` from `anthias_common/utils.py`, plus the unused `Thread` / `subprocess` / `os.path` imports.
- Tests: `tests/test_youtube_helper.py` for URL classification edges, 7 cases in `tests/test_celery_tasks.py` for the task (success, sub‑second floor, missing‑duration, missing/finalized rows, DownloadError propagation, on_failure cleanup), 2 cases in `tests/test_template_views.py` for the frontend create routing.

## Test plan

Verified end‑to‑end through the full compose stack (anthias‑server + anthias‑celery + redis on the dev compose, bind‑mounted to this branch):

- [x] `pytest -m "not integration"` — 544 pass, 0 fail
- [x] `ruff check .` — clean
- [x] API v2 POST with `?v=jNQXAC9IVRw` ("Me at the zoo", 19s) → finalized in 1.8s, name + duration + 614 KiB mp4
- [x] API v2 POST with `?v=aqz-KE-bpKQ` (Big Buck Bunny, 10:35) → finalized in 16.7s, 288 MB mp4
- [x] API v2 POST with `?v=dQw4w9WgXcQ` (Rickroll, 3:33) → finalized in 5.7s, 84 MB mp4
- [x] API v2 POST with bogus video id → `DownloadError` raised, `on_failure` cleared `is_processing`, no orphan files
- [x] Frontend `/assets/new/` POST with a YouTube URL → row created with `mimetype=video`, `is_processing=true`, task dispatched, finalized

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -222,6 +222,8 @@ module = [
     "pydbus",
     "sh",
     "vlc",
+    "yt_dlp",
+    "yt_dlp.*",
 ]
 ignore_missing_imports = true
 
--- a/src/anthias_common/utils.py
+++ b/src/anthias_common/utils.py
@@ -7,10 +7,8 @@ import re
 import socket
 import string
 from datetime import datetime, timedelta
-from os import getenv, path, utime
+from os import getenv, utime
 from platform import machine
-from subprocess import call, check_output
-from threading import Thread
 from time import sleep
 from typing import Any
 from urllib.parse import urlparse
@@ -27,7 +25,6 @@ from tenacity import (
     wait_fixed,
 )
 
-from anthias_server.app.models import Asset
 from anthias_server.settings import settings
 
 arch = machine()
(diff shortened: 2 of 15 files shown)


vpetersson:
Addressed in e09912f4 — wired `dispatch_download` into the v1.2 create view (`src/anthias_server/api/views/v1_2.py`) and added a parametrized `test_create_youtube_asset_dispatches_celery_task` across v1/v1_1/v1_2/v2 so the same omission for any version fails CI.

vpetersson:
Addressed in e09912f4 — `download_youtube_asset` now uses `asset.uri` (with `youtube_destination_path()` as a defensive fallback). New test `test_download_youtube_asset_writes_to_persisted_uri` pins this by writing a non-default uri and asserting yt-dlp's `outtmpl` matches.

vpetersson:
Addressed in 2ea559e3 — the post-success update now sets `uri = location` when `asset.uri` was falsy on entry, so the defensive fallback path also stamps the resolved path back on the row. Operator-set `uri` values (the normal path) are left alone. New test `test_download_youtube_asset_backfills_uri_when_row_uri_missing` covers this.

PR #2883, opened 2026-05-12
refactor(viewer): consolidate Qt6 boards onto cage + Wayland

vpetersson:
> **Note:** Stacked on top of #2879 (the generic-arm64 PR), so the diff here is the single consolidation commit. Base is `worktree-jiggly-sauteeing-stonebraker`; rebase onto `master` once #2879 lands.

### Description

Pi4-64, Pi5, x86, and arm64 now all run the viewer under `cage` (a wlroots kiosk compositor) with Qt on `QT_QPA_PLATFORM=wayland` and mpv on `--vo=gpu --gpu-context=wayland`. Previously only x86 and (per #2879) generic arm64 took this path; Pi4-64 and Pi5 ran Qt linuxfb + mpv `--vo=drm` directly on KMS. Folding all four onto one display stack drops a per-board branch each in `image_builder`, `Dockerfile.viewer`, `start_viewer.sh`, and `media_player.py`.

Notable per-file changes:

- **`tools/image_builder/utils.py`** — `cage` + `qt6-wayland` move out of the per-board branch into the shared `is_qt6` block (now in every Qt6 image, not just x86/arm64). `va-driver-all` stays x86-only (no VAAPI on ARM SoCs — Pi uses v4l2-request/m2m, Rockchip/Allwinner/Amlogic use V4L2 M2M too).
- **`docker/Dockerfile.viewer.j2`** — `ENV QT_QPA_PLATFORM=wayland` gated on `is_qt6` instead of `board in ('x86', 'arm64')`.
- **`bin/start_viewer.sh`** — `case "$DEVICE_TYPE" in x86|arm64|pi4-64|pi5)` now wraps the viewer in `cage` on all four boards. The render-GID mirror that previously only mattered on x86 (VAAPI / `/dev/dri/renderD128`) also applies on Pi — the GL context (V3D render node) needs the same access for `--vo=gpu --gpu-context=wayland`.
- **`src/anthias_viewer/media_player.py`** — `--vo=gpu --gpu-context=wayland` for every Qt6 board; the previous `--drm-mode=1920x1080@60` pin is dropped on Pi4-64/Pi5 (no-op under cage anyway, and GPU scaling handles 4K without the A72/A76 CPU zimg upscale that the pin was working around). `--vd-lavc-threads=4` stays.
- **`tests/test_media_player.py`** — assertions updated; new parametrised test covers all four Qt6 device types getting `--vo=gpu --gpu-context=wayland`.
- **`website/data/faq.yaml`** — two entries that claimed "no Wayland compositor in the stack" are corrected.

### Validation

On-device frame-drop comparison on a Pi4 Model B (Debian Trixie, DEVICE_TYPE=pi4-64), 30 s of 1080p H.264 over `mpv --hwdec=auto-safe` from inside the viewer container:

| VO flags | Asset 1 drops/30s | Asset 2 drops/30s |
|---|---|---|
| `--vo=drm --drm-mode=1920x1080@60` (current) | 59–75 | 19–20 |
| `--vo=gpu --gpu-context=drm` (no cage) | 3–6 | 3 |

`--vo=gpu --gpu-context=wayland` under cage measured separately on the rebuilt viewer image (results in the PR comments). decoder-frame-drop-count was 0 in every run — the drops are all on the VO side, which is exactly the path the move from CPU zimg upscale to GPU scaling improves.

### Checklist

### DIFF
--- a/bin/start_viewer.sh
+++ b/bin/start_viewer.sh
@@ -116,25 +116,34 @@ fi
 # we just set; -E alone is subject to env_check / env_delete and is not
 # guaranteed for XDG_* on Debian's default sudoers.
 #
-# x86 boards run under `cage`, a kiosk wlroots compositor, because
-# balenaOS x86 doesn't expose /dev/fb0 (Qt's linuxfb plugin has nothing
-# to draw to) and there's no host display server. cage acquires DRM
-# master as root, exports WAYLAND_DISPLAY for its child, and exits when
-# the child exits — so the existing kill -0 watchdog below still works.
-# The inner sudo drops back to the viewer user; WAYLAND_DISPLAY has to
-# be added to --preserve-env to survive sudo's env scrub.
-if [ "$DEVICE_TYPE" = "x86" ] || [ "$DEVICE_TYPE" = "arm64" ]; then
+# All Qt6 boards (pi4-64, pi5, x86, arm64) run under `cage`, a kiosk
+# wlroots compositor. This is required on x86 (balenaOS doesn't expose
+# /dev/fb0) and on generic arm64 (Armbian boots Mesa straight to KMS
+# with no fbdev); the Pi boards historically used Qt linuxfb + mpv
+# --vo=drm directly on KMS, but were consolidated onto the same
+# Wayland path so the four Qt6 images share one display stack. cage
+# acquires DRM master as root, exports WAYLAND_DISPLAY for its child,
+# and exits when the child exits — so the existing kill -0 watchdog
+# below still works. The inner sudo drops back to the viewer user;
+# WAYLAND_DISPLAY has to be added to --preserve-env to survive sudo's
+# env scrub. Qt5 boards (pi2/pi3) stay on the legacy linuxfb path.
+case "$DEVICE_TYPE" in
+    x86|arm64|pi4-64|pi5)
     # /dev/dri/renderD128 carries the host's `render` group, whose
     # numeric GID is distro-dependent (typically 992 on Debian/Ubuntu,
-    # 109 elsewhere) and not present in the container's /etc/group.
-    # Without membership the `viewer` user can open card0 (group
-    # `video`, GID 44 — already a member) but not the render node, and
-    # VAAPI silently fails with "wayland: failed to open
-    # /dev/dri/renderD128". mpv then falls back to software decode and
-    # frames drop at 1080p on entry-level x86. Mirror the host GID
-    # into the container as a synthetic `host-render` group and add
-    # `viewer` to it, so the supplementary group list `sudo -u viewer`
-    # later resolves from /etc/group already includes render access.
+    # 109 elsewhere, 106 on Pi OS Bookworm) and not always present in
+    # the container's /etc/group. Without membership the `viewer`
+    # user can open card0 (group `video`, GID 44 — already a member)
+    # but not the render node. On x86 that means VAAPI silently fails
+    # with "wayland: failed to open /dev/dri/renderD128" and mpv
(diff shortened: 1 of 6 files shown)


vpetersson:
### On-device validation blocked the Pi4-64/Pi5 consolidation

Tested the rebuilt viewer image on a real Pi4 Model B (Debian Trixie, 4K display via HDMI). Same 1080p H.264 asset, same 30-s window, frame drops via mpv `\${frame-drop-count}`:

| Path | mpv flags | Drops / 30s |
|---|---|---|
| Current production | `--vo=drm --drm-mode=1920x1080@60 --vd-lavc-threads=4 --hwdec=auto-safe` | 59–75 |
| GPU output, no cage | `--vo=gpu --gpu-context=drm --hwdec=auto-safe` | **3–6** |
| **This PR (cage + Wayland)** | `--vo=gpu --gpu-context=wayland --vd-lavc-threads=4 --hwdec=auto-safe` | **738+, slow-motion playback** |

`decoder-frame-drop-count` was 0 in every run — the regression is on the VO side.

#### Why this regresses on Pi

1. **The Pi V3D can't composite 4K in real time on top of a separate Wayland surface.** `cage` selects the connector's preferred mode (4K on this TV) and mpv renders into a Wayland surface that cage composites and scans out. The current `--drm-mode=1920x1080@60` pin sidesteps this by running the connector at 1080p; under cage that flag is a no-op (cage holds DRM master).
2. **mpv 0.40 in Debian Trixie has no `v4l2request` hwdec** — only `v4l2m2m-copy`, which falls back to software decode under cage in this image (`hwdec=no` confirmed via `\${hwdec-current}`). No DMA-BUF zero-copy path is available.
3. **`--vo=dmabuf-wayland` is not viable** — it would zero-copy from decoder to wayland surface, but it has no hwdec to receive from (see point 2), and the existing code comment notes it segfaults under cage on x86 anyway.

#### Options

- **(A)** Drop Pi4-64/Pi5 from this PR. Keep x86+arm64 on cage+Wayland (already there from #2879). Effectively close this PR.
- **(B)** Force a 1080p output mode on Pi via kernel cmdline (`video=HDMI-A-1:1920x1080@60` in `/boot/firmware/cmdline.txt`), so cage runs at 1080p and the V3D has the same upscale headroom it had with `--drm-mode`. Requires an Ansible/install change to ship.
- **(C)** Wait for newer mpv (`v4l2request` hwdec + `drm-prime` interop) and revisit.

Leaning toward (A) — the consolidation goal hinged on Pi being able to keep up, and on this hardware + mpv combo it can't. Marking the PR draft pending direction.

#### Side-finding worth keeping

Even without cage, just switching the Pi `--vo=drm` → `--vo=gpu --gpu-context=drm` (offload scaling from CPU zimg to V3D) cuts drops from 59–75 → 3–6 on this Pi4. That's a separate small win worth landing on its own if we don't ship the full consolidation.

(Tested on a 4K-connected Pi4-64 Rev 1.5; the Pi5 has roughly 2× the V3D throughput so it might just squeak through, but I'd want the same test there before assuming.)

vpetersson:
Closing. Folded into #2885, which now consolidates Pi 5 / x86 / arm64 onto cage + Wayland (Pi 4 stays on linuxfb — on-device testing showed the V3D 6.0 can't keep up with cage's composite pass on top of mpv, even with a 1080p HDMI mode pin).

The base of this PR (`worktree-jiggly-sauteeing-stonebraker`) merged to master as #2879, so the cage path is already live for x86 / arm64. #2885 picks up Pi 5 (and a Pi 4 perf improvement on the linuxfb path) on top of it.

PR #2902, opened 2026-05-14
docs(faq): running Anthias behind a custom reverse proxy

vpetersson:
### Issues Fixed

Follow-up doc to PR #2901 — surfaces the `CSRF_TRUSTED_ORIGINS` knob (and the cleaner Host-preservation path) on the website FAQ so operators hitting the symptom can find the fix without spelunking through `settings.py`.

### Description

New FAQ entry under **Operations**, right after "How do I enable HTTPS?".

Names the symptom (`Origin checking failed` 403 on every POST after wiring up nginx / Apache / IIS), gives the one-line Host-preservation directive for the five reverse proxies operators actually deploy, and documents `CSRF_TRUSTED_ORIGINS=https://signage.example.com` as the escape hatch when the proxy can't be reconfigured. Notes that the bundled `./bin/enable_ssl.sh` Caddy sidecar already does the right thing — the entry is only relevant for third-party proxy setups.

No code change — `website/data/faq.yaml` only.

### Checklist

### DIFF
--- a/website/data/faq.yaml
+++ b/website/data/faq.yaml
@@ -178,6 +178,34 @@
       answer: |
         From `~/anthias/`, run `./bin/enable_ssl.sh`. The default mode uses Caddy's built-in local CA, which is fine for IP-based LAN access. For Let's Encrypt or bring-your-own certificate, see the [TLS / SSL section in the docs](/docs/#tls-ssl).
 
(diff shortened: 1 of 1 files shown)


vpetersson:
Folding the FAQ entry into the parent PR #2901 — same logical change.

PR #2905, opened 2026-05-15
feat(viewer,webview): embed QtMultimedia in AnthiasWebview, eliminate two-process DRM contention + Pi 4 drops

### LINKED ISSUE #2904: viewer: embed libmpv in AnthiasWebview to eliminate two-process DRM-master contention
## Background

The viewer currently runs as two processes on the same framebuffer:

- ``AnthiasWebview`` (C++/Qt6, holds the framebuffer for web + image rendering)
- ``mpv`` (subprocess.Popen'd by ``MPVMediaPlayer``, takes ``--vo=drm`` and does its own DRM-master juggling for video assets)

When the operator's board has a 4K-EDID-reporting display, the contention between these two processes for ``/dev/fb0`` / DRM master drops frames during video playback even when the HW decode path is engaged. Measured on a real Pi 4 with the BBB test bed during PR #2885 validation: 600-2800 vo drops per 60 s clip on H.264 (``v4l2m2m-copy``) and HEVC (``drm-copy``) variants.

The same mpv flags, on the same hardware, **with no Qt webview competing for the framebuffer**, get 0 drops. The drops are infrastructure contention, not a mpv configuration issue.

## Proposed fix (Option A from the PR review)

Embed libmpv directly inside ``AnthiasWebview`` so a single Qt process owns the framebuffer and renders both web pages and video into the same Qt window.

### Concrete changes

1. **``src/anthias_webview/``** — C++ side:
   - Add ``libmpv-dev`` + ``libqt6openglwidgets6-dev`` build deps (apt — see the Conan discussion below).
   - New ``VideoView : QOpenGLWidget`` class wrapping an ``mpv_render_context`` (reference: ``mpv-examples/libmpv/qt_opengl/``). ~200 LOC.
   - Add ``playVideo(QString uri)`` + ``stopVideo()`` D-Bus slots on ``MainWindow``.
   - ``View`` switches its central widget between ``QWebEngineView`` (web), ``QLabel`` (image), and ``VideoView`` (video) based on which slot was invoked.
   - mpv properties (``hwdec``, ``video-sync``, ``audio-device``, ``drm-mode``, etc.) set via ``mpv_set_property_string``, replicating today's per-codec dispatch logic.

2. **``docker/Dockerfile.viewer.j2``** — Pi 4 specifically:
   - Change ``QT_QPA_PLATFORM=linuxfb`` to ``eglfs`` so Qt has the OpenGL context libmpv needs to render into.
   - Pi 5 / x86 / arm64 stay on cage+wayland (already have GL inside cage).

3. **``src/anthias_viewer/media_player.py``** — Python side:
   - ``MPVMediaPlayer`` replaces ``subprocess.Popen`` with a ``QDBusConnection`` call to the ``anthias.webview`` service's ``playVideo`` slot.
   - Per-codec hwdec probe (``_pi_hwdec_for_uri``) sends its result alongside the URI or via a prior ``setPlaybackOptions`` slot.
   - ``stop()`` calls ``stopVideo()`` D-Bus.

4. **Startup ordering** — Python already waits for ``Anthias service start`` on the webview stdout (current D-Bus readiness signal). Same pattern works for video readiness.

### Package management note

The right libmpv build per board comes from each board's apt repo:
- Pi 4 / Pi 5 / Rock Pi 4 → ``archive.raspberrypi.com`` (+rpt1) ships libmpv with ``--enable-v4l2-request`` for the HW decoders the viewer mpv binary already uses.
- x86 → Debian apt ships libmpv with ``--enable-vaapi``.

Each board's per-board viewer image already does ``apt install`` from the correct source, so linking against ``libmpv-dev`` from the same apt run gets the byte-identical backend as the system ``mpv`` binary. Conan was considered but adds machinery to pin a single libmpv version across boards, when we genuinely want per-board variance (v4l2_request on arm, vaapi on x86). qmake → CMake migration is worth doing eventually but is a separate scope.

## Why this isn't fixed in PR #2885

- It's a C++ architecture change in ``src/anthias_webview/`` — distinct scope from the codec gate work in #2885.
- It requires Pi 4 to switch ``QT_QPA_PLATFORM`` (``linuxfb`` → ``eglfs``), which needs its own perf retest across boards.
- The frame-drop issue is pre-existing in master (master with master's mpv flags measured 43 vo drops / 30 s in the same isolation test; #2885's flags actually do *better* in isolation). #2885 doesn't regress vs master; this issue tracks the deeper fix.

## Acceptance criteria

- ``AnthiasWebview`` builds with libmpv linked; image stays per-board.
- Video playback on each of Pi 4, Pi 5, Rock Pi 4, x86 with the BBB test bed produces:
  - The same hwdec banner as today (``v4l2m2m-copy`` / ``drm-copy`` / ``vaapi-copy`` per board+codec).
  - **< 10 vo drops per 60 s clip** at 1080p30 H.264 / HEVC and 1080p60 H.264 / HEVC on every board.
- Pi 4 specifically: ``QT_QPA_PLATFORM=eglfs`` confirmed working with V3D 6.0; no regression in image / web-page rendering.
- The codec gate from #2885 still functions — rejected codecs surface the same banner + recipe in the Edit Asset modal.

## Estimated effort

~2-4 days including per-board perf retest. C++ work is ~300-500 LOC; the rest is build wiring and Python D-Bus glue.

## Related

- PR https://github.com/Screenly/Anthias/pull/2885 (codec gate + viewer hwdec dispatch; lands the contract this issue's fix implements behind)

vpetersson:
### Issues Fixed

Closes https://github.com/Screenly/Anthias/issues/2904. Built on top of #2885 (now merged); base branch is `master`.

### Description

The viewer's Qt6 video path was two processes contending for one framebuffer: ``AnthiasWebview`` (Qt) holding the surface, and an external ``mpv`` subprocess painting video via ``--vo=drm`` (linuxfb) or ``--vo=gpu --gpu-context=wayland`` (cage). PR #2885 documented 600-2800 vo drops per 60 s clip on Pi 4 even with HW decode engaged.

This PR moves video playback **inside** AnthiasWebview's Qt process via QtMultimedia (``QMediaPlayer`` + ``QGraphicsVideoItem`` inside a ``QGraphicsView`` / ``QGraphicsScene``). Single Qt process, single GL/Wayland surface, no more mpv subprocess.

#### Implementation arc (preserved from the branch history, replayed in this single commit)

I tried two embedding approaches; the second is what's shipping:

1. **libmpv via ``mpv_render_context``** — engaged HW decode correctly on every board (verbose mpv logs confirm ``drm-copy`` engaged on rpi-hevc-dec via ``/dev/video19``, ``v4l2m2m-copy`` on bcm2835-codec via ``/dev/video10``) but **Pi 4 frame drops stayed at 600-2800/60 s**. Root cause: libmpv-render uploads decoded frames to a GL texture, then ``QOpenGLWidget`` blits that into an offscreen FBO, then Qt's compositor blits the FBO into the eglfs window — three GPU passes per frame on a V3D 6.0 that can't sustain that at 60 fps. A ``QOpenGLWindow`` workaround to bypass the FBO crashed because eglfs is single-native-window-per-process.

2. **QtMultimedia ``QMediaPlayer`` + ``QGraphicsVideoItem``** — ships. The graphics-item path was picked over ``QVideoWidget`` so the ``video-rotate`` option actually rotates the displayed frames (``QGraphicsItem::setRotation`` is honoured by the painter, while ``QVideoWidget`` has no rotation property — the first review caught a ``setProperty("rotation", angle)`` shortcut as a silent no-op on Pi 4). Qt 6.5 dropped the upstream gstreamer media backend, so Debian Trixie ships only the ffmpeg-backed ``libffmpegmediaplugin.so`` — decode runs through libavcodec directly. The +rpt1 libavcodec packages pinned in #2885's ``_rpt1-ffmpeg-pin.j2`` carry ``--enable-v4l2-request`` / ``--enable-v4l2-m2m``, so the existing Pi-family hardware decoders (rpi-hevc-dec, bcm2835-codec, Hantro G2 on Pi 5, rkvdec on Rock Pi 4) all engage automatically without any per-codec dispatch from the application.

#### Real-device validation (BBB pack)

All three reachable Qt6 boards measured against the same 8-clip BBB pack (1080p / 4K, 30 / 60 fps, in H.264 and HEVC, each clip ~60 s @ ~8 Mbps). Drop counts come from the in-process counter in ``VideoView::onVideoFrameDelivered()`` (frames the QVideoSink received) compared to ``container_fps × elapsed_s``, written to ``/data/.anthias/playback-stats.log`` per play. The numbers below are the median across n full-clip plays in a ≥1200 s observation window (Pi 5 ran 5 cycles; Pi 4 / x86 ran 2-3 cycles given the 8-clip rotation).

| Asset | x86 | Pi 4 | Pi 5 | libmpv-era Pi 4 baseline |
|---|---|---|---|---|
| BBB 1080p30 H264 | 2 / 1798 (n=4) | 3 / 1799 (n=3) | — (HEVC-only gate) | 600-1500 |
| BBB 1080p30 HEVC | 2 / 1799 (n=4) | 2 / 1799 (n=3) | 2 / 1799 (n=9) | 562 |
| BBB 1080p60 H264 | 0 / 3597 (n=5) | **0** / 3594 (n=3) \* | — | **2973** |
| BBB 1080p60 HEVC | 3 / 3601 (n=4) | 3 / 3601 (n=3) | 2 / 3600 (n=10) | 2402 |
| BBB 4K30 H264 | 1 / 1798 (n=5) \* | 124 / 1464 (n=3) \* | — | 883 |
| BBB 4K30 HEVC | 1 / 1797 (n=4) | 2 / 1798 (n=3) | 1 / 1797 (n=9) | 1183 |
| BBB 4K60 H264 | 1 / 3597 (n=4) \* | 242 / 1713 (n=3) \* | — | 951 |
| BBB 4K60 HEVC | 3 / 3601 (n=4) | 2 / 3598 (n=4) | 2 / 3600 (n=9) | — |

Format: `dropped / expected (n=plays)`. `*` = derived from the viewer's STOP log line when ``asset_loop`` preempted the clip at its declared duration before ``QMediaPlayer::EndOfMedia`` could fire (``expected = position_ms × container_fps``).

Pi 4's **4K H.264** rows (124 / 242 drops) are a hardware ceiling: stream position at the 60 s wall-clock mark only reached 48.8 s / 28.5 s — bcm2835-codec running at ≈ 80 % / ≈ 47 % real time, not a presentation-side drop. Pre-existed PR #2885; the rest of the matrix shows the codec gate's intent (H.264 ≤ 1080p, HEVC at any resolution) already matches what the hardware can sustain. Out of scope for #2904.

#### 12 h burn-in (steady-state mixed media)

Asset rotation per board: 8 BBB clips + 3 webpages + 3 images, every 60 s for 12 h. Snapshots every 10 min: viewer restart count, RSS, ``END_FILE`` cumulative, early-stop (``STOP`` with ``elapsed_ms < 5 s``) cumulative.

| Board | Restarts | Early-stops | Clips played | Drops/clip 1st q → last q | RSS trajectory |
|---|---|---|---|---|---|
| x86 | 0 | 0 | 450 | 2 → 2 | 1540 MB flat 8 h → step to 1900-2100 MB on webpage add → flat |
| Pi 4 | 0 | 0 | 344 | 3 → 2 | 944-1261 MB clip-dependent → 2000-2068 MB → flat |
| Pi 5 | 0 | 0 | 660 | 2 → 2 | 1595 MB ±5 MB / 7 h → 1920 MB → flat 4 h |

The RSS step is QWebEngineView allocating its working set when webpage assets were added mid-burn; not a leak (post-shift trajectory is flat). Pi 5 is the cleanest signal — 7 h of <5 MB variance.

#### 3 h asset-churn (mutation under playlist reload)

Every 90 s the dev host toggles ``is_enabled`` on a random subset of the 14 assets (per board) via the API. 120 churn events × 3 boards = 1,080 toggle operations.

| Board | Churn ops | Restarts | **<100 ms stops** | 100-5000 ms stops | Clips finished | Drops/clip 1st q → last q |
|---|---|---|---|---|---|---|
| x86 | 120 | 0 | **0** | 57 | 112 | 1 → 1 |
| Pi 4 | 120 | 0 | **0** | 33 | 59 | 3 → 2 |
| Pi 5 | 120 | 0 | **0** | 90 | 156 | 1 → 1 |

The ``<100 ms stops`` column is the regression signature for the ``loadImage('null')`` issue (foreground video torn down ~66 ms after the first PLAYING event). Zero on all boards across 1,080 toggles — the fix holds under churn. 100-5000 ms stops are the legitimate asset_loop behaviour when a churn event disables a currently-playing clip.

#### Pi 4 rotation — end-to-end on real hardware

For each angle in {0, 90, 180, 270}: set ``screen_rotation`` in ``~/.anthias/anthias.conf``, restart viewer, play ``BBB 1080p30 H264``, grep ``playback-stats.log``:

| `screen_rotation` | LOADFILE option dict | PLAYING rotation field | Pass |
|---|---|---|---|
| 0 | `{audio-device=…}` (Python skips 0) | `rotation=0` | ✓ |
| 90 | `{audio-device=… video-rotate=90}` | `rotation=90` | ✓ |
| 180 | `{audio-device=… video-rotate=180}` | `rotation=180` | ✓ |
| 270 | `{audio-device=… video-rotate=270}` | `rotation=270` | ✓ |

Drop counts under rotation on Pi 4 BBB 1080p30 H.264:
- rotation=0: 0 / 573 (0%)
- rotation=90: 7 / 605 (1.2%)

Rotation 90 pays a small GPU cost for the transform (V3D 6.0 fillrate), but it's well under any acceptance bar and orders of magnitude below the 2973-drops/60 s libmpv subprocess baseline.

**Rock Pi 4** validation is deferred — the SSH banner exchange has been timing out on the testbed under upload load. The arm64 image (built off the same commit) carries the same ``VideoView`` code path the other three boards just validated.

#### Architecture changes

- **``src/anthias_webview/``** — new ``VideoView`` is a ``QWidget`` hosting ``QMediaPlayer`` + ``QGraphicsView`` + ``QGraphicsScene`` + ``QGraphicsVideoItem`` + ``QAudioOutput``. MainWindow exposes the same D-Bus surface (``playVideo`` / ``stopVideo`` / ``videoEnded``) the libmpv variant used, so the Python contract is untouched. Stats logger writes ``/data/.anthias/playback-stats.log`` (renamed from ``mpv-stats.log``), capped at 8 MB on viewer start.
- **``src/anthias_viewer/media_player.py``** — option dict shrinks from {``hwdec``, ``video-sync``, ``vd-lavc-threads``, ``audio-device``, ``video-rotate``} to just {``audio-device``, ``video-rotate``}. ``_PI_HWDEC_BY_CODEC``, ``_pi_hwdec_for_uri``, and ``_probe_video_codec`` (ffprobe subprocess) are gone — libavcodec's auto-selection covers the per-board decoder dispatch. ``_marshal_dbus_options`` picks the GLib.Variant signature by Python type so int / bool / float options round-trip cleanly.
- **``docker/Dockerfile.viewer.j2``** — drops ``libmpv2``, adds ``libqt6multimedia6``, ``libqt6multimediawidgets6``, ``qt6-multimedia-dev``, ``qt6-image-formats-plugins``. Pi 4 switches ``QT_QPA_PLATFORM`` from ``linuxfb`` to ``eglfs`` (QtMultimedia needs a GL context for the QGraphicsView painter). The eglfs KMS config pinning 1920×1080 is shipped at ``docker/eglfs-kms-pi4.json``.
- **``tests/test_media_player.py``** — argv assertions become D-Bus options-dict assertions; per-codec dispatch tests dropped; rotation / audio / proxy / VLC-fallback / marshal-type tests preserved. 63 cases green.
- **``src/anthias_webview/tests/``** (new) — QtTest unit suite for ``VideoView`` covering construction, stop idempotency, audio-device fallback, rotation (asserts on ``QGraphicsItem::rotation()`` — the painter's input, not a no-op dynamic property), and the cardinal-only normalisation defence. ``bin/test_webview_cpp.sh`` builds + runs under ``QT_QPA_PLATFORM=offscreen``. 8 cases green. CI integration is a follow-up.

#### Key fixes during validation

- **``view_image('null')`` was stopping the freshly-started video.** ``src/anthias_viewer/__init__.py`` calls ``view_image('null')`` AFTER ``media_player.play()`` to clear any leftover image/web background. My initial ``View::loadImage`` always called ``hideVideoSurface()``, which stopped the QMediaPlayer 66 ms after PLAYING fired. On Pi 4 this was invisible (decoder past init); on Pi 5 it killed the Hantro G2 mid-CMA-allocation for 4K60 HEVC — 0 frames in 60 s. Fix: skip ``hideVideoSurface`` for the ``'null'`` sentinel.
- **``QUrl(uri)`` with scheme-less local paths refused to load.** ``QMediaPlayer`` silently rejected ``/data/anthias_assets/abc.mp4`` because the QUrl had no scheme. Use ``QUrl::fromLocalFile`` for any path starting with ``/``.
- **``GLib.Variant`` wrap for D-Bus ``a{sv}``** — pydbus refuses to auto-coerce a Python scalar to ``GLib.Variant`` when the slot is declared with ``QVariantMap``. Wrap each option value via ``_marshal_dbus_options`` (type-aware, regression test covers str / int / bool / float).
- **``playStartedAt`` deferred to ``LoadedMedia``** so the elapsed-ms window measures real playback rather than decoder init — removes a ~5-10-frame inflation in clip-start drop counts.
- **``resolveAlsaDevice`` extracts ``CARD=<name>``** from the ALSA spec and matches the ``QAudioDevice`` id on that segment specifically; logs the resolved device id at INFO so multi-HDMI mismatches are visible from journalctl.

#### Notes for upgrade

Pre-PR-#2885 testbed images carry ``QT_QPA_PLATFORM=linuxfb`` hardcoded in the device's deployed ``docker-compose.yml``. The current ``docker-compose.yml.tmpl`` no longer pins the platform — ``bin/upgrade_containers.sh`` regenerates compose from the template, so devices on the new image automatically pick up the Dockerfile's per-board default.

#### Known pre-existing issue (not introduced here)

``view_video`` in ``src/anthias_viewer/__init__.py`` has a race: ``skip_event.clear() → wait(timeout=duration)``. If something else (e.g. an API state change firing ``reload``) sets ``skip_event`` between the ``clear()`` and ``wait()``, the very first asset after a viewer restart can be stopped immediately. Not introduced by this PR; worth filing separately.

### Checklist

### DIFF
--- a/bin/test_webview_cpp.sh
+++ b/bin/test_webview_cpp.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+# Build and run AnthiasWebview's QtTest unit tests.
+#
+# Requires Qt 6 (qt6-base-dev, qt6-multimedia-dev). The viewer
+# Docker image already ships those for the per-board builder stage
+# — running this inside that container is the canonical
+# environment.
+#
+# Usage: bin/test_webview_cpp.sh
+#
+# The tests run under ``QT_QPA_PLATFORM=offscreen`` so no real
+# display server is needed; QtMultimedia's playback pipeline won't
+# fully initialise (no rendering target) but the API surface plus
+# the QGraphicsVideoItem rotation transform are testable that way.
+# Decoder engagement + drop counts are exercised on real devices
+# via the BBB test bed. CI integration is a follow-up.
+
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+WEBVIEW_DIR="src/anthias_webview"
+BUILD_DIR="${WEBVIEW_DIR}/tests/build"
+
+mkdir -p "${BUILD_DIR}"
+pushd "${BUILD_DIR}" >/dev/null
+
+qmake6 ../tests.pro
+make -j"$(nproc)"
+
+# QTEST_MAIN's generated binary exits non-zero on any test failure.
+# ``QT_QPA_PLATFORM=offscreen`` skips connecting to a real display
+# server / framebuffer — the tests don't render, they exercise
+# the QMediaPlayer / QGraphicsVideoItem API surface plus the
+# rotation transform.
+QT_QPA_PLATFORM=offscreen ./AnthiasWebviewTests
+
+popd >/dev/null
--- a/docker/Dockerfile.viewer.j2
+++ b/docker/Dockerfile.viewer.j2
@@ -18,12 +18,22 @@ RUN --mount=type=cache,target=/var/cache/apt,id=qt6-builder-apt-{{ artifact_boar
     apt-get install -y --no-install-recommends \
         build-essential \
         qt6-base-dev \
+        qt6-multimedia-dev \
         qt6-webengine-dev
+# qt6-multimedia-dev: AnthiasWebview's VideoView wraps QMediaPlayer +
+# QGraphicsVideoItem (issue #2904, see
+# ``src/anthias_webview/AnthiasWebview.pro``). Qt 6.5 dropped the
+# upstream gstreamer media backend, so Debian Trixie ships only the
+# ffmpeg-backed ``libffmpegmediaplugin.so`` — decode runs through
+# libavcodec directly against the +rpt1 ``libav*`` packages pinned in
+# ``_rpt1-ffmpeg-pin.j2`` (which carry ``--enable-v4l2-request`` /
+# ``--enable-v4l2-m2m`` so libavcodec can drive the Pi-family
+# hardware decoders).
 COPY src/anthias_webview/AnthiasWebview.pro /src/anthias_webview/AnthiasWebview.pro
 COPY src/anthias_webview/src /src/anthias_webview/src
 COPY src/anthias_webview/res /src/anthias_webview/res
 WORKDIR /src/anthias_webview
-# webview is 3 .cpp files (~6s on x86 native, ~30s under arm64 QEMU);
+# webview is 4 .cpp files (~6s on x86 native, ~30s under arm64 QEMU);
 # adding ccache would only save a couple of seconds on incremental
 # rebuilds and the cache mount + apt install + PATH wiring would
 # outweigh the saving.
@@ -76,10 +86,35 @@ ENV QT_QPA_EGLFS_FORCE888=1
 # the viewer in cage; this env var tells Qt to load the qt6-wayland
 # platform plugin so it renders into cage's surface.
 ENV QT_QPA_PLATFORM=wayland
+{% elif board == 'pi4-64' %}
+# Pi 4 switched from ``linuxfb`` to ``eglfs`` for issue #2904:
+# QtMultimedia's video pipeline needs an OpenGL context for the
+# QGraphicsVideoItem painter, and ``linuxfb`` has none. ``eglfs``
+# (the Qt KMS/EGL platform) gives Qt a GL context paired with
+# KMS scanout. ``QT_QPA_EGLFS_KMS_CONFIG`` pins 1080p — the V3D
+# 6.0 can't composite Chromium + the video graphics view on top
+# of the connector's native 4K, and Qt's ``eglfs_kms``
+# integration reads this JSON at QPA init time. Validated on
+# real Pi 4 hardware: under eglfs QWebEngineProcess still spawns
+# its zygote + 2 renderers and loads pages, so the platform
+# switch doesn't regress web / image rendering.
+ENV QT_QPA_PLATFORM=eglfs
+ENV QT_QPA_EGLFS_KMS_CONFIG=/etc/anthias/eglfs-kms.json
+# bin/start_viewer.sh auto-detects QT_SCALE_FACTOR from
+# /sys/class/drm/*-*/modes — but the kernel mode list still
+# reports the connector's EDID-preferred mode (4K on most TVs),
+# even though eglfs is rendering into a 1080p framebuffer per the
+# KMS config above. Pinning scale=1 short-circuits the auto-detect
+# (which respects an existing value) so a 1920-wide page lays out
+# at 1920 physical px on the 1080p surface — i.e. CSS px ≡ physical
+# px, the same end-state the linuxfb+4K-with-scale=2 path produced.
+ENV QT_SCALE_FACTOR=1
+RUN mkdir -p /etc/anthias
+COPY docker/eglfs-kms-pi4.json /etc/anthias/eglfs-kms.json
 {% else %}
-# Pi 4 (Qt6) and Pi 2 / Pi 3 (Qt5) keep the classic linuxfb path —
-# the Pi 4 V3D 6.0 can't keep up with cage's composite pass on top
-# of video, so the viewer renders directly to the framebuffer.
+# Pi 2 / Pi 3 (Qt5) keep the classic linuxfb path — VLC paints
+# directly to the framebuffer and there's no libmpv embedding on
+# those boards.
 ENV QT_QPA_PLATFORM=linuxfb
 {% endif %}
 
--- a/docker/eglfs-kms-pi4.json
+++ b/docker/eglfs-kms-pi4.json
@@ -1,0 +1,15 @@
+{
+  "device": "/dev/dri/card1",
+  "hwcursor": false,
+  "pbuffers": true,
+  "outputs": [
+    {
+      "name": "HDMI1",
+      "mode": "1920x1080"
+    },
+    {
+      "name": "HDMI2",
+      "mode": "1920x1080"
+    }
+  ]
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -237,6 +237,8 @@ exclude = [
 module = [
     "cec",
     "channels_redis.*",
+    "gi",
+    "gi.*",
     "pillow_heif",
     "pydbus",
     "sh",
--- a/src/anthias_viewer/__init__.py
+++ b/src/anthias_viewer/__init__.py
@@ -20,6 +20,7 @@ from anthias_viewer.constants import SERVER_WAIT_TIMEOUT as SERVER_WAIT_TIMEOUT
 from anthias_viewer.constants import SPLASH_DELAY as SPLASH_DELAY
 from anthias_viewer.constants import SPLASH_PAGE_URL as SPLASH_PAGE_URL
 from anthias_viewer.constants import STANDBY_SCREEN as STANDBY_SCREEN
+from anthias_viewer import media_player as _media_player_module
 from anthias_viewer.media_player import MediaPlayerProxy
 from anthias_viewer.playback import (
     navigate_to_asset,
@@ -907,6 +908,12 @@ def setup() -> None:
 
     bus = pydbus.SessionBus()
     browser_bus = bus.get('anthias.webview', '/Anthias')
+    # MPVMediaPlayer calls AnthiasWebview's playVideo / stopVideo
+    # slots via this same proxy now that video lives in-process
+    # (issue #2904). Inject after load_browser so the proxy is
+    # already bound to the running AnthiasWebview; load_browser
+    # would otherwise race the D-Bus name registration.
+    _media_player_module.set_browser_bus(browser_bus)
 
 
 def start_loop() -> None:
--- a/src/anthias_viewer/media_player.py
+++ b/src/anthias_viewer/media_player.py
@@ -233,7 +233,7 @@ class MediaPlayer:
         raise NotImplementedError
 
 
-def _marshal_dbus_options(options: dict[str, Any]) -> dict:
+def _marshal_dbus_options(options: dict[str, Any]) -> dict[str, Any]:
     """Wrap each value as a ``GLib.Variant`` for pydbus.
 
     AnthiasWebview's ``playVideo`` slot is declared with
@@ -261,7 +261,7 @@ def _marshal_dbus_options(options: dict[str, Any]) -> dict:
     return {key: variant_for(value) for key, value in options.items()}
 
 
-def _build_mpv_options(uri: str) -> dict[str, str]:
+def _build_video_options(uri: str) -> dict[str, Any]:
     """Build the per-file option dict sent over D-Bus to AnthiasWebview.
 
     Qt 6.5 dropped the upstream gstreamer media backend, so the
@@ -285,10 +285,6 @@ def _build_mpv_options(uri: str) -> dict[str, str]:
     with the libmpv era (where it fed ffprobe). It's no longer
     read because libavcodec handles codec probing internally —
     but a future codec-specific tuning may re-introduce it.
-
-    Return type is annotated as ``dict[str, str]`` for historic
-    compatibility; values may be ``int`` (``video-rotate``) — the
-    marshalling layer accepts either.
     """
     del uri  # see docstring; kept for signature compatibility.
     device_type = os.environ.get('DEVICE_TYPE', '')
@@ -316,11 +312,12 @@ class MPVMediaPlayer(MediaPlayer):
     def __init__(self) -> None:
         MediaPlayer.__init__(self)
         self.uri: str = ''
-        # No mpv subprocess any more — the C++ AnthiasWebview owns
-        # the libmpv handle. Track local playback state so
-        # is_playing() (called only by tests today; the asset_loop
-        # sleeps for ``duration``) can still answer without a D-Bus
-        # round-trip.
+        # No mpv subprocess any more — playback runs inside
+        # AnthiasWebview via QtMultimedia (``QMediaPlayer`` +
+        # ``QGraphicsVideoItem``), reached over D-Bus. Track local
+        # playback state so ``is_playing()`` (called only by tests
+        # today; the asset_loop sleeps for ``duration``) can still
+        # answer without a D-Bus round-trip.
         self._playing: bool = False
 
     def set_asset(self, uri: str, duration: int | str) -> None:
@@ -332,7 +329,7 @@ class MPVMediaPlayer(MediaPlayer):
         # subprocess path and VLCMediaPlayer.
         settings.load()
 
-        options = _build_mpv_options(self.uri)
+        options = _build_video_options(self.uri)
 
         bus = get_browser_bus()
         if bus is None:
--- a/src/anthias_viewer/media_player.py
+++ b/src/anthias_viewer/media_player.py
@@ -1,14 +1,39 @@
 import logging
 import os
-import subprocess
-from typing import IO, ClassVar
+from typing import Any, ClassVar
 
-from anthias_common.board import ARM64_DEVICE_TYPES, resolve_device_key
+from anthias_common.board import ARM64_DEVICE_TYPES
 from anthias_common.device_helper import get_device_type
 from anthias_common.utils import clamp_screen_rotation
 from anthias_server.settings import settings
 
 VIDEO_TIMEOUT = 20  # secs
+
+
+# Lazy import for the pydbus proxy: the viewer service hands
+# MPVMediaPlayer the same ``browser_bus`` object it uses for
+# loadPage / loadImage (created in src/anthias_viewer/__init__.py
+# during load_browser()). Tests inject a mock; importing pydbus at
+# module load time would force every test to have pydbus available
+# even when only exercising VLCMediaPlayer.
+_browser_bus: Any = None
+
+
+def set_browser_bus(bus: Any) -> None:
+    """Inject the AnthiasWebview D-Bus proxy.
+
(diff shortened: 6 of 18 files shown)


vpetersson:
### 12 h stability burn-in — PR #2905 holds up

Ran a mixed-media stability test against x86 / Pi 4 / Pi 5 in parallel from 2026-05-16T20:46Z to 2026-05-17T08:38Z (~12 h per board). Asset rotation per board: 8 BBB video clips (1080p / 4K, 30 / 60 fps, H.264 + HEVC where the codec gate accepts both) + 3 webpages (anthias.screenly.io, example.com, screenly.io) + 3 JPEG images (1080p, 4K, portrait) — to cover all three asset mimetypes Anthias supports and exercise the View's QVideoWidget ↔ QLabel ↔ QWebEngineView surface-switching paths the loadImage('null') regression lived in.

Snapshots every 600 s into ``/tmp/burnin-snap.log`` on the board: viewer container restart count, RSS, ``END_FILE`` cumulative, ``STOP`` entries with ``elapsed_ms < 5000`` cumulative (the "first-play killed" signature). Final harvest pulls the per-board snapshot log + the in-container ``/data/.anthias/mpv-stats.log``.

| Board | Restarts | Early-stops | Clips played | Drops/clip 1st q → last q | RSS trajectory |
|---|---|---|---|---|---|
| x86 | 0 | 0 | 450 | 2 → 2 | 1540 MB flat (8 h) → step to 1900-2100 MB on webpage add → flat |
| Pi 4 | 0 | 0 | 344 | 3 → 2 | 944-1261 MB clip-dependent → step to 2000-2068 MB → flat |
| Pi 5 | 0 | 0 | 660 | 2 → 2 | 1595 MB ±5 MB across 7 h → step to 1920 MB → flat 4 h |

- **Zero container restarts** across 36 board-hours.
- **Zero first-play / early-stop failures.** The ``loadImage('null')`` fix from ``69bb873`` holds under sustained surface-switching — the rotation does video → web → image → video … indefinitely and the videos start cleanly every time.
- **No drop-rate degradation.** Comparing the first quartile of ``END_FILE`` rows to the last quartile, the median ``dropped`` count is identical on every board. Playback quality is steady-state, not drifting upward over time.
- **RSS is not leaking.** Every board's RSS jump correlates exactly with the moment I added the webpage + image assets mid-burn (snap 49 → 55, ~05:00 UTC). Post-shift the RSS is flat at the new working set. Pi 5 is the cleanest signal: 7 hours of dead-flat 1595 MB (±5 MB) on the video-only phase, then a single step-up to 1920 MB on webpage add, then 4 more hours at 1920 MB ±5 MB. That's QWebEngineView allocating its caches once, not a leak.

Rock Pi 4 was unreachable for the burn-in window (SSH banner exchange timing out under upload load — the testbed-side issue I noted in the previous comment). The arm64 image (``69bb873-arm64``) is identical to what just survived 12 hours on the other three boards.

vpetersson:
### 3 h asset-churn stress test — surface switching holds up

Follow-up to the 12 h burn-in. The burn-in proved steady-state stability with a static playlist; this test pounds on the *transitions* by mutating ``is_enabled`` on a random subset of the 14 assets (8 video + 3 webpage + 3 image) every 90 seconds for 120 iterations. 90 s is tighter than the typical asset play time, so the asset_loop is constantly reloading mid-play and the View has to swap between QMediaPlayer, QLabel, and QWebEngineView surfaces under churn.

Driver: ``/tmp/churn-loop.sh`` running on the dev host, 120 iterations × 90 s = 3 h per board. Heartbeat monitor on each board snapshots RSS / restart count / END_FILE / STOP-with-elapsed_ms<5000 every 600 s.

| Board | Churn ops | Restarts | **<100 ms stops** | 100-5000 ms stops | Clips finished | Drops/clip 1st q → last q | RSS range |
|---|---|---|---|---|---|---|---|
| x86 | 120 | 0 | **0** | 57 | 112 | 1 → 1 | 1931-2526 MB (no monotonic trend) |
| Pi 4 | 120 | 0 | **0** | 33 | 59 | 3 → 2 | 1999-2503 MB (no monotonic trend) |
| Pi 5 | 120 | 0 | **0** | 90 | 156 | 1 → 1 | 1892-2862 MB (no monotonic trend) |

### What the columns mean

- **<100 ms stops** is the regression signature for the ``loadImage('null')`` issue PR #2905 fixed at ``69bb873``: the foreground video being torn down within ~66 ms of LOADFILE because ``View::loadImage`` called ``hideVideoSurface()`` unconditionally for the ``null`` sentinel. **Zero on all three boards across 360 churn events / 1,080 asset toggles.** The fix holds under the most demanding playlist-mutation workload the test can apply.
- **100-5000 ms stops** is the legitimate asset_loop behaviour: when a churn event disables the currently-playing asset, the asset_loop honours the playlist reload and stops the clip. Stop ranges (100 ms → 5 s elapsed) match the wall-clock distance between LOADFILE and the next playlist refresh. Pi 5 has the most because HEVC HW decode cycles assets fast and the 90 s churn often catches a clip 1-5 s into playback.
- **Drops/clip** is the median ``dropped`` value on END_FILE rows for clips that played to completion. First-quartile vs last-quartile is identical (or 1 lower on Pi 4) — no degradation under churn.
- **RSS range** is the per-snapshot min/max from the heartbeat. The 500-1000 MB swing per board is the QWebEngineView / QMediaPlayer working set expanding and contracting as the enabled subset rotates; min/max repeat across the window without monotonic growth.

### Combined evidence stack for PR #2905

1. Multi-cycle BBB validation table (8 clips × 3 boards): Pi 4 1080p60 H.264 went from 2973 → 0 drops/min.
2. 12 h steady-state burn-in: 0 restarts, 0 early stops, drops flat, RSS no leak.
3. 3 h asset-churn (this comment): 0 restarts, 0 <100 ms stops, drops flat, RSS bounded.

Rock Pi 4 remains unreachable (SSH banner timeout, testbed-side, not code). The arm64 image (``69bb873-arm64``) is identical to what just survived both the 12 h and 3 h tests on the other boards.

vpetersson:
### Pi 4 rotation — real-device validation

Following up on the review feedback (the prior `QVideoWidget::setProperty("rotation", angle)` was a silent no-op because `QVideoWidget` has no `rotation` Q_PROPERTY). The shipping path is `QGraphicsView` + `QGraphicsScene` + `QGraphicsVideoItem`, with rotation applied via `QGraphicsItem::setRotation` — the value the painter actually consumes.

#### Round-trip from Settings to playback log

For each angle in {0, 90, 180, 270}, set `screen_rotation` in `~/.anthias/anthias.conf`, restart the viewer, play `BBB 1080p30 H264`, and grep `playback-stats.log`:

| `screen_rotation` | LOADFILE option dict | PLAYING rotation field | Pass |
|---|---|---|---|
| 0 | `{audio-device=…}` (no video-rotate, Python skips 0) | `rotation=0` | ✓ |
| 90 | `{audio-device=… video-rotate=90}` | `rotation=90` | ✓ |
| 180 | `{audio-device=… video-rotate=180}` | `rotation=180` | ✓ |
| 270 | `{audio-device=… video-rotate=270}` | `rotation=270` | ✓ |

#### Drop counts under rotation (Pi 4, BBB 1080p30 H.264, V3D 6.0)

Pulling the SAMPLE rows for ~20 s of playback at each angle:

| Rotation | Delivered / Expected | Drops | Drop rate |
|---|---|---|---|
| 0 | 573 / 573 | **0** | 0% |
| 90 | 598 / 605 | **7** | 1.2% |

Rotation 90 introduces a small GPU cost on V3D 6.0 (QGraphicsItem transform has a few-percent overhead vs. the unrotated path), but it's well under any acceptance bar and orders of magnitude below the 2973-drops/60 s libmpv subprocess baseline #2904 opened against.

#### Test coverage

`src/anthias_webview/tests/test_videoview.cpp::playRotatesVideoItem` asserts on `QGraphicsItem::rotation()` (the painter's input) for each cardinal angle, with both `str` and `int` D-Bus marshalling. `playSnapsBadAngleToZero` covers the defensive `45° → 0°` normalisation. 8 / 8 QtTest cases pass under `QT_QPA_PLATFORM=offscreen`.

The other review items landed in the same commit (rebased PR head):
- `playback-stats.log` (renamed from `mpv-stats.log`) capped at 8 MB on viewer start
- `playStartedAt` now starts on `MediaStatus::LoadedMedia`, removing the decoder-init inflation in clip-start drop counts
- `resolveAlsaDevice` extracts `CARD=<name>` and logs the resolved `QAudioDevice` id at INFO so multi-HDMI Pi 4 / Pi 5 mismatches surface in journalctl
- `_marshal_dbus_options` picks the GLib.Variant signature by Python type so int / bool / float options work without bespoke call-site wrapping
- Stale `gstreamer` / `libmpv` comments swept from the codebase (the actual pipeline is `QtMultimedia` → ffmpeg-backed libavcodec via the +rpt1 `libav*` pin)

PR #2915, opened 2026-05-19
feat(viewer,server): 1 GB SBC enablement — low-RAM degradation gates

vpetersson:
## Summary

Boards with less than 1.5 GiB MemTotal (Pi 2/Pi 3 1GB, Pi 4 1GB, Rock Pi 4 1GB,
generic-arm64 1GB SKUs) OOM-cycled when QtMultimedia loaded a 4K HEVC asset
alongside the two QtWebEngine renderers introduced by #2905. On-device repro on
a 1 GB Rock Pi 4 captured `global_oom` on the docker container's bash process
followed by a restart loop, plus the kernel keeping sshd in banner-exchange
until power-cycle. This patch puts the device into a graceful degraded mode
before the OOM cascade fires.

- Viewer collapses to a single `QWebEngineView` on low-RAM (no preloaded
  crossfade; page swap in-place with a brief blank during load). Saves
  ~100 MB physical RAM per device by dropping the second Chromium renderer.
- Codec gate gains a 1080p resolution cap on low-RAM boards. Above-1080p
  uploads are rejected at upload with the existing recipe machinery, extended
  to inject `-vf scale=1920:1080:force_original_aspect_ratio=decrease` so the
  operator's re-encode also lands within the envelope.
- `anthias_host_agent` publishes `host:total_mem_kb` to Redis alongside the
  existing `host:board_subtype`; `anthias_common.board` gains `LOW_RAM_THRESHOLD_KB`
  / `get_total_mem_kb()` / `is_low_ram_device()` so server-side code reads
  MemTotal without re-opening `/proc/meminfo`.
- Diagnostics page (Memory card + `/api/v2/info`) surfaces a "Low-RAM mode"
  badge with the threshold MiB so operators can see *why* the device degraded.
- `docs/board-enablement.md` — rewrote the stale `--hwdec=drm-copy` /
  per-codec dispatch text removed in #2905; documented the known mainline
  rkvdec limitation on Armbian 6.18 (HEVC stateless engages via `-hwaccel drm`
  but produces decode errors; H.264 has no v4l2_request binding in `+rpt1`
  7.1.3) and the new low-RAM mode.

## On-device measurements (Rock Pi 4 1 GB, RK3399)

|                       | Before (`db999f1`) | After (`22abfb2`, low-RAM gate active) |
|---|---|---|
| Idle viewer + Chromium RSS | ~440 MB                  | **~244 MB**                                  |
| QtWebEngine renderers      | 2                        | **1**                                        |
| 4K HEVC asset              | OOM-loops the container  | Rejected at upload with downscale recipe     |
| CMA free at idle           | ~1.5 MB                  | ~12 MB                                       |

## Out of scope (separate work)

HEVC HW-decode on arm64 is gated by an upstream rkvdec/Hantro driver fix that
isn't in any Debian-shipped Armbian kernel today. Anthias is userspace-only and
doesn't carry its own kernel patches, so this PR ships graceful degradation
rather than chasing the driver.

## Test plan

- [x] Unit tests: 130 in `tests/test_processing.py` + `tests/test_host_agent_board_subtype.py` pass, including new coverage for the four matrix cells (low/high-RAM × in/over-cap), recipe shape with/without `cap_to_1080p`, the cap defence against unknown / zero dimensions, host_agent's MemTotal parser, and the API endpoint's new `low_ram` field. Full suite: **894 passed** non-integration.
- [x] `ruff check` + `ruff format --check` clean across changed files.
- [x] **On-device validation — all four testbeds**:
  - x86 (8 GB): `ANTHIAS_LOW_RAM=0`, 2 renderers, `low_ram=False`.
  - Pi 4 (4 GB): `ANTHIAS_LOW_RAM=0`, 2 renderers, `low_ram=False`.
  - Pi 5 (8 GB): `ANTHIAS_LOW_RAM=0`, 2 renderers, `low_ram=False`.
  - Rock Pi 4 (1 GB): `ANTHIAS_LOW_RAM=1`, **1 renderer**, `low_ram=True`, 4K HEVC upload **rejected** with the downscale recipe (`ffmpeg -i upload.mp4 -vf scale=1920:1080:force_original_aspect_ratio=decrease -c:v libx264 …`).
- [ ] Manual: confirm the diagnostics page badge renders correctly (browser, not headless).

### DIFF
--- a/bin/upgrade_containers.sh
+++ b/bin/upgrade_containers.sh
@@ -21,6 +21,20 @@ export SHM_SIZE_KB="$(echo "$TOTAL_MEMORY_KB" \* 0.3 | bc | cut -d'.' -f1)"
 # against a decompression-bomb fixture or runaway ffprobe, not
 # because routine workloads come anywhere near it.
 export CELERY_MEMORY_LIMIT_KB=$(echo "$TOTAL_MEMORY_KB * 0.6" | bc | cut -d'.' -f1)
+# Low-RAM gate. Boards with < 1.5 GiB MemTotal (Pi 2/Pi 3 1GB, Pi 4 1GB,
+# Rock Pi 4 1GB, generic-arm64 SBCs in that class) can't keep two
+# QtWebEngine renderer processes resident *and* play 1080p+ video
+# without OOM-thrashing through swap. The viewer reads this env to
+# drop into single-WebEngineView mode (no preloaded crossfade); the
+# asset processor reads it via Redis (host:total_mem_kb) and rejects
+# uploads above 1080p. Threshold is 1.5 GiB so 1 GB SKUs land below
+# and 2 GB SKUs sit above; both 1024 MB and 2048 MB boards exist in
+# the supported fleet.
+if [ "${TOTAL_MEMORY_KB:-0}" -lt 1572864 ]; then
+    export ANTHIAS_LOW_RAM=1
+else
+    export ANTHIAS_LOW_RAM=0
+fi
 GIT_BRANCH="${GIT_BRANCH:-master}"
 
 MODE="${MODE:-pull}"
--- a/docker-compose.yml.tmpl
+++ b/docker-compose.yml.tmpl
@@ -59,6 +59,14 @@ services:
       - LANG=${LANG}
       - LANGUAGE=${LANGUAGE}
       - LC_ALL=${LC_ALL}
+      # ``1`` on boards with < 1.5 GiB MemTotal. AnthiasViewer reads
+      # this to fall into single-QWebEngineView mode (no preloaded
+      # crossfade — the inactive renderer cost ~100 MB physical RAM
+      # per box, untenable on 1 GB SBCs). Derived from /proc/meminfo
+      # in bin/upgrade_containers.sh so the server side
+      # (host_agent → ``host:total_mem_kb`` in Redis) and the viewer
+      # side both observe the same total.
+      - ANTHIAS_LOW_RAM=${ANTHIAS_LOW_RAM}
     privileged: true
     restart: always
     # cage (the wlroots kiosk compositor used on Pi 5 / x86 /
--- a/docs/board-enablement.md
+++ b/docs/board-enablement.md
@@ -80,20 +80,83 @@ upstream of the upload.
 `bin/install.sh` sets `DEVICE_TYPE=arm64` for every aarch64 SBC it doesn't
 recognise as a Pi. `anthias_host_agent` runs on the host and reads
 `/proc/device-tree/model`; when it sees "Radxa ROCK Pi 4" it writes
-`host:board_subtype = 'rockpi4'` to Redis. The viewer reads that key to
-upgrade its `--hwdec=` choice from the catch-all `arm64` default to the
-RK3399-specific `--hwdec=drm-copy` (v4l2_request, served by `rkvdec` for HEVC
-and the Hantro VPU for H.264).
+`host:board_subtype = 'rockpi4'` to Redis. The server reads that key to
+pick the right entry in `processing._HW_DECODE_VIDEO_CODECS` — Rock Pi 4
+accepts H.264 + HEVC uploads, the catch-all `arm64` accepts nothing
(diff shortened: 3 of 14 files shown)


vpetersson:
## On-device leak test (Rock Pi 4 1 GB): partial result

Kicked off the 4-hour mixed-rotation test Viktor asked for (5 assets: image, 1080p H.264, 1080p HEVC, `example.com`, `engadget.com`; 30s/asset).

**~30 min in, the viewer container OOM-restarted.** Captured in `dmesg`:

> `oom-kill: ... task=bash, pid=14893, uid=0, oom_score_adj:1000`
> `Out of memory: Killed process 14893 (bash) total-vm:4020kB`

The victim is `start_viewer.sh`'s bash (intentionally given `oom_score_adj=1000` so it dies first), container's `restart: always` brought it back, single-view mode re-engaged. **The mitigations in this PR helped** — first 30 min ran cleanly where pre-patch the box OOMed within seconds on a 4K load — but sustained mixed-rotation pressure on a 1 GB box still exceeds the budget.

Detailed findings + suggested follow-up mitigations filed as #2916.

**Recommendation:** ship this PR as-is. The Phase A value is clear:
- 4K HEVC upload immediately fails with a recipe (was: OOM loop)
- Idle viewer + Chromium drops from ~440 MB → ~244 MB RSS
- First 30 min of mixed playback works (was: minutes-to-OOM with 4K in rotation)

The residual sustained-rotation OOM is a tractable next problem — heavyweight-URL gate, Chromium memory caps, libavcodec thread limit, or removing swap. None of those belong in this PR's scope.

Continuing the 4-hour run to characterize the steady-state restart cadence; will post a follow-up comment with the full data when done.

vpetersson:
## Leak test wrap-up

The 4-hour test ended at ~50 min — second OOM cascade was severe enough to wedge sshd entirely (kernel/ping responding, userspace stuck on `fork`), required a physical power-cycle. CSV was on `/tmp` (tmpfs on Armbian) so the full sample stream didn't survive the reboot, but I have the first 24 samples and the live monitor data:

### Steady-state restart cadence on the 5-asset mixed rotation

| Event | Wall-clock from start | Recovery |
|---|---|---|
| 1st OOM | ~22 min | Container `restart: always` re-spawned the viewer; box stayed reachable |
| 2nd cascade | ~51 min (= 29 min after the 1st recovery) | sshd wedge — required physical power-cycle |

So on a 1 GB Rock Pi 4 with `image + 1080p H.264 + 1080p HEVC + example.com + engadget.com` cycling at 30s/asset, the box trends to **one OOM event roughly every 20–30 min**, and the second event tends to be worse than the first (kernel reclaim doesn't quite recover the slack the container restart freed).

### viewer_rss trend over the first 22 min (24 samples)

```
172, 206, 199, 220, 235, 244, 228, 233, 257, 231, 222, 223, 298, 263 ...
```

Bouncy but **no clear monotonic climb** — the OOM is from the steady-state working set being too close to the budget, not from a leak. Combine that with engadget.com renderer at ~120 MB + HEVC SW decode working-set + AnthiasViewer main process at ~250 MB and you blow past the 966 MB physical envelope on the worst cycle phase.

### Takeaways for the follow-up (#2916)

This test reinforces the prioritization:

1. **Swap-off on low-RAM** is now my strong lean — the slow swap thrash (sampler `sleep 30` taking 16 min) is what made the second OOM wedge sshd rather than restart cleanly.
2. **Heavyweight-URL gate** would directly address the engadget.com contribution to the working set.
3. **Chromium memory caps** are a softer lever but cheaper to ship.

### Recovery / future-test hygiene

- The sampler should write to `/data/.anthias/leak_test.csv` (persistent, already bind-mounted into the viewer container's `/data/.anthias/playback-stats.log` path), not `/tmp/`.
- The Monitor's PID-change check should treat `viewer_pid=0` (process gone, container respawning) as an event, not skip it.

The PR's Phase A value still stands — 4K immediately fails with a recipe, idle viewer RSS drops by ~200 MB, first 22 min of churn ran cleanly where pre-patch the box wedged within seconds on a 4K rotation. The residual sustained-rotation OOM is real but not a regression; it's a known limit that #2916 will chase.

PR #2923, opened 2026-05-19
fix(ci): annotate per-board snippets with Pi Imager devices tags

vpetersson:
### Issues Fixed

Pi Imager filters the OS list against the user-selected device. The Pi 5 entry in the upstream root JSON uses `matching_type: exclusive` — an image lacking a `devices` array is silently hidden when a user picks Pi 5 in the device picker. Raspberry Pi has signalled (Tom Dewey email, Nov 2023) that the same exclusive filter will roll out to older boards.

Our per-board JSON snippets currently emit **no** `devices` field, so even after we switch the upstream Anthias entry to `subitems_url: https://anthias.screenly.io/rpi-imager.json`, our Pi 5 build wouldn't surface to users.

### Description

- Map BOARD slug → device tags in the workflow's package step:
  - `pi2` → `["pi2-32bit"]`
  - `pi3` → `["pi3-64bit"]`
  - `pi4-64` → `["pi4-64bit"]`
  - `pi5` → `["pi5-64bit"]`
  - `x86` → `[]` (not a Pi Imager target — empty array keeps the snippet schema-valid)
- Add `devices` to `REQUIRED_FIELDS` in `build_pi_imager_json.py` so the contract is documented and the existing required-fields test enforces it.
- Extend the test fixture's `make_image_metadata` to include `devices`, and add a per-board parametrized test that the field round-trips through `retrieve_and_patch_json` unmodified.

Tags taken from the canonical `imager.devices[*].tags` in `https://downloads.raspberrypi.org/os_list_imagingutility_v4.json`.

### Checklist

### DIFF
--- a/.github/workflows/build-balena-disk-image.yaml
+++ b/.github/workflows/build-balena-disk-image.yaml
@@ -404,7 +404,7 @@ jobs:
           # array so the snippet still validates as schema-conformant.
           case "$BOARD" in
             pi2)    DEVICES='["pi2-32bit"]' ;;
-            pi3)    DEVICES='["pi3-64bit"]' ;;
+            pi3)    DEVICES='["pi3-32bit"]' ;;
             pi4-64) DEVICES='["pi4-64bit"]' ;;
             pi5)    DEVICES='["pi5-64bit"]' ;;
             x86)    DEVICES='[]' ;;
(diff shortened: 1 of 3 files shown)


vpetersson:
Superseded by https://github.com/Screenly/Anthias/pull/3090

PR #2927, opened 2026-05-20
fix(viewer): default x86 audio fallback to ALSA 'default'

Xzzz:
On x86, `get_alsa_audio_device()` returns `sysdefault:CARD=HID` as the fallback, which matches no real ALSA card on Intel/AMD/Nvidia HDA chipsets. This bypasses the container's `~/.asoundrc` bind mount and leaves operators with no way to route audio.

The ARM64 branch a few lines above already returns `'default'` for the same reason — heterogeneous SoCs, no portable card name — and lets operators override via `~/.asoundrc`. This PR mirrors that approach for x86.

## Reproduction
- x86 host (Intel HDA `PCH` card with HDMI output to a TV)
- Debian minimal install, Anthias v2026.05.0+
- mpv launched by the viewer logs `--audio-device=alsa/sysdefault:CARD=HID` → no audio
- `/proc/asound/cards` shows only card 0 (`HDA Intel PCH`); no `HID` card exists

## Verification
After this patch, `~/.asoundrc` (e.g. routing `default` to `hw:0,3` for the HDMI sink) is honored, and audio plays correctly on the TV.

### DIFF
--- a/src/anthias_viewer/media_player.py
+++ b/src/anthias_viewer/media_player.py
@@ -213,7 +213,13 @@ def get_alsa_audio_device() -> str:
         elif device_type in ['pi1', 'pi2', 'pi3']:
             return 'sysdefault:CARD=vc4hdmi'
         else:
-            return 'sysdefault:CARD=HID'
+            # x86 fallback: ALSA card names vary across Intel/AMD/Nvidia
+            # HDA chipsets, so there is no portable per-SoC name we can
+            # hard-code. Defer to ALSA's `default` device and let
+            # operators override via ~/.asoundrc (already bind-mounted
+            # into the viewer container — see docker-compose.yml.tmpl),
+            # mirroring the ARM64 path above.
+            return 'default'
 
 
 class MediaPlayer:


vpetersson:
Validated against an x86 testbed (Intel HDA PCH + a USB Audio+HID dongle) — change is structurally correct, but the PR ships a failing unit test.

## Validation
- `/proc/asound/cards` on the testbed shows card 0 = `HDA Intel PCH`, card 1 = `HID` (CSCTEK USB Audio+HID dongle). Minor nit on the PR body: a `HID` card *does* exist on this box — it's just a USB peripheral, not the HDMI sink. Softening the wording to "no portable card name across Intel/AMD/Nvidia HDA chipsets" (matching the ARM64 comment this mirrors) would be more accurate; the symptom you describe (audio routed away from HDMI) still applies.
- C++ side handles `default` cleanly: `src/anthias_webview/src/videoview.cpp:380` short-circuits when the extracted card name is literally `default` and uses `QMediaDevices::defaultAudioOutput()`. No D-Bus / `CARD=` parsing regression from dropping the `CARD=` token.
- `~/.asoundrc` bind mount is in place (`docker-compose.yml.tmpl:82`), so the override hook the PR relies on really exists.

## Blocker — CI will fail
`tests/test_media_player.py:341` still asserts the old value:

```python
def test_hdmi_on_x86_falls_back_to_hid(alsa_settings: Any) -> None:
    ...
    assert get_alsa_audio_device() == 'sysdefault:CARD=HID'
```

Running it against this branch:

```
AssertionError: assert 'default' == 'sysdefault:CARD=HID'
```

Please rename to `test_hdmi_on_x86_falls_back_to_default` and assert `'default'`. (The status rollup only shows SonarCloud right now — the pytest job will catch this once it runs.)

## Nice-to-have (non-blocking)
The ARM64 branch logs the chosen ALSA card once per process via `_log_arm64_alsa_default_once()` so silent-HDMI reports carry breadcrumbs. The x86 path now has the same "debug via `.asoundrc`" failure mode — symmetric once-per-process logging here would help future operators. Fine to defer to a follow-up.

## Risk
Low. For devices without an `.asoundrc`, behavior changes from "fall back to a `CARD=HID` device that probably isn't the intended sink" to "fall back to Qt's platform default" — strictly better, and overridable as documented.

vpetersson:
Thank you for the contribution @Xzzz - I've merged this into https://github.com/Screenly/Anthias/pull/3038 and added some more functionality.

Xzzz:
You welcome @vpetersson, Glad to hear that this is helpful to everyone :)
Thanks for your work!

PR #3014, opened 2026-06-07
fix(server): stop the GitHub update check from firing Sentry events

vpetersson:
### Issues Fixed

Sentry event `139b36ce6ee04699cbae1c1c30be2619` (release 2026.6.2): `ReadTimeout fetching latest release from GitHub: no data` on `/splash-page/`.

### Description

Every page render calls `is_up_to_date()` via the shared page context, which fetches GitHub's `/releases/latest` with a 5s timeout. A network blip is already handled gracefully (5-minute backoff key + cached-verdict fallback), but the failure was logged at `error` level — and Sentry's logging integration turns error-level logs into events, so every flaky-network device fired a Sentry event for a non-bug.

- Downgrade the update-check failure logs (transport errors, malformed/missing/unparseable `tag_name`) to `warning` — these are environmental conditions on device networks (timeouts, rate-limited NATs, captive portals), not application errors, and the check is best-effort by design.
- Log `str(exc)` instead of the useless `no data` when the failure has no HTTP response, so the log line carries the host/timeout detail.
- Add a regression test pinning the warning level and the exception detail.

### Checklist

### DIFF
--- a/src/anthias_server/lib/github.py
+++ b/src/anthias_server/lib/github.py
@@ -50,9 +50,14 @@ def handle_github_error(
     if exc.response is not None:
         errdesc = exc.response.content
     else:
-        errdesc = 'no data'
+        errdesc = str(exc) or 'no data'
 
-    logging.error(
+    # Warning, not error: the update check is best-effort (5-minute
+    # backoff + cached-verdict fallback), and timeouts / rate limits
+    # are routine on device networks. Sentry's logging integration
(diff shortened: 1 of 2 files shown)


vpetersson:
Superseded: the warning-level downgrade shipped in #3019, and the one improvement #3019 didn't carry (`errdesc = str(exc) or 'no data'`) is ported with its test in #3023.

PR #3018, opened 2026-06-07
fix(sentry): stop reporting transient redis blips and client disconnects

vpetersson:
### Issues Fixed

Sentry: [ANTHIAS-M](https://anthias.sentry.io/issues/7534699071/) / [ANTHIAS-H](https://anthias.sentry.io/issues/7534605276/) (viewer resolution reporter, redis refused / DNS not yet resolvable), [ANTHIAS-K](https://anthias.sentry.io/issues/7534698885/) (celery consumer reconnect log), [ANTHIAS-J](https://anthias.sentry.io/issues/7534613775/) (ASGI redis connection reset), [ANTHIAS-N](https://anthias.sentry.io/issues/7534704829/) (`CancelledError` on client disconnect), [ANTHIAS-P](https://anthias.sentry.io/issues/7534806426/) (celery beat reconnect log).

### Description

**Two layers, per review feedback:**

1. **Orchestration (root cause for startup ordering):** redis now has a `redis-cli ping` healthcheck, and `anthias-server` / `anthias-viewer` / `anthias-celery` are gated on `condition: service_healthy` in the prod template, dev, and test composes. Bare `service_started` only orders container *creation*, so consumers could race a redis still loading its RDB.
2. **Sentry-side handling for what compose can't gate:** the balena supervisor doesn't support `depends_on` conditions (fleet devices get no ordering guarantee), and nothing gates a redis container recycling *mid-life* (OOM, upgrade) — long-running services will always see a window of `ConnectionError` then. Every consumer already self-heals, so those windows are filtered rather than reported.

Redis being briefly unreachable (its container recycling, or compose startup before the `redis` DNS name resolves) is an expected state on a signage device, and every consumer already self-heals — celery's consumer reconnects with backoff, the viewer's reporter retries on its next tick, Channels re-establishes on the next WebSocket frame. Same for an HTTP client hanging up mid-request under ASGI: Django/uvicorn cancel the handler by design. Neither is a code bug, but together they produced 6 noise events per redis blip / disconnect.

- `before_send` hook drops events whose exception chain (walking `__cause__`/`__context__`, since channels-redis/kombu wrap the underlying error) contains `redis.exceptions.ConnectionError` or `asyncio.CancelledError`
- `ignore_logger('celery.worker.consumer.consumer')` — the reconnect retry arrives as an ERROR log line, not an exception
- The viewer's resolution reporter now logs a redis blip at WARNING (retry next tick) instead of ERROR-with-traceback; tick body extracted into a testable helper

A *persistent* redis outage still surfaces — the device visibly stops working and the restart loop shows in balena — but a 5-second blip no longer fans out into Sentry issues.

### Checklist

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -18,7 +18,8 @@ services:
       - CELERY_RESULT_BACKEND=redis://redis:6379/0
       - ENVIRONMENT=development
     depends_on:
-      - redis
+      redis:
+        condition: service_healthy
     restart: always
     volumes:
       - anthias-data:/data
@@ -33,7 +34,7 @@ services:
       anthias-server:
         condition: service_started
       redis:
-        condition: service_started
+        condition: service_healthy
     command: >
       nice -n 19 ionice -c 3
       celery -A anthias_server.celery_tasks.celery worker -B -n worker@anthias
@@ -51,6 +52,12 @@ services:
   redis:
     platform: "linux/amd64"
     image: mirror.gcr.io/library/redis:alpine
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      timeout: 3s
+      retries: 12
+      start_period: 10s
 
 volumes:
     anthias-data:
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -52,7 +52,7 @@ services:
       anthias-test:
         condition: service_started
       redis:
-        condition: service_started
+        condition: service_healthy
     environment: *test_env
     restart: always
     volumes:
@@ -61,6 +61,12 @@ services:
 
   redis:
     image: mirror.gcr.io/library/redis:alpine
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      timeout: 3s
+      retries: 12
+      start_period: 10s
 
 volumes:
   anthias-data:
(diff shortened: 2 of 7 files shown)


vpetersson:
Good call — added the orchestration layer: redis now has a `redis-cli ping` healthcheck and server/viewer/celery are gated on `condition: service_healthy` (prod template + dev + test composes).

Kept the Sentry-side handling alongside it, for three reasons:
1. **balena** — the supervisor doesn't support `depends_on` conditions, so fleet devices get no startup ordering from compose; their first-boot reconnects would keep generating events.
2. **mid-life restarts** — `depends_on` only gates first start. A redis container recycling later (OOM, upgrade roll) still throws `ConnectionError` into the long-running services, and each of them already self-heals.
3. `CancelledError` (ANTHIAS-N) is client-disconnect noise unrelated to redis ordering.

vpetersson:
Balena compatibility check on the compose changes:

- The balena deploy paths (`bin/balena_ota_deploy.sh`, `bin/deploy_to_balena.sh`) render `docker-compose.balena.yml.tmpl` / `docker-compose.balena.dev.yml.tmpl` — **untouched by this PR**, so fleet devices never see the new syntax.
- The `condition: service_healthy` gating is only in `docker-compose.yml.tmpl` (plain compose installs via `upgrade_containers.sh`), `docker-compose.dev.yml`, and `docker-compose.test.yml`.
- Confirmed against the [balena supervisor compose reference](https://docs.balena.io/reference/supervisor/docker-compose/): `depends_on` supports *"only array form and `service_started` condition"* — which is exactly why the balena templates stay on the supervisor's defaults and the Sentry-side filter carries the fleet.

PR #3071, opened 2026-06-11
fix(server): classify RTSP/streaming URLs as streaming, not webpage

vpetersson:
### Issues Fixed

Not tied to a tracked issue. Adding an RTSP (or HLS/DASH) URL through the **Add asset → From URL** tab created the asset as `mimetype='webpage'`, so the viewer handed it to QtWebEngine instead of the video player and the stream never played — even though the modal advertises "streaming URLs are all supported." Regression from the React→Django frontend migration (#2818), whose new `assets_create` handler dropped the scheme check that the old React `getMimetype` had.

### Description

- Add `is_streaming_uri()` to `anthias_common.remote_video`, reusing the existing stream-scheme and manifest-extension (`.m3u8/.mpd/.m3u/.ism`, http(s) only) sets so it can't drift from `is_downloadable_remote_video`.
- `assets_create` now classifies stream URLs as `mimetype='streaming'` (after YouTube, before the extension heuristic) and applies the dedicated `default_streaming_duration` window — previously dead config.
- **Reject `rtmp://`** at `validate_url` (gates the UI form and the API) with a clear "use RTSP/HLS/DASH" message, and drop it from `_STREAM_SCHEMES`. On-device testing showed RTMP renders black on every board: Qt6's QMediaPlayer sets a `timeout` option the rtmp protocol misreads as TCP *listen* mode, so the open fails (`libavformat` itself plays rtmp fine). Rather than let operators add a stream that never shows, we don't accept it.
- Regression tests for the classifier, the rtmp rejection, and the form handler.

### Testing

Unit: classifier + form-handler + rtmp-rejection tests; full suite green, ruff clean.

On-device, full stack, across the entire testbed. An ffmpeg-fed feed (Big Buck Bunny) was streamed from a LAN host via mediamtx (RTSP) and nginx (HLS/DASH); playback was verified from `playback-stats.log` (`position-ms` advancing + `frames-rendered` climbing), HW-decoder fds, and screenshots.

| Board (decode) | RTSP H.264 | RTSP H.265 | HLS | DASH | RTMP |
|---|---|---|---|---|---|
| x86 (SW) | ✅ visual+reader | ❌ QtMultimedia FormatError¹ | ✅ played to EOF | ✅ progressive+visual | ❌ rejected |
| Pi 5 | ✅ frames advancing | ✅ **HW** (`/dev/video19`, Hantro G2) | — | — | ❌ rejected |
| Pi 4 | ✅ **HW** (`/dev/video10`) | — | — | — | ❌ rejected |
| Rock Pi 4, 1 GB (SW, 360p) | ✅ no OOM | — | — | — | ❌ rejected |

RTSP — the regression target — plays on every board, with hardware decode on the Pi 4 (H.264) and Pi 5 (H.265). HLS/DASH play on x86 (HLS needs a Range-capable HTTP origin — any real server/CDN).

¹ **x86 H.265 over RTSP** fails inside QtMultimedia (`FormatError`, 0 frames) even though x86's own `libavcodec` decodes the same stream fine over both UDP and TCP, and H.264-over-RTSP works on x86. It's an x86 QtMultimedia HEVC limitation, unrelated to this change.

### Known follow-up (pre-existing, out of scope)

- **`srt/udp/mms`** are recognised by `is_streaming_uri` but rejected by `validate_url`, so they're unreachable end-to-end (untouched here).

### Checklist

- [x] I have done an end-to-end test for Raspberry Pi devices. (RTSP H.264 on Pi 5/Pi 4/Rock Pi 4; H.265 HW-decoded on Pi 5)
- [x] I have tested my changes for x86 devices. (RTSP, HLS, DASH)

### DIFF
--- a/src/anthias_common/remote_video.py
+++ b/src/anthias_common/remote_video.py
@@ -230,9 +230,16 @@ def is_streaming_uri(uri: str) -> bool:
     """
     if not uri:
         return False
-    if _url_scheme(uri) in _STREAM_SCHEMES:
+    scheme = _url_scheme(uri)
+    if scheme in _STREAM_SCHEMES:
         return True
-    return _url_path_ext(uri) in _STREAM_MANIFEST_EXTS
+    # Manifests are HTTP-delivered (HLS/DASH); only treat a manifest
+    # extension as streaming over http(s). A ``file:///…/index.m3u8``
+    # (or any other scheme) is not a live stream and must not be
+    # classified as one.
+    if scheme in ('http', 'https'):
+        return _url_path_ext(uri) in _STREAM_MANIFEST_EXTS
+    return False
 
 
 def remote_video_destination_path(
--- a/src/anthias_common/remote_video.py
+++ b/src/anthias_common/remote_video.py
@@ -203,6 +203,31 @@ def is_downloadable_remote_video(uri: str) -> tuple[bool, str]:
     return _head_probe(uri)
 
 
+def is_streaming_uri(uri: str) -> bool:
+    """True when *uri* is a live stream the viewer plays directly via
+    its video pipeline rather than a downloadable file or a renderable
+    web page.
+
+    Covers both streaming-by-construction schemes (``rtsp://``,
+    ``rtmp://``, ``srt://``, ``udp://``, ``mms://``) and HTTP-delivered
+    streaming manifests (HLS ``.m3u8``, DASH ``.mpd``, legacy ``.m3u``,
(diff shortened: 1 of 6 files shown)


vpetersson:
The x86 H.265/HEVC playback failure noted in the testing section is now tracked separately in #3072 (Qt6 VAAPI-under-cage HEVC decode — not caused by this PR; HEVC hardware-decodes fine on Pi 5).

PR #3093, opened 2026-06-24
fix: bias yt-dlp format_sort toward board's HW-decoded codec (Pi 5 HEVC)

### LINKED ISSUE #3092: [BUG] Pi 5: YouTube URL download picks H.264, fails on device that only supports HEVC hardware decode
Description
When adding a YouTube URL as an asset, Anthias downloads the video successfully but then fails with:

Processing failed

Video codec 'h264' is not hardware-decoded on this device. Supported: hevc.

Anthias correctly detects that the Pi 5 only supports HEVC hardware decode, but the yt-dlp download step does not use that information to request a compatible stream upfront. The capability detection code already exists (it produces the error message) but is not fed back into the yt-dlp format selection.
Steps to Reproduce

Open the Anthias dashboard on a Raspberry Pi 5
Click Add Asset and paste a standard YouTube URL
Wait for the download to complete
Asset shows Processing failed — Video codec 'h264' is not hardware-decoded on this device. Supported: hevc.

Expected Behavior
Anthias should detect the Pi 5's supported codecs before downloading and pass an appropriate format selector to yt-dlp, e.g.:
-f "bestvideo[vcodec=hevc]/bestvideo[vcodec^=av01]+bestaudio/best"
The asset should download, process, and play without any manual intervention.
Screenshots
N/A
Environment

Device: Raspberry Pi 5
OS: Raspberry Pi OS 64-bit Bookworm
Anthias Version: v2026.6.3
Installation Method: Manual Installation

TekScene:
## Summary

Fixes #3092.

On Pi 5, `download_youtube_asset` used a hardcoded `vcodec:h264` format_sort preference. yt-dlp downloaded H.264 (the only codec YouTube typically serves), which then failed at the `normalize_video_asset` codec gate:

> Video codec 'h264' is not hardware-decoded on this device. Supported: hevc.

Two root causes fixed:

**1. yt-dlp format selection (`celery_tasks.py`):** Call `_hw_decoded_codecs()` before building `ydl_opts` and prefer HEVC in `format_sort` whenever the board supports it. If YouTube ever serves an HEVC stream, it'll be picked automatically. When it doesn't (the common case), yt-dlp falls back to H.264.

**2. Pi 5 codec gate too strict (`processing.py`):** Pi 5's BCM2712 has no H.264 hardware decoder, but its Cortex-A76 cores software-decode 1080p H.264 without frame drops. YouTube rarely serves HEVC, so the previous HEVC-only gate rejected every YouTube download. H.264 is now accepted on Pi 5 alongside HEVC. On-device transcoding was explicitly ruled out — a prior attempt wedged a Pi 4's Celery worker for 99 minutes on a single H.264→HEVC pass.

**Changes:**
- `src/anthias_server/celery_tasks.py`: prefer HEVC in format_sort on any board that has it; fall back to H.264
- `src/anthias_server/processing.py`: add `h264` to Pi 5's accepted codec set
- `tests/test_celery_tasks.py`: 4 new tests for per-board format_sort selection; Pi 4 test updated (now prefers HEVC when available)
- `tests/test_processing.py`: replace Pi 5+H.264 rejection tests (no longer valid) with Pi 3+HEVC rejection tests using mocked ffprobe; add new test asserting H.264 is accepted on Pi 5

## Test plan

- [x] Verified on physical Pi 5: YouTube URL downloads and plays without "Processing failed"
- [ ] `pytest tests/test_celery_tasks.py -k "format"` passes (4 new tests)
- [ ] `pytest tests/test_celery_tasks.py` fully green
- [ ] `pytest tests/test_processing.py` fully green

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -779,6 +779,47 @@ def test_video_h264_accepted_on_pi5_software_decode(
     mock_notify.assert_called_once_with('vid-h264-pi5-sw')
 
 
+@pytest.mark.django_db
+def test_video_unsupported_codec_still_rejected_on_pi5(
+    asset_dir: str, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """VP9 / AV1 uploads on Pi 5 are still rejected even though H.264
+    software-decode was added to the Pi 5 codec set. The rejection recipe
+    tells the operator to re-encode to H.264 (accepted via software decode —
+    faster to encode than HEVC and still plays fine on Cortex-A76)."""
+    monkeypatch.setenv('DEVICE_TYPE', 'pi5')
+    src = path.join(asset_dir, 'sample.mp4')
+    with open(src, 'wb') as f:
+        f.write(b'\x00' * 16)
+    asset = _make_processing_asset('vid-vp9-pi5', src, mimetype='video')
+    fake_summary = {
+        'container': 'mp4',
+        'video_codec': 'vp9',
+        'video_width': 1920,
+        'video_height': 1080,
+        'video_fps': 30.0,
+        'audio_codec': 'aac',
+        'duration_seconds': 60,
+    }
+
+    with (
+        mock.patch.object(processing, '_notify'),
+        mock.patch.object(
+            processing, '_ffprobe_summary', return_value=fake_summary
+        ),
+    ):
+        with pytest.raises(processing.UnsupportedVideoCodecError) as excinfo:
+            processing._run_video_normalisation(asset)
+
+    import shlex as _shlex
+
+    recipe = excinfo.value.recipe
+    tokens = _shlex.split(recipe)
+    assert tokens[-1] == 'upload.h264.mp4'
+    assert 'libx264' in recipe
+    assert 'libx265' not in recipe
+
+
 @pytest_ffmpeg
 @pytest.mark.django_db
 def test_video_unsupported_codec_h264_board_recipe(
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -680,7 +680,9 @@ def test_video_unsupported_codec_raises_with_ffmpeg_recipe(
     assert "'hevc'" in msg
     assert 'h264' in msg
     recipe = excinfo.value.recipe
-    assert 'libx264' in recipe  # Pi 3 supports H.264 — recipe encodes to H.264.
+    assert (
+        'libx264' in recipe
+    )  # Pi 3 supports H.264 — recipe encodes to H.264.
     tokens = _shlex.split(recipe)
     assert tokens[1] == '-i'
     assert tokens[2] == 'beach-clip.mp4'
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -632,83 +632,104 @@ def test_video_supported_codec_writes_metadata_and_clears_processing(
     assert path.exists(src)
 
 
-@pytest_ffmpeg
 @pytest.mark.django_db
 def test_video_unsupported_codec_raises_with_ffmpeg_recipe(
     asset_dir: str, monkeypatch: pytest.MonkeyPatch
 ) -> None:
-    """An H.264 upload on a Pi 5 (HEVC only — Pi 5's mpv has no v4l2-
-    request H.264 hwdec) is rejected. The exception's message names
-    the rejected codec and supported set; its ``recipe`` attribute
-    carries an ffmpeg command pre-filled with the upload's filename
-    (taken from ``metadata.upload_name``, which the upload view
-    stashes at create time) so the operator can copy-paste it
-    verbatim."""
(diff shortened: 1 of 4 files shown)


vpetersson:
## On-device test report — physical Pi 5

Tested both halves of this PR on a physical Pi 5 (`DEVICE_TYPE=pi5`, current GHCR viewer image), then restored the board to its original state.

### 1. Playback: does Pi 5 actually play SW-decoded 1080p H.264 smoothly? ✅

The viewer image is unchanged by this PR, so I measured presented frame rate directly via the kernel `drm_vblank_event_delivered` tracepoint (zero observer effect) on 1080p Big Buck Bunny clips:

| Clip (1080p) | Decode path | Presented fps | Viewer CPU |
|---|---|---|---|
| 1080p30 **H.264** (High@L4.1) | software (Cortex-A76) | **27.4 fps** | 58% of one of 4 cores |
| 1080p30 HEVC | hardware (V4L2-request) | 27.0 fps | 18% of one core |

Both present at the 30 fps source cadence — **SW-decoded H.264 is just as smooth as HW-decoded HEVC**, at a higher but far-from-saturating CPU cost (~14% of total CPU). The "no frame drops on Cortex-A76" claim in the PR description holds on real hardware. (The old ~11 fps Pi 5 presentation bottleneck was fixed in #2975, which is already in the shipping viewer image.)

### 2. Gate change: clean before/after ✅

Patched the two hunks into the celery container and ran the real `_run_video_normalisation` on an actual 1080p H.264 file:

- **Stock (pre-PR):** `REJECTED → Video codec 'h264' is not hardware-decoded on this device. Supported: hevc.` — the exact error from #3092.
- **This PR:** `ACCEPTED, is_processing=False, codec=h264`; HW-decode set `{h264, hevc}`; `format_sort[0] = vcodec:hevc` (biases HEVC, falls back to H.264 as intended).

### 3. Live YouTube download — blocked by environment, not the PR ⚠️

The end-to-end `download_youtube_asset` run hit `HTTP Error 403: Forbidden` from yt-dlp (YouTube bot-blocking the media fetch from this network). That's at the download stage, *after* format selection, and unrelated to this PR's logic — which I validated deterministically above.

### One code-review note (not tested here)

The `format_sort` change biases **every** multi-codec board toward HEVC, not just Pi 5 — including x86, Pi 4, and Rock Pi 4 (previously all got `vcodec:h264`). For x86 specifically, HEVC playback in the cage compositor is known-broken (VAAPI `FormatError` → black screen), and the gate already accepts `hevc` on x86. So if YouTube ever serves an HEVC stream, an x86 device could now download HEVC, pass the gate, and black-screen at playback. The trigger is rare (YouTube seldom serves HEVC) but it's scope beyond the stated Pi 5 fix — worth considering whether to narrow the bias to Pi 5 only.

**Verdict:** the Pi 5 fix works as intended and the playback premise checks out on real hardware. 👍

vpetersson:
@TekScene Thank you for the contribution but please fix the linter error.

TekScene:
Good catch. The original implementation biased all HEVC-capable boards (Pi 4, x86, Rock Pi 4) toward HEVC in `format_sort`, which was wider than intended.

The updated approach introduces `_PREFERRED_DOWNLOAD_VCODEC` in `processing.py` alongside `_HW_DECODE_VIDEO_CODECS`, keeping all per-board codec knowledge in one place. Only Pi 5 is listed with a `hevc` preference — x86 and Pi 4 are intentionally absent:

- **x86**: HEVC via VAAPI is known to produce a black screen (FormatError in the cage compositor), so biasing toward HEVC downloads would cause silent playback failures if YouTube ever serves an HEVC stream.
- **Pi 4**: hardware-decodes both codecs equally well, so H.264 is preferred for its wider YouTube availability — no behaviour change from before this PR.

A dedicated test (`test_download_youtube_asset_x86_prefers_h264_format`) now guards the x86 path explicitly.

TekScene:
@vpetersson Just double checking you have seen the changes, in case the notification did not make it.
Thank you.

vpetersson:
Hey @TekScene - thank you for the reminder. Running now.

tests/test_processing.py
```diff
@@ -752,6 +853,36 @@ def test_pi3_64_hw_decode_set_is_h264_only(
...
+    'device_type, expected',
+    [
+        # Pi 5: HEVC is the hardware decode path; H.264 is software-only.
+        ('pi5', 'hevc'),
+        # All other boards: H.264 is the primary HW path and more available
+        # on YouTube. x86 HEVC via VAAPI is known-broken (black screen).
+        ('pi4-64', 'h264'),
+        ('pi3', 'h264'),
+        ('pi3-64', 'h264'),
+        ('x86', 'h264'),
```
vpetersson, line 871: Fine for now but we should probably move x86 to HEVC as most boards will support it but this is a bit more conservative and fine.

PR #3100, opened 2026-06-25
fix(install): add python3-dev for armv7l cffi build

vpetersson:
### Issues Fixed

32-bit Raspberry Pi OS (armv7l) installs of v2026.6.x abort while building `cffi` with:

```
fatal error: Python.h: No such file or directory
```

Reported on the forum: https://forums.screenly.io/t/anthias-v2026-6-3-installation-stops-halfway-on-pi3-b/6716

### Description

`bin/install.sh` uses `uv`, which normally manages its own Python interpreter (headers included). But `python-build-standalone` has no armv7l build, so on 32-bit Pi OS uv falls back to the **system** `python3`. There is also no prebuilt `cffi` wheel for armv7, so uv compiles it from sdist — which needs the CPython development headers (`Python.h`). The apt list already had `libffi-dev` and `libssl-dev` (cffi's C-lib deps) but not the Python headers.

Adds `python3-dev` to the apt install list. 64-bit/arm64/x86 installs use uv's managed Python and prebuilt wheels, so they never compile cffi — installing the package there is harmless and keeps the path uniform.

### Checklist

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -261,6 +261,11 @@ function install_packages() {
         "git"
         "libffi-dev"
         "libssl-dev"
+        # On 32-bit ARM (armv7l) there is no python-build-standalone
+        # binary, so uv falls back to the system interpreter and has to
+        # compile cffi from its sdist (no armv7 wheel exists). That build
+        # needs the CPython headers (Python.h), provided by python3-dev.
+        "python3-dev"
         "whois"
     )
 


vpetersson:
### Validated on native armv7 (32-bit) hardware

Ran the exact `install.sh` Python step — `uv sync --no-default-groups --group host --no-install-project` (uv pinned 0.9.17) — inside a native `arm32v7/debian:trixie` container on a Pi 4 host (arm64 kernel, 4 KB pages, so armv7 runs without qemu). Trixie ships Python 3.13, matching `.python-version`, so uv falls back to the system interpreter exactly as it does on a 32-bit Pi 3.

**Before (no `python3-dev`)** — reproduces the forum failure verbatim:

```
arm-linux-gnueabihf-gcc ... -I/usr/include/python3.13 -c src/c/_cffi_backend.c ...
src/c/_cffi_backend.c:2:10: fatal error: Python.h: No such file or directory
help: `cffi` (v2.0.0) was included because `anthias:host` depends on `ansible-core`
      which depends on `cryptography` which depends on `cffi`
=== EXIT CODE: 1 ===
```

**After (with `python3-dev` from this PR)** — succeeds:

```
Built cffi==2.0.0
Installed 17 packages  (+ cffi==2.0.0  + cryptography==49.0.0  + ansible-core==2.21.0 ...)
=== EXIT CODE: 0 ===
cffi 2.0.0 OK
```

Note: the Pi 5's default `2712` kernel uses **16 KB pages**, which can't load 4 KB-page armhf userspace (`libz.so.1: ELF load command ... not page-aligned`), so the 32-bit validation was done on the Pi 4 host instead — still native armv7, no emulation.

PR #3134, opened 2026-07-07
fix(viewer): restart to recover /dev/fb0 hotplugged after a headless boot

vpetersson:
## Problem

On linuxfb boards (32-bit Pi 1/2/3, Qt5), `wait_for_framebuffer` in `bin/start_viewer.sh` waits for `/dev/fb0` before launching the viewer — this is the #2959 guard that stops the headless crash-loop.

But the (privileged) viewer container's `/dev` is a **devtmpfs snapshot taken at container start**. If the container boots headless and a display is plugged in *later*, the host gains `/dev/fb0` but it **never appears inside the container's stale `/dev`** — so the wait loop blocks forever and the viewer never starts. The comment claimed the wait "self-heals on hotplug"; in a running container it silently does not.

Reproduced live on the Pi 2 testbed: plugging a monitor into a headless-booted board left the host with `/dev/fb0` while the container saw only `/sys/class/graphics/fb0`, and the viewer stayed parked (`still no /dev/fb0 after 6540s; waiting for a display`) until a manual `docker restart`.

## Fix

sysfs is bind-through from the host and stays truthful across the stale snapshot. When `/sys/class/graphics/fb0` exists but the container's `/dev/fb0` does not, exit non-zero so the `restart: always` policy recreates the container with a fresh `/dev` that includes `fb0` — the same exit-to-restart recovery the OOM and viewer-death paths already rely on. Genuinely headless devices (no sysfs `fb0` either) keep waiting quietly, so the #2959 behavior is preserved.

Note `/dev/fb0` and `/sys/class/graphics/fb0` are the same fbdev object (created/removed together by `register_framebuffer`/`unregister_framebuffer`), so a real headless box has neither.

## Validation (live on the Pi 2, `DEVICE_TYPE=pi2`, linuxfb/Qt5)

Exercised the real `wait_for_framebuffer` against each state:

| State | Setup | Result |
| --- | --- | --- |
| Display present | `/dev/fb0` present | returns 0, proceeds |
| Hotplug-stale | `/dev/fb0` removed, `/sys/.../fb0` present | prints message, `exit 1` |
| Headless | neither present | keeps waiting, never exits |

`bash -n` clean.

### DIFF
--- a/bin/start_viewer.sh
+++ b/bin/start_viewer.sh
@@ -464,11 +464,20 @@ release_boot_splash
 # Wait for the framebuffer instead of launching into a guaranteed crash.
 # No assumptions about which connector the panel is on or its resolution:
 # when a display is (re)connected the KMS driver creates /dev/fb0 and we
-# proceed; a genuinely headless device idles here quietly and self-heals
-# on hotplug. Only the linuxfb path needs the /dev/fb0 wait — eglfs has
-# no /dev/fb0 and is guarded by wait_for_eglfs_display above instead,
-# while cage (wayland) tolerates a missing display without crashing — so
-# the QT_QPA_PLATFORM guard makes this a no-op for those paths.
+# proceed; a genuinely headless device idles here quietly. Only the
+# linuxfb path needs the /dev/fb0 wait — eglfs has no /dev/fb0 and is
+# guarded by wait_for_eglfs_display above instead, while cage (wayland)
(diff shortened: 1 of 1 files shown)


vpetersson:
Folded into #3075 — the two fixes are the pre-launch and post-launch bookends of the same linuxfb HDMI-hotplug problem and belong together. The `wait_for_framebuffer` headless-boot recovery is now commit daa7f8d6 on #3075's branch.

Issue #1755, opened 2023-02-27
No way to see or set time in web interface


MisterAnderson91:
#### Overview of the Issue

Anthias v0.18.6 image has no way to see the host's current time. As such, there is no way to know what time/date assets need to activate them.

Setting assets to display for "1 day" etc. seems to only use the local time of the guest computer to set the asset start/end times.

There seems to be an assumption that the anthias host will have direct unfiltered internet access constantly and can track the time via NTP using whatever NTP server is added by default, in whatever timezone is set by default.

Requesting the following fixes:

* Display the host's current time in the web interface
* Use the host's local time to calculate asset schedules when using the premade "1 day" etc. instead of the time from the guest machine, as an immediate workaround for systems without the correct time set 
* Add the ability to change the host's current time directly in the web interface 
* And if possible, add the ability to configure an alternate NTP server (and set the correct timezone) for networks with one locally available

#### Reproduction Steps

* Apply v0.18.6 image from releases to microSD
* Boot Raspberry pi and leave on anthias splash screen until Anthias WiFi Connect AP is available on another device
* Connect to the AP, manually go to 192.168.42.1 in a browser.

#### Environment

* Raspberry Pi Hardware Version: 4B
* Raspberry Pi Network Setup: No network connection (wifi connection desired)
* Anthias Version: 0.18.6 pre-release image

vpetersson:
Great observation, @MisterAnderson91. Yes, the time on the device is set using NTP as you correctly assumed. However, as things stand, there's no way to set the timezone when running on Balena. This is obviously a problem when trying to schedule assets.

We'll have to look into this and probably need to add a timezone setting.

I did just have a look however, and the timezone is currently set to UTC by BalenaOS (which is a sensible default of course).

MisterAnderson91:
> 

A timezone option would be great, but just being able to see the current system time and set it is arguably more important because at least you could still workaround an incorrect time. It also needs the option to pick the NTP server for environments where there's a local one, or only specific URLs are whitelisted for internet access, etc. At the moment it's all just invisible.

extremeskillz:
[...]

PR #3163, opened 2026-07-08
feat(settings): add userspace timezone selection and a device clock

### LINKED ISSUE #1912: Add support for changing timezone settings via the Settings page.

vpetersson:
### Issues Fixed

- Closes https://github.com/Screenly/Anthias/issues/1912
- Substantially addresses https://github.com/Screenly/Anthias/issues/1755 and https://github.com/Screenly/Anthias/issues/2996 (timezone + visible device time; see note below on NTP/clock-setting)

### Description

Adds an operator-selectable **Timezone** in Settings that drives all schedule evaluation and time display, plus a **live device clock** on the System Info page.

The timezone is stored in `anthias.conf` and applied entirely in userspace — no host access, no reboot. It resolves with the precedence `anthias.conf` → `TZ` env → `/etc/timezone` → `UTC`, and every candidate is validated so a bad value can never crash-loop the settings module. A per-request middleware re-activates it so a save takes effect immediately across the UI, the REST API, and the server-evaluated viewer playlist; the viewer's own scheduler re-activates it on startup and on every `reload`.

This matters most on **balenaOS**, where the host is permanently UTC and there is no host-level timezone to inherit — so today every balena device schedules and displays in UTC. Setting the timezone in userspace is in fact balena's own recommended approach, and it's the only way to get correct local-time scheduling there.

Highlights:
- New `timezone` device setting; dropdown populated from the IANA zone list.
- Wired through both write paths (HTML form + `/api/v2/device_settings`) and the v2 serializers, with validation.
- System Info gains a **Device time** card that ticks client-side but is seeded from the *server* instant (so a wrong device clock is revealed, not masked) and renders in the device's own zone via `Intl`.
- `/api/v2/info` gains a `time` object (`iso` / `timezone` / `offset`).

**Out of scope (deliberately):** setting the absolute clock or a custom NTP server from the web UI. On balena that isn't possible from the app container — the Supervisor API has no time endpoint, and NTP is managed by the host (config.json `ntpServers` / DHCP). Those asks from #1755/#2996 are left as a possible docker-compose-only follow-up.

### Checklist

### DIFF
--- a/ansible/roles/anthias/files/anthias.conf
+++ b/ansible/roles/anthias/files/anthias.conf
@@ -4,6 +4,11 @@ configdir = .anthias
 database = .anthias/anthias.db
 use_ssl = False
 
+; IANA timezone used for scheduling and time display (e.g.
+; Europe/London). Empty follows the host / the TZ env var / UTC.
+; Set it from the Settings page rather than by hand.
+timezone =
+
 ; Opt-out from analytics
 analytics_opt_out = False
 
--- a/src/anthias_server/api/serializers/v2.py
+++ b/src/anthias_server/api/serializers/v2.py
@@ -367,7 +367,7 @@ class UpdateDeviceSettingsSerializerV2(Serializer[Any]):
         value = (value or '').strip()
         if value and not is_valid_time_zone(value):
             raise serializers.ValidationError(
-                f'Unknown or unavailable timezone: {value!r}.'
+                f'Unknown or unavailable timezone: {value}.'
             )
         return value
 
--- a/src/anthias_server/api/serializers/v2.py
+++ b/src/anthias_server/api/serializers/v2.py
@@ -336,10 +336,10 @@ class UpdateDeviceSettingsSerializerV2(Serializer[Any]):
         required=False, min_value=0, max_value=DURATION_S_MAX
     )
     date_format = CharField(required=False)
-    # Blank = "follow the host" (resolve_time_zone precedence). A
-    # non-blank value must be a zone Django will accept, validated the
-    # same way as the host zone so a save can never land a value that
-    # crash-loops the settings module on the next read.
+    # Blank defers to resolve_time_zone() (TZ env -> /etc/timezone ->
+    # UTC). A non-blank value must be a zone Django will accept,
+    # validated the same way as the host zone so a save can never land a
+    # value that crash-loops the settings module on the next read.
     timezone = CharField(required=False, allow_blank=True)
     show_splash = BooleanField(required=False)
     default_assets = BooleanField(required=False)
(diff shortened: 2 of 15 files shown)


vpetersson:
## End-to-end validation — Raspberry Pi 5 (`latest-pi5`)

Deployed this branch's code onto a Pi 5 (4 GB, `latest-pi5` stack) via a full-`src` overlay, restarted the server/celery/viewer containers, exercised the feature, then restored the device to its pristine image. All green.

### Functional
| Check | Result |
|---|---|
| Server boots with the change | ✅ `/api/v2/info` → 200 |
| Baseline, no override | ✅ follows host → `Europe/London` / `UTC+01:00` (BST) |
| New `time` object on `/api/v2/info` | ✅ `iso` + `timezone` + `offset` |
| **Live** override → `America/New_York` | ✅ persisted; `/api/v2/info` immediately reported `UTC-04:00` with no restart (per-request activation) |
| Scheduling honours the active zone (on-device) | ✅ same asset + same UTC instant: `is_active=True` at 09:00 Berlin (in a 09:00–17:00 window), `False` at 02:00 Chicago |
| Invalid zone (`Mars/Phobos`) | ✅ rejected `400` |
| Clear override | ✅ falls back to the host zone |
| Viewer container | ✅ healthy, regenerating the playlist |

### UI/UX (screenshotted via headless Chromium against the device)
Ran a short design review of the two new surfaces and applied three fixes (second commit):
- **Device time card** was stretching to the tall Memory card's height, leaving a dead vertical band — reordered the System Info grid so the two short text cards (Uptime + Device time) share a row and the two chart cards share theirs.
- **Humanised IANA names** (underscore → space) in the timezone dropdown and the clock sub-label; the stored value and the `Intl` data attribute keep the real id.
- Refreshed the **Display & playback** section blurb to mention timezone.

Settings shows the Timezone select styled identically to Audio output / Date format; System Info shows a live-ticking **Device time** card seeded from the *server* instant (so a wrong device clock is revealed, not masked) and rendered in the device's own zone.

### Not covered
x86 was not tested end-to-end (that testbed was in use); the x86 checklist box is left unchecked.

vpetersson:
Copilot review is clean ✅ — its latest pass on `cef1c946` reviewed all 15 changed files and generated no comments, and every finding from the earlier passes has been addressed and its thread resolved.

Fixes made in response to Copilot across the review rounds:
- Clock seed (`device_time.iso` + `/api/v2/info`) emits **seconds** precision so microseconds can't make an older JS `Date.parse()` return `NaN` and stall the live clock.
- The System Info clock script falls back to `Date.now()` where `performance.now()` is unavailable.
- Extracted `format_utc_offset()` into `anthias_server.lib.timezone`, shared by the page and the API (removed the duplicate).
- `get_configured_time_zone()` builds its default path from the shared `CONFIG_DIR` / `CONFIG_FILE` constants instead of hard-coding it.
- Snapshot the local instant + zone name once when building `device_time` (no intra-second skew).
- Reworded the "follow the host" comments to state the real blank-value fallback: `TZ env → /etc/timezone → UTC`.

CI is green across the board (mypy, python + typescript tests, linter, OpenAPI schema, build, CodeQL, SonarCloud). Ready for a maintainer's review.

PR #3164, opened 2026-07-08
pi3-64 hardware video playback (30fps, zero drops) + audio

### LINKED ISSUE #3084: Anthias v2026.6.3 - Blank Screen on Video Assets
This post is related to forum post: https://forums.screenly.io/t/anthias-v2026-6-3-blank-screen-on-video-assets/6720/4

I'm attaching the logs here for review. Let me know what you find.

[anthias-debug-20260618-105110.tar.gz](https://github.com/user-attachments/files/29100974/anthias-debug-20260618-105110.tar.gz)

vpetersson:
## What

Enables real hardware video playback on the **Raspberry Pi 3 arm64** board (`pi3-64`, VideoCore IV, eglfs) — **30 fps with zero ongoing dropped frames, with audio**.

Closes https://github.com/Screenly/Anthias/issues/3084

## Why

On `pi3-64`, QtMultimedia's `VideoOutput` cannot present a video node on the vc4 GL driver, so video rendered **black**; and compositing a full-screen video widget through eglfs GL caps at ~9 fps regardless. Anthias has a lot of active Pi 3 users, so this needed a proper hardware path, not a workaround.

## How

**Video** (`VideoView::gstPlayOverlay`) — in-process GStreamer, scanning out on a vc4 DRM **overlay plane** HW-composited with the eglfs UI:

```
filesrc ! qtdemux ! h264parse ! v4l2h264dec capture-io-mode=mmap ! queue ! kmssink (overlay plane)
```

- Uses eglfs's own DRM master fd + connector via `QPlatformNativeInterface` and a libdrm free-overlay-plane finder.
- `capture-io-mode=mmap` is the key: kmssink then **copies** each frame into its own scanout buffer instead of pinning the decoder's DPB pool — that pin deadlocked every plain decoder-direct attempt. The bcm2835 decoder emits I420 at full rate; the ISP convert-to-NV12 alternative is a second M2M pass that caps at ~18 fps, so it's only a fallback (`ANTHIAS_GST_ISP_CONVERT=1`).
- Degrades gracefully: CPU-raster appsink path, then QMediaPlayer, if DRM/plane resources can't be resolved.

**Audio** (`VideoView::gstStartAudio`) — a **separate** GStreamer pipeline (`qtdemux ! decodebin ! audioconvert ! audioresample ! alsasink` on the HDMI card):

- Independent clock/preroll/bus, so no audio condition (missing track, slow device, decode gap) can stall the video. A shared pipeline froze the video at the first frame.
- Adds `gstreamer1.0-libav` for the audio decoders (`avdec_ac3`/`aac`/`mp3`/…); decodebin reported a missing plug-in on AC3 with only faad (AAC) present.

**Plumbing**: only `pi3-64` links GStreamer (`AnthiasViewer.pro`, Dockerfile build + runtime deps); `_build_webview_env` enables the path; `start_viewer` chowns the GStreamer registry cache to avoid a per-launch rescan; `main.cpp` installs a fatal-signal backtrace handler.

## Testing

Validated on a **real Pi 3** (192.168.200.99) with a 1080p30 H.264/AC3 clip over HDMI, deployed as the full production image (no docker-cp, no env overrides):

- **~30 fps**, `dropped` flat at the startup ramp only (zero ongoing drops)
- Clean across the 60 s loop boundary; `respawns=1` (no crash, no deadlock)
- `/dev/video10` held (HW decode) on the vc4 overlay plane
- **HDMI PCM RUNNING** (AC3 → 48 kHz stereo) with video unaffected

Python unit tests for the env wiring pass (`tests/test_viewer.py`); `ruff check`/`format` clean.

## Follow-ups (not blocking)

- Pin the GStreamer plugin set to the RPi `+rpt3` archive for production parity.
- Per-HDMI-port audio routing (currently the resolved `audio-device` card, default HDMI).

### DIFF
--- a/bin/lib/viewer/common.sh
+++ b/bin/lib/viewer/common.sh
@@ -87,11 +87,16 @@ chown -Rf viewer /data/.anthias
 mkdir -p /data/.local/share/AnthiasViewer/QtWebEngine \
     /data/.cache/AnthiasViewer \
     /data/.cache/fontconfig \
+    /data/.cache/gstreamer-1.0 \
     /data/.pki
 
 chown -Rf viewer /data/.local/share/AnthiasViewer
 chown -Rf viewer /data/.cache/AnthiasViewer/
 chown -Rf viewer /data/.cache/fontconfig
+# pi3-64's video path links GStreamer; without a writable registry cache
+# the viewer rescans every plugin on each launch (slow startup — enough to
+# blow the 30s D-Bus handshake budget on the memory-tight 1GB board).
+chown -Rf viewer /data/.cache/gstreamer-1.0
 chown -Rf viewer /data/.pki
 
 # Qt + dbus + various Linux apps look up XDG_RUNTIME_DIR; without it they
--- a/docker/Dockerfile.viewer.j2
+++ b/docker/Dockerfile.viewer.j2
@@ -84,16 +84,30 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 # pi3-64 runtime for the in-process GStreamer HW video path (issue #3084):
 # gstreamer1.0-plugins-good carries the V4L2 elements (v4l2h264dec →
 # bcm2835 decoder, v4l2convert → bcm2835 ISP); -base/-bad round out
-# qtdemux / h264parse / appsink; -alsa is the audio sink. gstreamer1.0-libav
-# supplies the audio decoders (avdec_ac3 / avdec_aac / avdec_mp3 …) the
-# separate audio pipeline autoplugs — without it decodebin reports "missing
-# a plug-in" on anything but the few codecs -bad ships. Only this board
-# links GStreamer, so only this board ships the plugin set.
+# qtdemux / h264parse / appsink / kmssink; -alsa is the audio sink.
+# gstreamer1.0-libav supplies the audio decoders (avdec_ac3 / avdec_aac /
+# avdec_mp3 …) the separate audio pipeline autoplugs — without it decodebin
+# reports "missing a plug-in" on anything but the few codecs -bad ships.
+# Only this board links GStreamer, so only this board ships the plugin set.
+#
+# Pin base/bad/alsa to the RPi archive's ``+rpt3`` builds (the archive +
+# keyring are already configured by _rpt1-ffmpeg-pin.j2 above — pi3-64 is
+# in its board list). These are RPi-tuned for the Broadcom stack; -bad in
+# particular carries the vc4 ``kmssink`` the overlay-plane path scans out
+# through. -good (the V4L2 HW decode/convert elements) and -libav are NOT
+# in the RPi archive, so they stay on Debian (kernel-driven, works). The
+# +rpt3 plugins share upstream 1.26.2 with Debian's ``libgstreamer1.0-0``
+# core (not shipped in the RPi archive), so the ABI matches.
 {% if disable_cache_mounts %}
 RUN \
 {% else %}
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 {% endif %}
+    printf '%s\n' \
+        'Package: gstreamer1.0-plugins-base gstreamer1.0-plugins-bad gstreamer1.0-alsa libgstreamer-plugins-base1.0-0 libgstreamer-plugins-bad1.0-0' \
+        'Pin: origin archive.raspberrypi.com' \
+        'Pin-Priority: 1001' \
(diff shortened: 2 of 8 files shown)


vpetersson:
### Added: RPi `+rpt3` gstreamer pin (commit fe0de290)

Pinned the pi3-64 gstreamer **plugins-base / plugins-bad / alsa** to the RPi archive's `+rpt3` builds (RPi-tuned for the Broadcom stack; plugins-bad carries the vc4 `kmssink` the overlay path scans out through). **plugins-good** (V4L2 HW decode/convert) and **libav** (audio decoders) aren't in the RPi archive, so they stay on Debian. The `+rpt3` plugins share upstream 1.26.2 with Debian's `libgstreamer1.0-0` core → ABI matches.

Verified on a real Pi 3 (clean reboot): base/bad/alsa install as `1.26.2*+rpt3*`, good/libav as Debian; **30 fps zero ongoing drops, audio playing, respawns=1** — unchanged from the Debian-plugins build.

### Regression testing (pi4/pi5/x86)
In progress — the pi3-64 changes are board-guarded (`{% if board == 'pi3-64' %}`) / `#ifdef ANTHIAS_GSTREAMER`; only `main.cpp`'s crash handler and a `common.sh` cache dir touch other boards. **pi4-64 viewer builds clean** (compiles without `ANTHIAS_GSTREAMER` — no cross-board build regression); runtime checks on pi4/pi5/x86 to follow.

_Gapless in-slot looping split into its own stacked PR (#3174)._

vpetersson:
### Addressed Copilot review (commit ef01e89c)

- **Crash handler installed fleet-wide (main.cpp):** scoped to the pi3-64 build (`#ifdef ANTHIAS_GSTREAMER`); other boards keep default crash handling. (It already re-raised via `SIG_DFL`, so core dumps were never actually suppressed — but this keeps the change within the PR's board scope.)
- **`filesrc` gets a raw/file:// URI (gstPlay & gstPlayOverlay):** `gstPlay` now normalises a `file://` URI to a filesystem path before either pipeline is built. (Anthias passes bare local paths today; this is defensive.)
- **Appsink/raster fallback had no audio:** audio is now started once in `gstPlay`, covering **both** the overlay and the appsink-raster paths (previously only `gstPlayOverlay`). `gstStartAudio` escapes its own path; `gstPlayOverlay` drops the now-unused `options` param.
- **play() ignored gstPlay() result:** it now logs when every GStreamer path is exhausted (there is no QMediaPlayer fallback on pi3-64 — its VideoOutput black-screens, #3084 — so the supervisor respawn is the recovery).
- **stop() routes by gstMode (compile-time):** on pi3-64 the video path is *always* GStreamer (play() never reaches QMediaPlayer when `gstMode`), and `gstStop()` is null-safe, so this is correct in practice — no change needed.
- **onGstFrame() clears gstDrainPending at the start:** intentional — the coalescing is via the single-slot `gstLatestFrame` (latest-wins), and clearing early keeps latency low while still collapsing bursts into one queued drain. This path is the raster fallback only.

Re-validated on a real Pi 3: 30 fps zero ongoing drops + audio (3.5 mm); pi3-64 (with GStreamer) and pi4-64 (without) both compile.

vpetersson:
### Addressed Copilot review round 2 (commit e5f631ed)

- **`ANTHIAS_GST_QUEUE` OOM:** clamp to the default 4 when the env is 0/negative/non-numeric (0 → unbounded queue).
- **`gstStartAudio` ignored `set_state`:** now checks `GST_STATE_CHANGE_FAILURE`, logs `start-failed`, and tears the audio pipeline down (video keeps playing silently).
- **`containerFps` data race:** made the pad-probe→GUI handoff atomic (`QAtomicInteger` fps×1000 via `containerFps()`/`setContainerFps()`).
- **Signal-handler safety:** replaced `fprintf`/`fflush` with `write()` + `backtrace_symbols_fd` (async-signal-safe); dropped the unused `<cstdio>`.
- **Tests:** also assert `ANTHIAS_VIDEO_OVERLAY` is set on pi3-64 and cleared on other boards.

Re-validated on a real Pi 3: 30 fps zero ongoing drops, `container-fps=30.00`, audio (3.5 mm) playing; pi3-64 + pi4-64 both compile; 18 env tests pass.

vpetersson:
Round 11: switched the crash handler to `sigaction(SA_RESETHAND)` and made the drain-flag release ordered (`storeRelease`). The `toImage()`-in-`onRasterPainted` note is the **dormant QMediaPlayer-raster path** — pi3-64 runs gstMode, which delivers frames via `onGstFrame`/`setImage` (no `toImage` in the paint handler); that path is never exercised on this board.

PR #3189, opened 2026-07-14
fix(viewer): rotate pi3-64 video by dropping the vc4 overlay when rotated

vpetersson:
## Problem

On the 64-bit Raspberry Pi 3 (`pi3-64`, VideoCore IV, eglfs), video plays but stays locked in **landscape** when the screen is set to a portrait `screen_rotation` — while images, web pages and the UI all rotate correctly. Reported on the forum: https://forums.screenly.io/t/on-raspberry-pi-3-b-with-anthias-2026-07-1-it-seems-that-the-videos-dont-work/6730

## Root cause

pi3-64 can't present video through the QML `VideoOutput` RHI path on the vc4 GL driver, so it scans the decoded frame out on a dedicated vc4 **DRM overlay plane** (`kmssink`), composited by the display controller. That plane is independent of the eglfs `QOpenGLCompositor`, so `QT_QPA_EGLFS_ROTATION` — which rotates the GL scene (UI / images / web pages) — never touches it. The vc4 planes also advertise only `rotate-0` / `rotate-180`, so the overlay can't be turned 90°/270° in hardware at all.

## Fix

Take the hardware overlay path **only when the screen is upright**. When rotated, drop `ANTHIAS_VIDEO_OVERLAY` so `VideoView` falls back to the CPU-raster blit, whose widget composites through the eglfs-rotated backing store and therefore inherits the rotation like images/web pages do.

- **Upright:** unchanged — full-rate 30 fps HW overlay.
- **Rotated:** video rotates correctly, on the raster path (lower fps — the right trade for a rotated screen, and the only option since the hardware plane can't do 90°/270°).

Also corrects two stale comments that assumed every Qt6 board rotates video uniformly at the platform layer (they predated the overlay path).

## Validation

Unit tests cover the env selection (upright → overlay, rotated → raster, pi3-64 only). Verified end-to-end on a **real 64-bit Pi 3**:

- **Unfixed**, `screen_rotation=90`: two active DRM planes — the eglfs-rotated UI plane **plus** a landscape video overlay plane (bug reproduced).
- **With the fix**, `screen_rotation=90`: only the composited UI plane remains; the video is folded into it and rotates with the UI. The viewer process shows `ANTHIAS_VIDEO_RASTER=1`, `QT_QPA_EGLFS_ROTATION=90`, and no `ANTHIAS_VIDEO_OVERLAY`.

### DIFF
--- a/src/anthias_viewer/__init__.py
+++ b/src/anthias_viewer/__init__.py
@@ -245,7 +245,11 @@ def _build_webview_env() -> dict[str, str]:
       AnthiasViewer is a QWidget app, so webpages, images and video
       all rotate uniformly — no per-content rotation needed (and the
       old ``video-rotate`` path in media_player must stay off here or
-      the video double-rotates).
+      the video double-rotates). Exception: on pi3-64 the HW overlay-
+      plane video path (kmssink on a vc4 DRM plane) is scanned out
+      independently of the QOpenGLCompositor, so QT_QPA_EGLFS_ROTATION
+      does NOT rotate it (forum 6730); it is therefore taken only when
+      the screen is upright — see the pi3-64 block below.
 
     * linuxfb (pi2/pi3, Qt5): the plugin reads ``:rotation=N`` from
       QT_QPA_PLATFORM once at QPA init and rotates the framebuffer for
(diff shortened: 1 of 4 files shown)


vpetersson:
## Hardware-validated on a real 64-bit Pi 3, with the trade-offs measured

Extended the fix after measuring frame drops on the raster path. Full picture (1080p30 H.264, 30s steady-state window each):

| Screen rotation | Path taken | Presented fps | Dropped | CPU / thermal |
|---|---|---|---|---|
| 0° | HW overlay (`kms-overlay`) | 30.0 | 0% | idle, cool |
| **180°** | **HW overlay + vc4 plane `rotation=180`** | **29.9** | **0%** | ~7% CPU, 70°C flat |
| 90° / 270° | eglfs-composited CPU raster | ~9.5 | ~68% | CPU-bound, thermally throttles |

**Why 90/270 can't be smooth (hardware limit):** VideoCore IV has no 90/270 video rotation anywhere — the vc4 DRM planes advertise only `rotate-0`/`rotate-180` (measured), the ISP can't rotate 90/270, and the GPU `VideoOutput` path is black on this chip (#3084). Software rotation is the only option there, and it's CPU-bound: the raster `toImage` path does ~9.5fps, and a GStreamer `videoflip` does ~15fps *when cool* but collapses to single digits as the passively-cooled Pi 3 thermally throttles. So 90/270 rotated video is degraded-but-correct; smooth portrait *video* on a Pi 3 needs a Pi 4/5.

**Why 180° is now full-rate:** a DRM plane `rotate-180` is a free scanout-time transform (zero per-frame CPU). `VideoView::gstPlayOverlay` sets the kmssink plane's `rotation=4` when `QT_QPA_EGLFS_ROTATION=180`, so 180° keeps the 30fps HW overlay instead of falling to raster.

**Validation evidence (on the board):** at 180°, `sink=kms-overlay plane-id=98`, the video plane's `rotation` property reads `value: 4` (rotate-180), 29.9 fps / 0 drops, 70°C flat. At 90°, the overlay plane is gone (1 active plane) and it correctly uses the raster fallback.

vpetersson:
### Note: the GPU / zero-copy path was measured and is *slower* than raster on VC-IV — don't re-litigate

The obvious "just rotate on the GPU zero-copy like mpv/Kodi" idea was tested on the board (GStreamer GL vout, `v4l2h264dec ! glupload ! glcolorconvert ! glvideoflip ! glimagesink`, 1080p, max rate):

| Path (1080p) | fps | Temp |
|---|---|---|
| GPU zero-copy **+ rotate** | 6.6 | cool |
| GPU zero-copy, **no rotate** (baseline) | 6.7 | cool |
| CPU raster (this PR's 90/270 fallback) | ~9.5 | hot |
| DRM overlay scanout (0/180) | 30 | cool |

The GL path imports the decoder dmabuf fine and rotates for **free** (6.6 vs 6.7 — identical), and stays cool (GPU-bound, not CPU). But VideoCore IV's V3D can't push 1080p GLES video above **~6.7 fps even with no rotation** — *below* the CPU raster path. So zero-copy GPU rotation would be a regression for 90/270, not a fix. This is the same wall as #3084: the only thing on this SoC that presents 1080p30 is the display controller's scanout (the DRM overlay plane), which rotates 0/180 only. Smooth 90/270 portrait video needs Pi 4 (V3D 4.2) / Pi 5. Hence: overlay (+ HW plane rotate for 180) for 0/180, raster for 90/270.

PR #3275, opened 2026-08-07
fix(webview): reduce black screen during image-fetch after video

### LINKED ISSUE #3262: Black screen flash (~100ms) when transitioning between assets
### Description
When Anthias switches from one asset to the next (image/video/webpage) in the playlist, the screen briefly flashes black for roughly 100-200ms before the next asset appears. It's quite jarring visually and noticeably disrupts the viewing experience, especially when the playlist has short-duration assets and transitions happen frequently.

### Steps to Reproduce
1. Set up a playlist with 2 or more assets (e.g., image → image, image → video)
2. Let the playlist run in the viewer
3. Observe the screen during the transition between assets
4. Screen flashes black for ~100-200ms before the next asset appears

### Expected Behavior
Assets should transition smoothly (e.g., instant cut or fade) without a visible black screen flash in between.

### Screenshots
N/A (brief timing-based visual glitch, hard to capture in a still screenshot

### Environment
- Device: Raspberry Pi 4
- OS: balenaOS
- Anthias Version: v2026.8.0
- Installation Method: Release Image

Glassto:
### Issues Fixed

Fixes #3262

### Description

Reduces black screen time during the image-fetch portion of a video-to-image transition, by falling back to the last real raster frame while a fresh loadImage() fetch is in flight, instead of painting black for the whole round-trip. Applies to any real image load that starts from a blanked state (also covers webpage-to-image), documented in the view.h comment on lastRasterImage.

Does not address the video-surface teardown time (~460ms on Pi 4, separate from the fetch) or the deliberate blanks in playVideo()/loadPage().

### Checklist

### DIFF
--- a/src/anthias_webview/AnthiasViewer.pro
+++ b/src/anthias_webview/AnthiasViewer.pro
@@ -32,12 +32,14 @@ CONFIG += c++17
 SOURCES += src/main.cpp \
     src/mainwindow.cpp \
     src/view.cpp \
-    src/rotation.cpp
+    src/rotation.cpp \
+    src/image_fallback.cpp
 
 HEADERS += \
     src/mainwindow.h \
     src/view.h \
-    src/rotation.h
+    src/rotation.h \
+    src/image_fallback.h
 
 greaterThan(QT_MAJOR_VERSION, 5) {
     QT += multimedia quickwidgets
--- a/src/anthias_webview/src/image_fallback.cpp
+++ b/src/anthias_webview/src/image_fallback.cpp
@@ -1,0 +1,11 @@
+#include "image_fallback.h"
+
+namespace image_fallback
+{
+
+bool shouldUseLastRasterImage(bool currentImageIsNull, bool fallbackAllowed)
+{
+    return currentImageIsNull && fallbackAllowed;
+}
+
+}  // namespace image_fallback
--- a/src/anthias_webview/src/image_fallback.h
+++ b/src/anthias_webview/src/image_fallback.h
@@ -1,0 +1,28 @@
+#pragma once
+
+// Pure predicate behind View::paintEvent()'s black-flash fallback (a
+// real loadImage() fetch that starts from a blanked currentImage —
+// typically right after a video asset: shows the last decoded raster
+// frame instead of solid black while the fetch is in flight). The
+// three intentional blanks (playVideo(), the ``"null"`` sentinel in
+// loadImage(), loadPage()) must stay pure black, which is why this is
+// gated by a flag rather than firing on every null currentImage.
+//
+// Kept free of View / QtWebEngine, the same way rotation.cpp is, so
+// this one piece of the fallback logic is unit-testable against
+// QtCore alone (see tests/tests.pro). View owns the QImage state
+// (currentImage, lastRasterImage) and the flag itself; this only
+// answers the yes/no question.
+namespace image_fallback
+{
+// currentImageIsNull: View::currentImage.isNull() at paint time.
+// fallbackAllowed: View::fallbackToLastImageOnBlank, true only while
+// a real (non-"null") loadImage() request is outstanding; cleared on
+// every terminal outcome of that request (success, network error, or
+// decode failure) so a failed fetch goes black rather than leaving a
+// stale asset on screen indistinguishable from a successful rotation.
+//
+// Returns true when paintEvent() should paint lastRasterImage in
+// place of currentImage.
+bool shouldUseLastRasterImage(bool currentImageIsNull, bool fallbackAllowed);
+}  // namespace image_fallback
--- a/src/anthias_webview/src/view.cpp
+++ b/src/anthias_webview/src/view.cpp
@@ -31,6 +31,7 @@
 #include <QtGlobal>
 
 #include "view.h"
+#include "image_fallback.h"
 #include "rotation.h"
 
 // Attaches the operator-configured per-asset request headers (#2215) to
@@ -521,6 +522,7 @@ void View::loadPage(const QString &uri, bool skipSslVerify)
     hideVideoSurface();
 #endif
     currentImage = QImage();
+    fallbackToLastImageOnBlank = false;
     stopAnimation();
     // Drop any per-asset reload timer left over from the previous
     // webpage AND the prior asset's pending interval — the viewer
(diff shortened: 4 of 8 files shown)


Glassto:
Thanks for the thorough review, especially for pushing on the Qt5 handover question rather than letting me hand-wave past it.

**Flag leak on failure**: fixed exactly as suggested, cleared in the `finished` handler right after the staleness guard, before checking `error()`, so it's a single choke point for all three terminal paths (network error, errorOccurred, decode failure). A failed/corrupt fetch now goes black again instead of leaving the previous asset up.

**Test coverage**: extracted the predicate into `image_fallback.{h,cpp}`, same shape as `rotation.cpp`, QtCore-only, no WebEngine. `test_image_fallback.cpp` pins the three cases: non-null currentImage never falls back, the intentional blanks stay black, and a real fetch from a blanked state uses the fallback. Wired into `tests.pro` and the combined test runner.

**Qt5 / gst_fbdev_player question**: traced this through rather than asserting. `GstFbdevMediaPlayer.stop()` (media_player.py) blocks on `Popen.wait()`, SIGTERM then SIGKILL fallback, until the subprocess has actually exited, and `view_video()` doesn't return to `asset_loop` until that `stop()` call returns. So the next `view_image()`, and therefore any repaint of `currentImage`/`lastRasterImage` onto `/dev/fb0`, can't happen until `gst_fbdev_player` has confirmed-exited and stopped writing frames. No cross-process lock enforces this, it's an ordering guarantee from the Python call sequence, unchanged by this PR. Added a comment above `playVideo()` documenting the dependency, since you're right it's non-obvious from the `#if QT_VERSION >= 6` guard alone.

**Memory**: agreed it's worth a conscious call rather than silently accepting it. It's bounded to one extra buffer at a time (`lastRasterImage` is reassigned on every successful load, not accumulated), so I'd lean toward accepting the cost as-is unless your hardware numbers say otherwise on the 512 MB boards, in which case I'm open to gating the fallback off entirely below some RAM threshold. Will wait for your measurements before deciding.

**Board tested**: pi4-64 (Qt6/eglfs). I have not yet tested the Qt5/linuxfb path (Pi 2, Pi 3 32-bit) or the Qt6/Wayland path (Pi 5, x86) on real hardware,the Qt5 analysis above is from reading the code, not from running it on an armhf board.

Updated patch pushed to the branch..... :)

Glassto (commented):
Self-review pass on the full diff (not just the delta from the previous version).

- `view.h` / `view.cpp`: confirmed `fallbackToLastImageOnBlank` is reset in every place that blanks `currentImage` (loadImage's null sentinel, loadPage, playVideo) and cleared unconditionally in the QNetworkReply `finished` handler before the success/failure branch,
- `image_fallback.{h,cpp}`: pure function, no state, matches the shape of `rotation.cpp`.
- `test_image_fallback.cpp`: covers all three cases from the review discussion, ran locally via `bin/test_webview_cpp.sh`,   all green;.
- Comment above `playVideo()` accurately describes the Qt5/gst_fbdev_player ordering dependency, double-checked against `media_player.py`'s `stop()` implementation.

No leftover debug output, no unrelated changes. Marking this Comment(can't approve my own PR) since the self-review checklist item is genuinely done now, not just checked off.

Glassto:
Pushed. Going through in the order you raised things.

**Your two self-corrections, I  agree with you on both.** No RAM gating, dropped it as an option, nextImage already pinned that buffer before this PR existed. Rewrote the view.h comment to say that plainly instead of the ~8 MB claim. Appreciate you catching your own claim before I built on top of it.

**Qt5 comment**: fixed. You're right, play() runs before the sentinel, not after, and I had it backwards. Restated around the guarantee that actually holds, the flag only arms in a later real loadImage(), and stop() has already blocked before that can happen, not around the play()/null ordering.

**Test coverage comment**: narrowed it to what the file tests, the pure predicate, and said explicitly that the flag lifecycle (which paths set/clear it) isn't covered, since that's still in view.cpp. Kept the extraction, per your earlier suggestion, just stopped overclaiming what it guards.

**"re-arms" wording**: modified to "clears it again."

**hideVideoSurface() ordering**: moved the flag assignment before the call, so it's not resting on setVisible()'s repaint deferral anymore. Free, as you said.

**Scope**: going with your option (a), keeping it broad rather than restricting to the video case. Documented in the view.h comment now: any real loadImage() arms the fallback, so webpage-to-image gets it too, and lastRasterImage can be an asset from further back than the immediately preceding one. Decided a short-lived stale frame beats black here, same call as the video case, and the window's bounded by one round-trip either way.

Also took the two minor ones: comment now points at __init__.py:1504 instead of the stale :495, and "confirmed gone" is now "has exited," matching what wait() actually guarantees.

Ran `bin/test_webview_cpp.sh` again after all of this, still green. Didn't touch the CI gap, agreed that's pre-existing and out of scope here.

Glassto (commented):
Self-review on the updated diff.

Confirmed `fallbackToLastImageOnBlank` now sets before `hideVideoSurface()`, not after, and the Qt5 comment no longer claims an ordering it doesn't have. `view.h`'s memory note matches what's actually true now (no marginal cost on the static path). Test file's header only claims what it tests.

Nothing left unaddressed from either round. Good to go from my side... :)

Glassto:
@vpetersson checking in on this, it's been about a week since the last round of review feedback was addressed. Happy to make any further changes if something's still not right, just let me know. :)

No rush if you've just been busy, wanted to make sure it didn't fall through the cracks.....

vpetersson:
(Please note that you also need to sign your commits.)

Glassto:
Pushed the two `update()` calls from the last review, and re-signed the commit per your note (should show Verified now).

1. Network error branch (`view.cpp`, `loadImage()`): repaints on failure so the stale frame drops.
2. Decode-failure branch (`loadAsStaticImage()`): same fix.

Failed-fetch slot now goes black immediately, matching pre-PR behaviour, instead of freezing the previous asset for the whole slot. The 2s-slow-fetch improvement this PR targets is unaffected.

Glassto:
### Benefits of this PR

- Slightly reduces black screen time after transitions, from around 650 ms to approximately 460 ms, resulting in smoother visual updates.
- Significantly improves performance on slow networks, reducing the delay to display images from about 2.65 seconds to roughly 0.25 seconds: over 80% faster. :)
- Helps prevent outdated or corrupted images from being shown as fallback, improving visual stability.
- Overall, these changes contribute to a more responsive and seamless user experience.

I believe these adjustments will make the signage system more reliable and fluid for users.

vpetersson (approved):

PR #3302, opened 2026-08-19
fix(upload): tell the operator why an asset upload failed

mickzijdel:
### Issues Fixed

No existing issue for this one. I ran into accessing Anthias behind a Cloudflare tunnel. However, uploading a large video failed, and the UI just says `Upload failed — check the file and try again`. So I checked the console, and there I can see it is actually a Cloudflare 413 error about the file being too large which Anthias doesn't display.

Claude Opus 5 did the actual fixing but I did a lot of prodding to keep the change minimal. I have a plan for follow-up PRs to finish/fix the chunkable uploads feature which would bypass this issue entirely, but having clearer errors would be a good step regardless.

### Description

The Add asset upload path is raw XHR (`uploadFiles` / `uploadOne` in `home.ts`) rather than an htmx-managed request, so nothing replays a server toast for it and every failure resolved to that one string.

**A 413 on this path never comes from Anthias.** Django runs with `DATA_UPLOAD_MAX_MEMORY_SIZE = None`, and the Caddy sidecar that `bin/enable_ssl.sh` installs sets `request_body { max_size 0 }`. So the limit always belongs to a reverse proxy the operator put in front of the device. Cloudflare caps request bodies at 100 MB on its Free and Pro plans, which a 4K video easily exceeds. The response is generated at the edge and never reaches Django, so there is no `HX-Trigger` toast to replay and the status code is the only signal.

What this changes:

- **413** gets a message naming the size limit and saying it belongs to the server or a proxy, so you go looking in your CDN config instead of at the file.
- **403** gets a stale-page message for the CSRF case.
- **5xx** points at the device logs.
- **Everything else** keeps the original wording, so an unsupported file type reads exactly as it did before.

Transport failures are reported separately, and deliberately without a diagnosis. A proxy enforcing a body limit answers and closes while the browser is still writing the file, and the browser does not always salvage that response. Firefox drops it once enough of the body is still queued, where Chromium keeps it. A size rejection can therefore arrive indistinguishable from a dropped connection, so that message names both causes rather than guessing between them. When the status does survive, you still get the specific 413 message. We only hedge when the browser genuinely did not tell us.

Two things I deliberately did not do:

1. **No 507 branch.** The device does fill up, but `assets_upload` answers `ENOSPC` with a 200 plus an `HX-Trigger` toast carrying the shared `DISK_FULL_ERROR` string, which `fireToastFromHeader` already displays. Only the REST API returns 507, and this uploader never calls it. Adding one would have meant a second wording for a message `anthias_common.utils` centralises.
2. **No probe to guess at the lost-status case.** I built one (a `HEAD` to the upload URL after a transport failure, to see whether the route was still alive) and then dropped it. It needs a timeout, a 5xx-is-unhealthy rule and a `navigator.onLine` check just to be correct, it still misfires on a recovered wifi blip or a mid-upload restart, and I never measured how often the 413 actually gets lost at realistic Cloudflare sizes. Naming both causes costs nothing and is never wrong.

`UploadResult` becomes a discriminated union to carry the status from `uploadOne` up to the batch toast. Upload errors sit on screen for 8s rather than the 4s default, since they are the longest strings the store carries and a dismissed toast cannot be brought back.

### Testing

I ran a dev stack behind Caddy with a 1 MB `request_body` cap, which reproduces Cloudflare's behaviour at a size you can iterate on, and drove the real Add asset modal in both browsers:

| file | Chromium | Firefox |
|---|---|---|
| 2 MB (over cap) | size message | size message |
| 36 MB (over cap) | size message | transport message |
| 100 KB (under cap) | uploads, row created | uploads, row created |

The same files on `master` all produce the old generic string. 

11 new unit tests (`bun test`, 97 passing overall) cover the status mapping, the status surviving the trip up from `uploadOne`, the toast lifetime, a batch aborting on the first transport failure without attempting the rest, and a partial batch still committing the rows that succeeded. I mutation-checked them: dropping the batch `break`, discarding the status, removing either mapping, dropping the toast lifetime, and skipping the success commit each make them fail.

`bunx tsc --noEmit` is clean. No Python changed, so no server behaviour changed.

### Documentation

The reverse-proxy FAQ already warned that a proxy breaks uploads, but only for the rewritten `Host` header and the CSRF rejection that follows. A body-size cap is the other way a proxy breaks uploads, and it looks different from the outside: everything works except large videos, so you have no reason to suspect your proxy. The second commit adds that case to the same answer, with the directive for nginx, Apache and Caddy, and a note that Cloudflare's ceiling is set by your plan and can be lowered but not raised.

### Checklist

Frontend and FAQ only, tested in Chromium and Firefox against a dev stack rather than on a Pi or an x86 device. I only have one Pi and it is currently being used all the time so I can't test it. 

### DIFF
--- a/src/anthias_server/app/static/src/home.test.ts
+++ b/src/anthias_server/app/static/src/home.test.ts
@@ -1,0 +1,161 @@
+// Wiring tests for the upload batch. Run with
+// `bun test src/anthias_server/app/static/src/home.test.ts`.
+//
+// home/upload-error.test pins the status → message table; this drives
+// the real `uploadFiles` through a stubbed XMLHttpRequest so a
+// regression in the plumbing — a status dropped on the way up, a batch
+// that fails to abort — cannot pass just because the table still holds.
+
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+
+import './home'
+
+type Toast = { kind: string; message: string; ttlMs?: number }
+type Outcome = { status: number } | { transport: true }
+
+const realXhr = globalThis.XMLHttpRequest
+
+let sends: number
+
+// Each send() consumes the next outcome, so a batch can be given a
+// different fate per file.
+function stubXhr(outcomes: Outcome[]): void {
+  let index = 0
+  sends = 0
+  ;(globalThis as unknown as { XMLHttpRequest: unknown }).XMLHttpRequest =
+    function XMLHttpRequestStub() {
+      const outcome = outcomes[Math.min(index++, outcomes.length - 1)]
+      const handlers: Record<string, (() => void)[]> = {}
+      return {
+        status: 'status' in outcome ? outcome.status : 0,
+        upload: { addEventListener: () => {} },
+        open: () => {},
+        setRequestHeader: () => {},
+        getResponseHeader: () => null,
+        addEventListener(type: string, fn: () => void) {
+          ;(handlers[type] ??= []).push(fn)
+        },
+        send() {
+          sends += 1
+          const type = 'status' in outcome ? 'load' : 'error'
+          queueMicrotask(() => handlers[type]?.forEach((fn) => fn()))
+        },
+      }
+    }
+}
+
+// `uploadFiles` only reaches for `files`, `form` and `value`, so a
+// literal is steadier here than building a real FileList in happy-dom.
+function fileInput(...names: string[]): HTMLInputElement {
+  return {
+    value: '',
+    files: names.map((n) => new File(['x'], n, { type: 'video/mp4' })),
+    form: {
+      getAttribute: () => '/assets/upload/',
+      querySelector: () => ({ value: 'test-csrf' }),
+    },
+  } as unknown as HTMLInputElement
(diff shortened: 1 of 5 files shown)


vpetersson:
Very nice! I didn't consider this use case but it's a good one to support indeed.

mickzijdel:
Great! Let me know if you want any changes. Are you happy for me to follow-up with a PR to chunk uploads to stay under the limit?

vpetersson:
> Great! Let me know if you want any changes. Are you happy for me to follow-up with a PR to chunk uploads to stay under the limit?

Yeah let's do chunk'd uploads as a follow-up.

vpetersson (approved):

vpetersson:
@mickzijdel Actually, can't merge this since the commit aren't signed. Sign the commits and we should be good!

mickzijdel:
Did that @vpetersson (not sure if the force-push notifies you)

vpetersson:
It's live! Thanks @mickzijdel!

PR #3306, opened 2026-08-20
fix(auth): answer htmx callers with HX-Redirect, not a 302

mickzijdel:
### Issues Fixed

No existing issue. I found this while working on the upload error messages in #3302: if your session expires while the Add asset modal is open, the upload reports success and the file is never stored, but nothing on screen says otherwise.

The rest is all Claude 5 Opus.

### Description

`@authorized` answers an unauthenticated request with a 302 to `/login/`. That is the right answer for a browser navigating to a page, and but wrong for anything calling from JavaScript, because XMLHttpRequest and fetch both follow redirects themselves. The 302 is consumed before any client code runs, so what arrives is the login page with a status of 200.

Two callers reach `_login_redirect` from JavaScript, and both handle that badly:

1. **The six htmx endpoints.** htmx swaps the response into whatever fragment it asked for, so an expired session replaces the asset table with a whole login page rendered inside the table. The URL bar still says `/`.
2. **The uploader in `home.ts`.** It is raw XHR and treats any 2xx without an error toast as success, so it counts the file as uploaded, closes the modal, refreshes the table, and reports nothing wrong.

The second one is silent data loss, which is why I went looking for a root fix rather than a client-side check. There is no client-side check available: you cannot see a redirect that the browser already followed for you.

So the server now says what it means. When the request carries `HX-Request`, `_login_redirect` answers `204` with an `HX-Redirect` header instead of a 302. htmx acts on that natively and navigates to the login page. The uploader is not htmx-managed, so it reads the header itself and does the same.

**Requests without the header are untouched.** A browser typing a URL or following a link still gets its 302 with `?next=` filled in, which is what it wants. Only the six htmx endpoints and the uploader change, and all seven were already doing something wrong.

The decorator already treated htmx requests specially: `_is_safe_login_next_source` drops the fragment URL from `?next=` so signing in doesn't land you on a bare table partial. This finishes that thought rather than introducing a new concept.

One existing test changed. `test_authorized_drops_next_for_htmx_partial` asserted the old 302 contract, so it now asserts the same guarantee against `HX-Redirect`, with its docstring updated to say why.

### Testing

I ran a dev stack with auth enabled and deleted the session cookie to simulate expiry.

Same URL, same expired session, with and without the header:

```
GET /_partials/asset-table/  HX-Request: true  ->  204, HX-Redirect: /login/
GET /_partials/asset-table/  (no header)       ->  302, Location: /login/?next=...
```

Uploading a file with an expired session, measured on both sides:

| | master | this branch |
|---|---|---|
| Assets in the database | 7 -> 7 | 7 -> 7 |
| Ends up on | `/`, modal closed | `/login/` |
| Toasts shown | none | navigated away |

Master stores nothing and says nothing: the modal closes exactly as it does on success.

I also left the dashboard idle for 12 seconds with an expired session and touched nothing. On master the 5s table poll leaves you on `/` with a password field where the asset table used to be. On this branch it navigates to the login page.

44 tests in `tests/test_auth.py`, 1992 in the suite overall, and 99 frontend tests pass. `ruff check`, `ruff format --check` and `mypy` are clean on the changed files.

### Checklist

Tested against a dev stack rather than on hardware. My only Pi is in use at the moment so I can't test it there.

### DIFF
--- a/src/anthias_server/app/static/src/home.test.ts
+++ b/src/anthias_server/app/static/src/home.test.ts
@@ -11,7 +11,9 @@ import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
 import './home'
 
 type Toast = { kind: string; message: string; ttlMs?: number }
-type Outcome = { status: number } | { transport: true }
+type Outcome =
+  | { status: number; hxRedirect?: string }
+  | { transport: true }
 
 const realXhr = globalThis.XMLHttpRequest
 
@@ -31,7 +33,10 @@ function stubXhr(outcomes: Outcome[]): void {
         upload: { addEventListener: () => {} },
         open: () => {},
         setRequestHeader: () => {},
-        getResponseHeader: () => null,
+        getResponseHeader: (name: string) =>
+          name === 'HX-Redirect' && 'hxRedirect' in outcome
+            ? (outcome.hxRedirect ?? null)
+            : null,
         addEventListener(type: string, fn: () => void) {
           ;(handlers[type] ??= []).push(fn)
         },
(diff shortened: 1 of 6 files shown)


vpetersson:
Love it. Not able to look at this for a few days due to travels. Will review when back.

mickzijdel:
Great, sounds good. I'll submit the PR for chunking uploads too. Looking forward to your review, and safe travels.

PR #3328, opened 2026-09-14
fix(viewer): anchor the per-asset wait on the tick start, not a flat duration

DiasonD:
### Issues Fixed

Related to the discussion in issue 3319 (wall-clock playlist alignment). This does not implement that feature and does not close the issue. It reduces the accumulating drift caused by work before each asset's wait; video teardown remains outside the deadline.

### Description

`asset_loop` waited a flat `duration` after everything the tick did first (scheduler refresh, DB reads, the D-Bus load call), so every rotation took `duration` plus a device-specific overhead. Two players cycling the same playlist drifted apart by seconds within a few hours. The docs already promise an asset "is displayed for the specified duration", and the D-Bus `loadImage` / `loadPage` calls return before rendering, so the flat wait never bought on-screen time; it only lengthened the cycle.

- `tick_start = monotonic()` is now the first statement of `asset_loop`, and both wait sites (image/web in `asset_loop`, video in `view_video`) wait until `tick_start + duration` through one helper, `_seconds_until`.
- Clamped at zero: a mid-tick browser respawn can outlast a short duration, and the asset then just moves on. Not logged separately, since both wait sites log the computed wait next to the configured duration and a zero-duration asset would otherwise warn every tick.
- The "Sleeping for" line (image/web) and the new "Playing video for" line (video) keep the configured integer duration next to the computed wait, at info level, so a zero clamp is visible without debug logging.

**Video semantics.** For a video asset, `duration` is the clip length probed at upload, so the wait is the playback window rather than a display timer. Anchoring the wait on the tick start therefore trims the tick's pre-play overhead off the end of every clip: a few tens of ms on x86, proportionally more on a slow board or when a browser respawn lands mid-tick. Master already truncates a clip's fractional tail through `clamp_duration`'s `int()`, so this is a small extension of an existing cut. It is the intended trade: reducing drift from pre-play overhead at the cost of the last few frames of a clip.

**Not covered here.** `view_video` still calls `media_player.stop()` after the wait returns, so player teardown falls outside the deadline. On x86 that is a cheap D-Bus `stopVideo`; on the Qt5 linuxfb boards `GstFbdevMediaPlayer.stop()` is a `killpg` plus up to two 3 s waits. Teardown finishes before `asset_loop` returns, and the next tick records `tick_start` only after that return. Its cost therefore adds to every video rotation and is not absorbed by the following asset's deadline. In the [reviewer's Pi 2 measurements](https://github.com/Screenly/Anthias/pull/3328#pullrequestreview-5260697178), approximately 313 ms of video overhead remained per rotation; the measured mixed playlist's drift fell from 36.4 to 28.5 s/hour. This behavior predates the PR. Accounting for teardown is a separate timing change that needs regression coverage for a slow `stop()` and device validation.

What this does not do: it does not phase-align players (each still starts its counter wherever it was at boot) and it does not remove the residual crystal-rate drift between devices. It subtracts work before the wait from the asset's wait budget, reducing accumulated drift. It does not guarantee exact cycle timing: video teardown, work that exceeds an asset's duration, and scheduling delays can still extend the cycle.

Tests: three new cases drive `asset_loop` on a fake monotonic clock (image, video, and the zero clamp), and the existing max-duration test pins `monotonic` so its exact-timeout assertion stays valid. `ruff check`, `ruff format --check`, `mypy` and the viewer test files pass.

### Checklist

https://claude.ai/code/session_01Gos7GxdfJQkjapBvmXzGQS


DiasonD:
Thanks for the correction and the Pi 2 measurements in the [latest review](https://github.com/Screenly/Anthias/pull/3328#pullrequestreview-5260697178). You're right: `media_player.stop()` finishes before the next tick records `tick_start`, so teardown adds to every video rotation rather than being absorbed by the following asset.

I've corrected the original PR description's **Not covered here** paragraph, added a reference to your measurements, and revised the other timing claims so they no longer imply exact cycle timing or removal of all accumulating drift. Accounting for teardown remains a separate follow-up requiring regression coverage and device validation. No code changes were made.

vpetersson (approved):

vpetersson:
@DiasonD Approved, but please sign the commits. I can't merge it w/out that.

DiasonD:
Commits are now signed (SSH signature, both show as Verified). Code is unchanged from the approved revision, only the hashes moved.

